PR #11, opened 2011-07-05
Find APE tags even if there's a Lyrics3v2 tag present

HessiJames:
Currently taglib can't find APE tags if there's a Lyrics3v2 tag present.
I wrote a patch some time ago but it didn't get noticed, so I hope that requesting a pull here is the proper way.

If you need test files, please write me a mail or private message.

http://bugs.kde.org/show_bug.cgi?id=254223

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,7 @@ TagLib 1.8 (In Development)
  * Support for writing ID3v2.3 tags.
  * Added methods for checking if WMA and MP4 files are DRM-protected.
  * Started using atomic int operations for reference counting.
+ * Find APE tags even if there's a Lyrics3v2 tag present (BUG:254223).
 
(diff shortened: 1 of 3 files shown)


TsudaKageyu:
Hi guys, I'd like to hear someones opinion about whether this feature is still needed. 

If it's needed, I'm willing to merge this (or lyric3v2 branch) manually. However, I'm not sure if we should do it.

sbooth:
My opinion is that this isn't needed- I guess that MP3 files with the combination of APE and lyrics3v2 tags are rarely seen in the wild.

TsudaKageyu:
Sorry @HessiJames, I closed this PR since it looks obsolete.
However, I would be willing to implement this feature again if it would turn out to be still useful.

Issue #98, opened 2013-01-25
FLAC metadata blocks with 0 length invalidate the file


sbooth:
Commit ad9ffc62e6fac5c47f46eb96b39c614e32742eb5 added the behavior where FLAC metadata blocks with a length of 0 cause the file to be marked as invalid.  I've received numerous reports from users with FLACs purchased/downloaded from third party websites that their tags are no longer showing up.  Specifically, it seems WaveLab is creating seektables with a length of 0 causing tag parsing to fail.  A 0 length seektable does not conform to the FLAC spec, however `metaflac` handles the files correctly and I believe the behavior might be a bit too strict.

FestusHagen:
#637 Closes this?

-Enjoy

PR #99, opened 2013-02-17
Stop loading a FLAC file if it has too many metadata blocks.

davidsansome:
A file containing all \0 bytes is a valid stream of empty STREAMINFO
METADATA_BLOCKs.  FLAC::File::scan reads all these into memory, if the file
is large Taglib may consume all the memory on the machine.

This patch adds a safety that stops reading after 1024 blocks.  (1024 should
be enough for anyone).

For a (large) example file that triggers the problem, see:
https://docs.google.com/file/d/0Byds9jlkR0IxQnpYUUxTZnVtQU0/edit
Uploaded by this user:
http://code.google.com/p/clementine-player/issues/detail?id=3500


TsudaKageyu:
This issue is already fixed by #98. Thanks!

PR #222, opened 2013-05-26
Fixed some MSVC specific warnings

TsudaKageyu:
Fixed some MSVC warnings when compiling a DLL.
I confirmed that it won't break the existing DLL exports despite the fact that it modifies the public header.

### DIFF
--- a/taglib/toolkit/tiostream.cpp
+++ b/taglib/toolkit/tiostream.cpp
@@ -69,44 +69,59 @@ namespace
   }
 }
 
-// If WinNT, stores a Unicode string into m_wname directly.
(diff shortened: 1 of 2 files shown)


CyberSinh:
That commit has been reverted in the master branch. Is there any plan to fix that MSVC warning in next release of TagLib?

TsudaKageyu:
Currently, I still have no idea how to fix the warnings without breaking the binary compatibility.
We can use ```#pragma warning```, but I hate to do it, especially in a public header.

They will be fixed in version 2.0. However, we have no concrete schedule for it.

PR #301, opened 2013-10-16
Handle iTunes ID3v2.3 hacks

dukeyin:
iTunes writes the 2.4 frames TSOA, TSOT, TSOP to 2.3 files.  (It
additionally defines TSO2 and TSOC for both 2.3 and 2.4.)  TagLib should
not delete these frames.


gogglesguy:
Maybe nicer to put the #ifdef within the unsupportedFrames array. Easier to see which fields don't get added and you don't need to separately comment on it. Either way, no reason why this shouldn't be merged.

PR #303, opened 2013-10-19
Stop Release Date data loss when reading ID3v2.3

dukeyin:
The ID3v2.4 release date (TDRC) replaced multiple 2.3 frames (TYER,
TDAT, and TIME).  TagLib currently throws away TDAT and TIME.  This data
loss is fixed with the following changes:
- FrameFactory no longer renames TYER to TDRC (this change isn't
actually necessary)
- FrameFactory no longer discards TDAT or TIME
- During Tag::parse(ByteVector), the pointers to TYER, TDAT, and TIME
that FrameFactory::createFrame(...) returns are saved.  After the
createFrame loop, TDRC is manually created from TYER, TDAT, and TIME.
(TIME is only used if TDAT exists, and TDAT is only used if TYER
exists).  TIME is inserted into TDRC with least-necessary precision as
specified in "id3v2.40-structure.txt"

This change should not break any existing apps using an unaltered copy
of TagLib to retrieve the TDRC frame.  (It was never possible to find
TYER, TDAT, or TIME in an unaltered copy of TagLib.)

### DIFF
--- a/taglib/mpeg/id3v2/id3v2framefactory.cpp
+++ b/taglib/mpeg/id3v2/id3v2framefactory.cpp
@@ -378,12 +378,11 @@ bool FrameFactory::updateFrame(Frame::Header *header) const
 
   case 3: // ID3v2.3
   {
+    // TDAT and TIME are deprecated, but need to be merged with TYER into TDRC.
     if(frameID == "EQUA" ||
        frameID == "RVAD" ||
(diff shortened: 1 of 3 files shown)


TsudaKageyu:
Sorry for closing this, but we merged an equivalent patch. Thanks!

PR #362, opened 2014-03-30
Add support for synchronized lyrics (SYLT) and event timing codes (ETCO) for ID3v2

ufleisch:
Could you please consider merging this branch, which contains two commits, the first to support SYLT frames, the second to support ETCO frames.

### DIFF
--- a/taglib/CMakeLists.txt
+++ b/taglib/CMakeLists.txt
@@ -67,6 +67,7 @@ set(tag_HDRS
   mpeg/id3v2/id3v2tag.h
   mpeg/id3v2/frames/attachedpictureframe.h
   mpeg/id3v2/frames/commentsframe.h
+  mpeg/id3v2/frames/eventtimingcodesframe.h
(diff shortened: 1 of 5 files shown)


velihdzen:
The toString() method behaves differently between USLT and SYLT. USLT(UnsynchronizedLyricsFrame) toString() returns the 'text' field which is the actual lyric, but SYLT(SynchronizedLyricsFrame) toString() returns the 'description' field which could be empty. Why don't SYLT be consistent with USLT, return the lyric field 'synchedText'?

ufleisch:
The description is already available via `description()`, therefore returning it also via `toString()` is not extremely useful. Returning a textual representation of the synchronized lyrics (e.g. in LRC format) would be more useful. Then it would also be consistent to offer the possibility to import in LRC format via `setText()`. However, this is not compatible with the existing behavior, but we could consider this change for TagLib 2.x.

PR #396, opened 2014-07-11
Implementation of reading and writing Acidized loop information to a WAV file

rupertdaniel:
Based on the spec published here - http://www.kvraudio.com/forum/viewtopic.php?t=172636


taglib/riff/wav/acidinfo.cpp
```diff
@@ -0,0 +1,186 @@
...
+}
+
+void RIFF::WAV::AcidInfo::setTempo(float tempo)
+{
+  d->tempo = tempo;
+}
+
+bool RIFF::WAV::AcidInfo::isEmpty() const
+{
+  if(d->tempo >= 0.00)
```
gogglesguy: d->tempo >= 0.0f

PR #444, opened 2014-10-05
cmake: improve distro integration allowing to use the standard CMake options

tSed:
Hi all,

This patch makes taglib more standard wrt CMake options, so makes easier its integration in distros.

Regards,

Samuel


TsudaKageyu:
Sorry for being this late. Thanks!

PR #463, opened 2014-12-13
cmake: Now compiles as a library within a larger cmake based project

krumberg:
I had to make these changes to make taglib compile as a library within my cmake based project


FestusHagen:
Can be closed ...
  Commit 9824367 has resolution already merged.
  Commit 9e4fab9 is unnecessary.

PR #464, opened 2014-12-14
Make embPos unsigned

xhochy:
This fixes:

```
/home/xhochy/Development/taglib/taglib/mpeg/id3v2/frames/tableofcontentsframe.cpp: In member function ‘virtual void TagLib::ID3v2::TableOfContentsFrame::parseF
ields(const TagLib::ByteVector&)’:
/home/xhochy/Development/taglib/taglib/mpeg/id3v2/frames/tableofcontentsframe.cpp:254:16: warning: comparison between signed and unsigned integer expressions [
-Wsign-compare]
   while(embPos < size - header()->size()) {
```

This change is valid as embPos is initially zero and only non-negative values are added to it.


gogglesguy:
This one can be closed.

PR #467, opened 2014-12-19
Check for bad ID3v2 frame header flags.

### LINKED ISSUE #466: Infinite loop with fuzzed file
Parsing the following file with the tagreader example results in an infinite (or just excessive) loop with memory allocation. It depletes my 16G quickly.
https://www.dropbox.com/s/pzqtbfwlkufcnce/taglib_loop.mp3

<pre>
$ <b>ulimit -v 8000000;</b> gdb --args ~/repository/taglib/build/examples/tagreader taglib_loop.mp3
GNU gdb (Gentoo 7.7.1 p1) 7.7.1
Copyright (C) 2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://bugs.gentoo.org/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /home/cocobo/repository/taglib/build/examples/tagreader...done.
(gdb) r
Starting program: /home/cocobo/repository/taglib/build/examples/tagreader taglib_loop.mp3
warning: Could not load shared library symbols for linux-vdso.so.1.
Do you need "set solib-search-path" or "set sysroot"?
******************** "taglib_loop.mp3" ********************
TagLib: String::UTF8toUTF16() - Unicode conversion error.
terminate called after throwing an instance of 'St9bad_alloc'
  what():  std::bad_alloc

Program received signal SIGABRT, Aborted.
0x00007ffff6d5e597 in raise () from /lib64/libc.so.6
(gdb) bt full
#0  0x00007ffff6d5e597 in raise () from /lib64/libc.so.6
No symbol table info available.
#1  0x00007ffff6d5f8d8 in abort () from /lib64/libc.so.6
No symbol table info available.
#2  0x00007ffff7642845 in __gnu_cxx::__verbose_terminate_handler() ()
   from /usr/lib/gcc/x86_64-pc-linux-gnu/4.8.3/libstdc++.so.6
No symbol table info available.
#3  0x00007ffff7640976 in __cxxabiv1::__terminate(void (*)()) ()
   from /usr/lib/gcc/x86_64-pc-linux-gnu/4.8.3/libstdc++.so.6
No symbol table info available.
#4  0x00007ffff76409a3 in std::terminate() () from /usr/lib/gcc/x86_64-pc-linux-gnu/4.8.3/libstdc++.so.6
No symbol table info available.
#5  0x00007ffff7640bc2 in __cxa_throw () from /usr/lib/gcc/x86_64-pc-linux-gnu/4.8.3/libstdc++.so.6
No symbol table info available.
#6  0x00007ffff76410ad in operator new(unsigned long) () from /usr/lib/gcc/x86_64-pc-linux-gnu/4.8.3/libstdc++.so.6
No symbol table info available.
#7  0x00007ffff7a963e9 in allocate (__n=1, this=<optimized out>)
    at /usr/lib/gcc/x86_64-pc-linux-gnu/4.8.3/include/g++-v4/ext/new_allocator.h:104
No locals.
#8  _M_get_node (this=<optimized out>) at /usr/lib/gcc/x86_64-pc-linux-gnu/4.8.3/include/g++-v4/bits/stl_list.h:334
No locals.
#9  _M_create_node (this=<optimized out>, __x=...)
    at /usr/lib/gcc/x86_64-pc-linux-gnu/4.8.3/include/g++-v4/bits/stl_list.h:484
No locals.
#10 _M_insert (this=<optimized out>, __x=..., __position=...)
    at /usr/lib/gcc/x86_64-pc-linux-gnu/4.8.3/include/g++-v4/bits/stl_list.h:1553
No locals.
#11 push_back (__x=..., this=0x614cf8) at /usr/lib/gcc/x86_64-pc-linux-gnu/4.8.3/include/g++-v4/bits/stl_list.h:1016
No locals.
#12 append (item=..., this=<optimized out>) at /home/cocobo/repository/taglib/taglib/toolkit/tlist.tcc:160
No locals.
#13 TagLib::ByteVectorList::split (v=..., pattern=..., byteAlign=1, max=max@entry=0)
    at /home/cocobo/repository/taglib/taglib/toolkit/tbytevectorlist.cpp:58
        offset = 110242575
        l = {<TagLib::List<TagLib::ByteVector>> = {
            _vptr.List = 0x7ffff7dd46d0 <vtable for TagLib::ByteVectorList+16>, d = 0x614cf0}, d = 0x6149f0}
        previousOffset = 110242575
#14 0x00007ffff7a96bf8 in TagLib::ByteVectorList::split (v=..., pattern=..., byteAlign=<optimized out>)
    at /home/cocobo/repository/taglib/taglib/toolkit/tbytevectorlist.cpp:42
No locals.
#15 0x00007ffff79d4681 in TagLib::ID3v2::TextIdentificationFrame::parseFields (this=this@entry=0x614840, data=...)
    at /home/cocobo/repository/taglib/taglib/mpeg/id3v2/frames/textidentificationframe.cpp:207
        byteAlign = 1
        <b>dataLength = -1414812758</b>
        l = {<TagLib::List<TagLib::ByteVector>> = {
            _vptr.List = 0x7ffff7dd46d0 <vtable for TagLib::ByteVectorList+16>, d = 0x614cf0}, d = 0x6149f0}
#16 0x00007ffff79d4f38 in TagLib::ID3v2::TextIdentificationFrame::TextIdentificationFrame (
    this=this@entry=0x614840, data=..., h=<optimized out>)
    at /home/cocobo/repository/taglib/taglib/mpeg/id3v2/frames/textidentificationframe.cpp:254
No locals.
#17 0x00007ffff79d5156 in TagLib::ID3v2::UserTextIdentificationFrame::UserTextIdentificationFrame (this=0x614840, 
    data=..., h=<optimized out>)
    at /home/cocobo/repository/taglib/taglib/mpeg/id3v2/frames/textidentificationframe.cpp:410
No locals.
#18 0x00007ffff795c96b in TagLib::ID3v2::FrameFactory::createFrame (
    this=0x7ffff7dda2b0 <TagLib::ID3v2::FrameFactory::factory>, origData=..., tagHeader=0x612298)
    at /home/cocobo/repository/taglib/taglib/mpeg/id3v2/id3v2framefactory.cpp:174
        f = <optimized out>
        data = {_vptr.ByteVector = 0x7ffff7dd4650 <vtable for TagLib::ByteVector+16>, static null = {
            _vptr.ByteVector = 0x7ffff7dd4650 <vtable for TagLib::ByteVector+16>, 
            static null = <same as static member of an already seen type>, d = 0x610b40}, d = 0x614b70}
        version = <optimized out>
        header = 0x614800
        frameID = {_vptr.ByteVector = 0x7ffff7dd4650 <vtable for TagLib::ByteVector+16>, static null = {
            _vptr.ByteVector = 0x7ffff7dd4650 <vtable for TagLib::ByteVector+16>, 
            static null = <same as static member of an already seen type>, d = 0x610b40}, d = 0x614980}
#19 0x00007ffff7975f09 in TagLib::ID3v2::Tag::parse (this=this@entry=0x612240, origData=...)
    at /home/cocobo/repository/taglib/taglib/mpeg/id3v2/id3v2tag.cpp:700
        frame = <optimized out>
        data = {_vptr.ByteVector = 0x7ffff7dd4650 <vtable for TagLib::ByteVector+16>, static null = {
            _vptr.ByteVector = 0x7ffff7dd4650 <vtable for TagLib::ByteVector+16>, 
            static null = <same as static member of an already seen type>, d = 0x610b40}, d = 0x611c80}
        frameDataPosition = 202
        frameDataLength = <optimized out>
#20 0x00007ffff7976809 in TagLib::ID3v2::Tag::read (this=this@entry=0x612240)
    at /home/cocobo/repository/taglib/taglib/mpeg/id3v2/id3v2tag.cpp:647
No locals.
#21 0x00007ffff7976a45 in TagLib::ID3v2::Tag::Tag (this=0x612240, file=0x611d00, tagOffset=0, 
    factory=0x7ffff7dda2b0 <TagLib::ID3v2::FrameFactory::factory>)
    at /home/cocobo/repository/taglib/taglib/mpeg/id3v2/id3v2tag.cpp:120
No locals.
#22 0x00007ffff794906b in TagLib::MPEG::File::read (this=0x611d00, readProperties=<optimized out>, 
    propertiesStyle=TagLib::AudioProperties::Average) at /home/cocobo/repository/taglib/taglib/mpeg/mpegfile.cpp:489
No locals.
#23 0x00007ffff7949777 in TagLib::MPEG::File::File (this=0x611d00, file=<optimized out>, 
    readProperties=<optimized out>, propertiesStyle=TagLib::AudioProperties::Average)
    at /home/cocobo/repository/taglib/taglib/mpeg/mpegfile.cpp:104
No locals.
#24 0x00007ffff7b9850b in TagLib::FileRef::create (fileName=0x7fffffffde78 "taglib_loop.mp3", 
    readAudioProperties=<optimized out>, audioPropertiesStyle=TagLib::AudioProperties::Average)
    at /home/cocobo/repository/taglib/taglib/fileref.cpp:241
        it = <optimized out>
        ext = {_vptr.String = 0x7ffff7dd4510 <vtable for TagLib::String+16>, static null = {
            _vptr.String = 0x7ffff7dd4510 <vtable for TagLib::String+16>, 
            static null = <same as static member of an already seen type>, 
            static WCharByteOrder = TagLib::String::UTF16LE, d = 0x610af0}, 
          static WCharByteOrder = TagLib::String::UTF16LE, d = 0x611d80}
#25 0x00007ffff7b99f72 in TagLib::FileRef::FileRef (this=0x7fffffffd8c0, fileName=<optimized out>, 
    readAudioProperties=<optimized out>, audioPropertiesStyle=<optimized out>)
    at /home/cocobo/repository/taglib/taglib/fileref.cpp:83
No locals.
#26 0x0000000000401d80 in main (argc=<optimized out>, argv=<optimized out>)
    at /home/cocobo/repository/taglib/examples/tagreader.cpp:41
        f = {_vptr.FileRef = 0x7ffff7dd6690 <vtable for TagLib::FileRef+16>, d = 0x7ffff6d31b60}
        i = <optimized out>
#27 0x00007ffff6d4adc5 in __libc_start_main () from /lib64/libc.so.6
No symbol table info available.
#28 0x0000000000405341 in _start ()
No symbol table info available.
</pre>

TsudaKageyu:
Fixes #466.


sbooth:
According to the ID3 header spec at http://id3.org/id3v2.3.0#ID3v2_header the undefined flags should be cleared so I don't see why this is an unacceptable fix.

lalinsky:
I'd like to know why does the infinite loop happen. Checking for the flags is a workaround that works because the fuzzer flipped also some of the bits in the flags. However, I don't think that is the root cause. An infinite loop typically means that we get some offset/length incorrectly and then keep trying to get more data and don't check that we in fact did not get all the data we asked for. This fix avoids getting into the infinite loop situation, but if the unused flags are cleared and the offset/length is still incorrect, the code will still get there.

TsudaKageyu:
As far as I checked, the direct cause of the infinite loop (or std::bad_alloc on MSVC) is the wrong input for ```Zlib``` functions, and it is caused by the incorrectly set ```Compression``` flag.
In this case, we should check the uncompressed length of the frame and it would be too complicated. So I adopted a somewhat simplified workaround.

lalinsky:
But the workaround only works for this specific file. If you clear the rest of the flags, but keep the compression flag, it will still cause problems.

TsudaKageyu:
Right, it's just a workaround for specific files. But it makes TagLib a little robuster in reality.

sbooth:
Is it possible that this is an issue with zlib that allows the infinite loop?  Granted it is with erroneous input from TagLib but it seems that zlib should be better equipped to detect this condition and handle it gracefully.

TsudaKageyu:
Sorry, I was wrong. The direct cause of the issue is excessive ```frameDataLength```.
At the line 248 of id3v2frame.cpp:
```
  if(d->header->compression() || d->header->dataLengthIndicator()) {
    frameDataLength = SynchData::toUInt(frameData.mid(headerSize, 4));
    frameDataOffset += 4;
  }
```
```frameDataLength``` is set to 2,880,154,539 due to broken data. It leads to an attempt to allocate gigabytes of memory (and very long loop or bad_alloc). 
However, I couldn't find the maximum length of a compressed frame from the ID3v2 spec (it looks like 2.8 gigs is legal), so I don't know how to check it properly.

scotchi:
You could check if the frame length is greater than either the tag length or the file length and assume it's a corrupted header if it is.

TsudaKageyu:
Yes, but the problem is the *uncompressed* size. An uncompressed frame can be larger than the file.
In this case, ```frameDataLength``` is millions of times larger than the file size, so we could assume it's an error. But how about the case that ```frameDataLength``` is two or three times larger than the file size?

scotchi:
Hmm, yes.  Looking at the zlib API, it looks like it might make more sense then to decode the compressed data as a stream (appending it to a growing buffer as necessary) and then just use the value from the frame header as a sanity check.  That would be slightly less performant, but compressed frames are so rare (can't remember if I've ever run across one in the wild) that it's probably more common to run across data corruption than an actual compressed frame.

TsudaKageyu:
Yes. Honestly, I believe it's quite unlikely that a frame is compressed and its uncompressed size is larger than the file. So I think that we could compare ```frameDataLength``` to the file size as a sanity check in practice.
However, I still think that checking if the unused header flags are cleared is not a bad way of checking the header integrity.

scotchi:
Aside from the debate about whether testing the flags makes sense, I fixed the issue of allocating a giant buffer here: caa53e8de5591a75138f90cced544908d016222b

This works both with the unit tests and correctly reads the corrupted file.

TsudaKageyu:
Great! It's obviously a better fix than mine.

TsudaKageyu:
I added some tests for this issue and fixed a compilation error on some compilers at #489.
Hope you merge it.

sbooth:
@lalinsky was right about needing to track down the root cause and I think the fix from @scotchi takes care of it.  These fuzzed file bugs are definitely helping track down the corner cases.

TsudaKageyu:
@scotchi's commit has fixed #486 too. It should be closed.

PR #491, opened 2015-01-05
Implement missing AIFF::File::hasID3v2Tag().

TsudaKageyu:
```hasID3v2Tag()``` is in the header file but doesn't exist in the source file.
This is part of #488, but I decided to create a separate pull request since this is a fatal error and should be fixed immediately.


sbooth:
I guess this isn't a commonly used method.

PR #493, opened 2015-01-06
Simplify overly complicated ByteVector::mid() implementation.

### LINKED ISSUE #121: Large image in an APE tag causes large memory allocations
Hi,

I am reporting an issue with parsing of APE tags. I am adding some pieces of code that I followed to find the problem.

Problem:
I have an MP3 file with ID3 and APE tags in it. The cover art is about 4MB. This is in APE and ID3 tag. When trying to play the file with VLC I get the following errors:

*TagLib: String::prepare() - Unicode conversion error.*

I also monitored the memory usage of VLC when playing this file and it went up to 500MB! Normally it is around 30MB when another MP3 file is being played with only ID3 tag (version not depended).

For normal PC, this is not a big issue as there is plenty of memory, but when playing the same file on embedded device, the player will eat app a lot of memory, which is very limited.

Possible issue:
I looked around the source code of TagLib and the problem occurred when doing the following:

*TagLib::FileRef val(filename)* (filename is the mp3 file)

This one will create a file with

*FileRefPrivate(create(fileName, readAudioProperties, audioPropertiesStyle))*

The *create* method will call 

*MPEG::File(fileName, readAudioProperties, audioPropertiesStyle);*

The constructor will call

*void MPEG::File::read(bool readProperties, Properties::ReadStyle propertiesStyle)*

Inside this method *findAPE()* will be called and then a new APE tag will be created

*d->tag.set(APEIndex, new APE::Tag(this, d->APEFooterLocation));*

The constructor of the APE tag will call *read* method and this one will call *parse(...)*

*parse(d->file->readBlock(d->footer.tagSize() - Footer::size()));*

The *parse* method will loop until the condition is true

>for(uint i = 0; i < d->footer.itemCount() && pos <= data.size() - 11; i++) {
>   APE::Item item;
>   item.parse(data.mid(pos));
>    d->itemListMap.insert(item.key().upper(), item);
>    pos += item.size();
>  }

The *item.parse(...)* will create strings with large sizes (same as the footer).

By looping multiple times, the *d->itemListMap.insert(item.key().upper(), item);* will grow a lot until the end of loop.

From the documentation I saw that APE tag should not be bigger that 8K. JPEG's are suggested to be put as links i.s.o in the tag.

As we know, users will not follow such rules. (ID3 tag could have 256MB of header).

Is it feasible to add a check on the size of the APE tag to not load APE tags bigger that 8K?

If the file is needed for debugging, please let me know. Or you can use MP3Tag program to attach a 5MB image to your MP3s and you will have the same problem.

TsudaKageyu:
Especially remove the useless nested RefCounters.

#### Background

I implemented the copy-on-write operation of ```ByteVector::mid()``` almost two years ago to fix #121 or to speed up parsing some huge tags, and I think it was successful. However, its implementation was overly complicated due to my hastiness and poor coding skill. (especially, it has nested ```RefCounter```s) I have been concerned with it for a long time.

Even though this is focused on simplifying the code, not improving the performance, it makes the ```mid()``` operations slightly faster.

### DIFF
--- a/taglib/toolkit/tbytevector.cpp
+++ b/taglib/toolkit/tbytevector.cpp
@@ -661,8 +661,9 @@ ByteVector::ReverseIterator ByteVector::rbegin()
 
 ByteVector::ConstReverseIterator ByteVector::rbegin() const
 {
-  // we need a const reference to the data vector so we can ensure the const version of rbegin() is called
-  const std::vector<char> &v = d->data->data;
+  // Workaround for the Solaris Studio 12.4 compiler.
+  // We need a const reference to the data vector so we can ensure the const version of rbegin() is called.
+  const std::vector<char> &v = *d->data;
   return v.rbegin() + (v.size() - (d->offset + d->length));
 }
 
@@ -674,8 +675,9 @@ ByteVector::ReverseIterator ByteVector::rend()
 
 ByteVector::ConstReverseIterator ByteVector::rend() const
 {
-  // we need a const reference to the data vector so we can ensure the const version of rbegin() is called
-  const std::vector<char> &v = d->data->data;
+  // Workaround for the Solaris Studio 12.4 compiler.
+  // We need a const reference to the data vector so we can ensure the const version of rbegin() is called.
+  const std::vector<char> &v = *d->data;
   return v.rbegin() + (v.size() - d->offset);
 }
 
--- a/taglib/toolkit/tbytevector.cpp
+++ b/taglib/toolkit/tbytevector.cpp
@@ -451,12 +451,12 @@ ByteVector &ByteVector::setData(const char *data)
 char *ByteVector::data()
 {
   detach();
-  return (size() > 0) ? (&d->data->front() + d->offset) : 0;
+  return (size() > 0) ? (&(*d->data)[d->offset]) : 0;
 }
 
 const char *ByteVector::data() const
 {
-  return (size() > 0) ? (&d->data->front() + d->offset) : 0;
+  return (size() > 0) ? (&(*d->data)[d->offset]) : 0;
 }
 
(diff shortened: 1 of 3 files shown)


taglib/toolkit/tbytevector.cpp
```diff
@@ -49,7 +49,7 @@
 //
 // http://www.informit.com/isapi/product_id~{9C84DAB4-FE6E-49C5-BB0A-FB50331233EA}/content/index.asp
 
-#define DATA(x) (&(x->data->data[0]))
+#define DATA(x) (&(*(x)->data)[0])
```
gogglesguy: Perhaps more readable:

#define DATA(x) (&(x->data->front()))

TsudaKageyu: Or is it better to remove ```DATA``` macro?

gogglesguy: I'd say at least remove them where you simply do:

DATA(d)[someindex]

to

d->data[someindex]

TsudaKageyu:
In the same way of simplifying the complicated implementation of ```ByteVector```, I suggest to use linear search instead of the Knuth-Morris-Pratt algorithm in ```ByteVector::find()```.

The data that ```ByteVector::find()``` handles are limited up to 1024 bytes in almost all cases and the patterns are limited up to 4 bytes. They are too small for KMP to work effectively. As far as I tested, this patch is 40-50% faster than the current implementation when looking 1024-byte data for 4-byte pattern.

TsudaKageyu:
Non-urgent. Close for now.

TsudaKageyu:
I have self-reviewed this patch and now I'm sure it works well.

PR #495, opened 2015-01-06
Add includedir to pkgconfig and taglib-config so that #include <taglib/t...

dfaure:
...ag.h> works.

### DIFF
--- a/taglib-config.cmake
+++ b/taglib-config.cmake
@@ -32,7 +32,7 @@ do
 	  flags="$flags -L$libdir -ltag"
 	  ;;
     --cflags)
-	  flags="$flags -I$includedir/taglib"
+	  flags="$flags -I$includedir -I$includedir/taglib"
 	  ;;
     --version)
 	  echo ${TAGLIB_LIB_MAJOR_VERSION}.${TAGLIB_LIB_MINOR_VERSION}.${TAGLIB_LIB_PATCH_VERSION} 
--- a/taglib-config.cmd.cmake
+++ b/taglib-config.cmd.cmake
@@ -28,7 +28,7 @@ goto theend
   * It would be preferable if the top level CMakeLists.txt provided the library name during config. ??
 :doit
 if /i "%1#" == "--libs#"    echo -L${LIB_INSTALL_DIR} -llibtag
-if /i "%1#" == "--cflags#"  echo -I${INCLUDE_INSTALL_DIR}/taglib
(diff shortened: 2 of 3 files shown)


dfaure:
ping?

lalinsky:
Sorry for the delay. The <taglib/foo.h> include is not really supposed to work. I don't mind changing it in TagLib 2.0 (possibly with keeping the whole structure, not flattening the sub-directories), but I'm not sure if this is a good idea for master.

dfaure:
This is an invasion of the global namespace :-)

#include <tag.h> in any application could be ambiguous, if the application (or some other library invading the global namespace) has a file called tag.h.
It would be much cleaner to be able to use #include <taglib/tag.h>.
Many people are already doing that since it works in a standard distro install where this is in /usr/include/taglib/tag.h
It just breaks in different setups where taglib is installed elsewhere.

People have been trying to work around that breakage by different means:
https://git.reviewboard.kde.org/r/120244/
https://projects.kde.org/projects/extragear/multimedia/plasma-mediacenter/repository/revision?rev=e449b1bcc54dfa3ee3b684f666fdae98a8468617
https://projects.kde.org/projects/extragear/multimedia/plasma-mediacenter/repository/revision?rev=8fba0e62df2277bc23cb4bfa28711c22b33ac78e

This is all quite a mess, because everyone agrees that <taglib/tag.h> is the right way, but it doesn't actually work that way for everyone compiling the code. This commit fixes that.

I can't comment on master vs 2.0, I don't know the branching structure for taglib.
But my commit preserves source compatibility, so I don't see an issue with committing it in "whatever is released next". Code using <tag.h> will still work.

dfaure-kdab:
ping?

lalinsky:
As I said, I really do not want to change this in master and I did not get any other opinions on this. I fully realize that it is an invasion of the global namespace, but it has been like that for years.

mkae:
I am in favour of this change, since avoiding global namespace invasion is the way to go.

gogglesguy:
Given its source compatibility I don't see a need to delay this until 2.0. Who knows when that may be ever released. Then again, kde has been polluting my /usr/include directory for years, so it's a bit funny to see that coming from kde.org...

dfaure-kdab:
gogglesguy: that "pollution" was at least k-prefixed, so not as bad as "tag.h", and is fixed with KDE Frameworks 5.

Issue #482, opened 2014-12-31
Infinite loop with fuzzed file


tholin:
Parsing the following file with the tagreader example results in an infinite loop.
https://www.dropbox.com/s/ly8u5ge3qp0atj8/taglib_wavpack_infloop.wv

```
$ gdb --args ~/repository/taglib/build/examples/tagreader taglib_wavpack_infloop.wv
GNU gdb (Gentoo 7.7.1 p1) 7.7.1
Copyright (C) 2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://bugs.gentoo.org/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /home/cocobo/repository/taglib/build/examples/tagreader...done.
(gdb) r
Starting program: /home/cocobo/repository/taglib/build/examples/tagreader taglib_wavpack_infloop.wv
warning: Could not load shared library symbols for linux-vdso.so.1.
Do you need "set solib-search-path" or "set sysroot"?
******************** "taglib_wavpack_infloop.wv" ********************
TagLib: String::UTF8toUTF16() - Unicode conversion error.
TagLib: String::UTF8toUTF16() - Unicode conversion error.
^C
Program received signal SIGINT, Interrupt.
0x00007ffff7a5bb50 in __gnu_cxx::__normal_iterator<char const*, std::vector<char, std::allocator<char> > >::operator- (this=0x7fffffffd490, __n=@0x7fffffffcb48: 4)
    at /usr/lib/gcc/x86_64-pc-linux-gnu/4.8.3/include/g++-v4/bits/stl_iterator.h:787
787           { return __normal_iterator(_M_current - __n); }
(gdb) bt full
#0  0x00007ffff7a5bb50 in __gnu_cxx::__normal_iterator<char const*, std::vector<char, std::allocator<char> > >::operator- (this=0x7fffffffd490, __n=@0x7fffffffcb48: 4)
[...]

Issue #503, opened 2015-02-15
File length in milliseconds


DidierMalenfant:
I need to get the length of the file in milliseconds.

Most formats let me recalculate the length using sampleFrames() and sampleRate() but mp4 and mpeg don't expose enough info for me to do this externally of the class. The precision seems to be there but it's truncated to seconds during the calculations.

Most of you guys here have more experience with the library than I do, would it be ok to expose more data from those two classes in order to calculate a more precise length?

I could also create a new property lengthInMilliseconds and calculate that inside the class itself but that seems more disruptive.

Thoughts?

sbooth:
This is something I've wanted to bring up as well.  I currently use [mp4v2] (https://code.google.com/p/mp4v2/) in lieu of TagLib for MP4 files because it has a fuller feature set.  TagLib's MP4 support is functional but doesn't have the same kind of support for the format as it does for FLAC or MP3.  This is partly because the MP4 spec is not very open.  I'd like to improve the MP4 support in TagLib to expose not only the type of information that @DidierMalenfant asked about but also to duplicate some of the features that mp4v2 has but TagLib lacks.  This would probably involve making some type of atom class public but I am not sure what a user-friendly interface should look like.

TsudaKageyu:
This is a feature that I have been wanted. I think that implementation itself is not very difficult. As @DidierMalenfant pointed out, the precision seems to be there.
About the interface, I have some ideas but they are confusing with ```length()```. For example:
```
// audioproperties.h

/*!
 * Returns the duration of the file in milliseconds.
 */
virtual int duration() const = 0;
```
So far I don't have an idea which is not confusing with existing ```length()``` function. I'd like to wait for someone to bring us a good idea.

DidierMalenfant:
I'm a big fan of using as few comments as possible, especially in cases like this. My preference would be something like:
```
// audioproperties.h

virtual int lengthInMilliseconds() const = 0;
```
which has the advantage of being clear, self-commenting and self explanatory even in places where the method is called. Saves you having to look up the header to find out what the unit it.

I do understand that this is not the standard way of naming things in taglib so I don't want this to stand out, and also to be available in parallel with the regular ```length()``` which is why I was thinking of exposing ways for the user to calculate the more precise length instead, like ```samplefames()``` in the other classes.

[...]

PR #484, opened 2015-01-01
Fix File::rfind() for small files.

TsudaKageyu:
Required to fix #482.

PR #506, opened 2015-02-17
Support the audio length in milliseconds

TsudaKageyu:
This patch enables TagLib to return the length of audio in milliseconds in addition to seconds.

As suggested in #503, it adds an abstract function like this to ```AudioProperties``` and its derived classes.
```
    /*!
     * Returns the length of the file in milliseconds.
     */
    virtual int lengthInMilliseconds() const = 0;
```
This also enables to calculate the actual average bitrate of Vorbis/Speex files and removes some useless data members from ```AudioProperties``` related classes.


scotchi:
This could only be added to the taglib2 branch since it breaks binary compatibility.

DidierMalenfant:
Maybe we could add one for 1.9.x that only adds the new property?

TsudaKageyu:
Oh! I forgot that adding a virtual function breaks the ABI...
Though it's a little inconvenient, change ```lengthInMilliseconds()``` to non-virtual will fix the problem. Is it OK?

TsudaKageyu:
I confirmed that ABI compliance checker doesn't report any errors now.

TsudaKageyu:
Added support for VBRI headers of MPEG files.

```MPEG::XingHeader``` now supports VBRI headers in addition to Xing headers. It's a little weird, but I couldn't change the name due to ABI compatibility, and hesitant to add a new ```VBRIHeader``` class because it would be almost the same as ```XingHeader```.

TsudaKageyu:
This also fixes the bug reported in #520.

scotchi:
I was just thinking about this now, and I kind of like the idea of specifying a unit a bit more -- what do others think about this?

    enum TimeUnit
    {
      Seconds,
      Milliseconds
    };

    int AudioProperties::length(TimeUnit unit = Seconds) const

This doesn't affect the binary compatibility stuff (it'd actually need to be added without the default argument and then merged in 2.x), but I kind of like the look of that a bit better.

DidierMalenfant:
I'm a big proponent of documenting the code with the code itself instead of comments so this is definitely a good idea in my opinion.

Any reason for using an argument as opposed to creating seperate methods?
````
   int AudioProperties::lengthInSeconds() const
   int AudioProperties::lengthInMilliSeconds() const
````
It reads better and would be slightly more efficient.

If anything, the enum values could be at least made to be ````InSeconds```` and ````InMilliseconds```` again for readability.

scotchi:
I think I'd be fine with that option too -- and deprecating `length()`.  My issue at the moment is that I don't like the inconsistency that `length()` and `lengthInMilliseconds()` introduces in the naming.

TsudaKageyu:
I agree. It's fine to do both -- clarify the meaning of the functions and straighten out the inconsistency.

lalinsky:
I personally prefer the option with deprecating `length()` and adding `lengthInSeconds() ` and `lengthInMilliSeconds()`.

TsudaKageyu:
I prefer it too. I will implement the changes soon.

gogglesguy:
:-1: 

Keep it simple please. Just offer the length in milliseconds only. Why bloat the library with two functions that essentially just give you the same information. And not it's not like as if writing length() / 1000 in the client code is rocket science. And the solution for 1.9.x is even more horrible, having to repeat the lengthInSeconds() in every subclass, while having to keep the deprecated length() as well.

TsudaKageyu:
I see your points. However, it comes from the discussion above.
We can neither remove existing functions and add virtual functions due to binary compatibility.

scotchi:
Also, I think in TagLib 2.x having a function with the same signature but that returns different values would be sub-optimal.

TsudaKageyu:
I am self-reviewing this PR, and I now find it too huge for us to make a thorough review.
So I am thinking of dividing this one into each audio format.

Any thoughts?

sbooth:
That's a good idea, it would definitely simplify the review process.

TsudaKageyu:
Thanks @sbooth, so could you please merge #484?
It will help me improve some audio property implementations and fix #482.

sbooth:
Merged, sorry for the delay.

TsudaKageyu:
No problem, thanks!

PR #511, opened 2015-03-20
Fix a bug that Tag::setProperties() clears the date instead of the track...

TsudaKageyu:
... number.
This bug was reported by Jingbo Zhang.


supermihi:
Whoops ... thanks for the fix!

PR #515, opened 2015-03-22
Skip empty ID3v2 frames when saving an ID3v2 tag.

### LINKED ISSUE #512: Invalid rendering of empty fields in ID3v2
The ID3v2 spec clearly states: `A frame must be at least 1 byte big, excluding the header.`
Frames with a zero length are thus not allowed; this is handled by TagLib when reading ID3v2 frames (although it completely stops parsing of the remaining frames, as reported by others already).

The problem lies here:
``` c++
ByteVector Frame::render() const
{
  ByteVector fieldData = renderFields();
  d->header->setFrameSize(fieldData.size());
  ByteVector headerData = d->header->render();

  return headerData + fieldData;
}
```
When rendering an empty frame, in my case an empty WOAF, renderFields() returns empty data and the frame size is set to zero. As a result, TagLib generates invalid data and is unable to re-read its own output.

Original frame data read from MP3 file:
`57 4F 41 46 00 00 00 01 00 00 00`
Data saved by TagLib back to the file:
`57 4F 41 46 00 00 00 00 00 00`

TsudaKageyu:
This fixes #512.

### DIFF
--- a/taglib/mpeg/id3v2/id3v2frame.cpp
+++ b/taglib/mpeg/id3v2/id3v2frame.cpp
@@ -193,6 +193,9 @@ void Frame::setText(const String &)
 ByteVector Frame::render() const
 {
   ByteVector fieldData = renderFields();
+  if(fieldData.isEmpty())
+    fieldData = ByteVector("\x00", 1);
+
   d->header->setFrameSize(fieldData.size());
   ByteVector headerData = d->header->render();
 
--- a/taglib/mpeg/id3v2/id3v2frame.cpp
+++ b/taglib/mpeg/id3v2/id3v2frame.cpp
@@ -193,9 +193,6 @@ void Frame::setText(const String &)
 ByteVector Frame::render() const
 {
   ByteVector fieldData = renderFields();
-  if(fieldData.isEmpty())
-    fieldData = ByteVector("\x00", 1);
-
   d->header->setFrameSize(fieldData.size());
   ByteVector headerData = d->header->render();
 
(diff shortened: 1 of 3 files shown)


sbooth:
It makes sense to stick to the spec so I think this change is good.  However, is actually rendering a frame like this useful?  Would it make more sense to change `ID3v2Tag::render(int)` to skip frames that are empty?

lalinsky:
I think we should definitely not add frames like this. How can you be sure that a single null character is valid content for all frame types? I'm pretty sure it would make some frame types invalid and cause problems with parsing them. We would just print a warning for debugging purposes and skip the frame.

TsudaKageyu:
@lalinsky, your point seems reasonable. I will modify this one soon.

EvilSquirrel:
@lalinsky Skipping empty frames is a reasonable approach, but the spec does seem pretty clear: `A frame must be at least 1 byte big, excluding the header.` This is found in section 3.3 of the spec, which describes frames in general, thus it should be valid for all frame types.

TsudaKageyu:
@EvilSquirrel and @lalinsky, both your points make sense. However, I think that skipping empty frames is like going on the safe side.

lalinsky:
@EvilSquirrel Every frame should have some data, but some frames need more than one byte and having just a single null character would make them invalid. For example, if you add empty `APIC` to the tag and this code would add a single null character as the data, it would be invalid because a valid APIC frame needs at least 4 bytes in case there is absolutely no content.

EvilSquirrel:
@lalinsky You may well be right - as @TsudaKageyu says skipping the frame is probably the safer option.

TsudaKageyu:
Added a test as pointed out by @scotchi at #512.

PR #521, opened 2015-04-17
Fix a GCC/Clang warning about singed/unsigned comparison.

TsudaKageyu:
This is definitely duplicate with #464, but it looks like he won't return...


gogglesguy:
:+1:

PR #522, opened 2015-04-17
Avoid memmove across different buffers in ByteVector::replace()

TsudaKageyu:
detach() is called inside ByteVector::data().

### DIFF
--- a/taglib/toolkit/tbytevector.cpp
+++ b/taglib/toolkit/tbytevector.cpp
@@ -629,17 +629,19 @@ ByteVector &ByteVector::replace(const ByteVector &pattern, const ByteVector &wit
       break;
 
     if(diff < 0) {
+      char *pData = data();
       ::memmove(
-        data() + offset + withSize,
-        data() + offset + patternSize,
+        pData + offset + withSize,
+        pData + offset + patternSize,
         size() - offset - patternSize);
       resize(size() + diff);
     }
     else if(diff > 0) {
       resize(size() + diff);
+      char *pData = data();
       ::memmove(
-        data() + offset + withSize,
-        data() + offset + patternSize,
+        pData + offset + withSize,
+        pData + offset + patternSize,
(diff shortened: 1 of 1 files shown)


lalinsky:
Can you please explain why this is not necessary?

TsudaKageyu:
Non-const version of ```ByteVector::data()``` calls ```detach()``` internally and the first parameter of ```memmove()/memcpy()``` is non-const.

lalinsky:
I don't think that the code with `memmove()` is correct like this. For example, the first `data()` call returns the current pointer because there is only one reference, then some other thread adds a reference, so the second `data()` call has to actually detach and it returns a different pointer. Now you are moving across two different data pointers.

TsudaKageyu:
Hmm, but I think calling ```detach()``` first won't solve it. We have to do, for example, like this.
```
    if(diff < 0) {
      char *p = data();
      ::memmove(
        p + offset + withSize,
        p + offset + patternSize,
        size() - offset - patternSize);
      resize(size() + diff);
    }
    else if(diff > 0) {
      resize(size() + diff);
      char *p = data();
      ::memmove(
        p + offset + withSize,
        p + offset + patternSize,
        size() - diff - offset - patternSize);
    }
```

lalinsky:
Well, the first detach makes sure there is only one reference and no other code knows about the newly created reference, so currently it should work, but it's not very clear.

TsudaKageyu:
I think the new change makes things a bit clearer.

gogglesguy:
I guess all bets are off when you're talking multithreaded. In the single threaded case, just a single call to detach outside the loop would have been sufficient. I believe TsudaKageyu is correct: the detach() won't work if another thread comes by and references this vector again. However, I'd be really afraid if this would actually happen, since after all we're in the middle of modifying this vector. Are we aiming for a complete multithread safe bytevector?

Anyway, the new changes look fine to me.

PR #527, opened 2015-04-28
Fix saving WMA files with some GUID fields.

### LINKED ISSUE #524: WMA file with GUID information generates corrupted files
We have a wma file here and wondered why the file was corrupt after storing it with new information. After some investigation we found out that the GUID information (e.g. **WM/MediaClassPrimaryID, WM/MediaClassSecondaryID, ...**) will be read correctly, but if it comes to write the data to file, the file will get corrupt. Only the VLC player can play those audio files after that.

Our workaround for now is to iterate all information of the map and remove every information of the type GUID. After that everything is fine.

TsudaKageyu:
This (looks like) fixes #524.

After playing with WMP for some time, I got an idea that some GUID fields like ```WM/MediaClassPrimaryID```, ```WM/WMCollectionID```... should belong to MetadataLibraryObject. It looks like the bug has been fixed, but I'm not sure this solution is robust enough.

Any thoughts?

### DIFF
--- a/taglib/asf/asffile.cpp
+++ b/taglib/asf/asffile.cpp
@@ -56,17 +56,33 @@ public:
   ASF::File::MetadataLibraryObject *metadataLibraryObject;
 };
 
-static ByteVector headerGuid("\x30\x26\xB2\x75\x8E\x66\xCF\x11\xA6\xD9\x00\xAA\x00\x62\xCE\x6C", 16);
-static ByteVector filePropertiesGuid("\xA1\xDC\xAB\x8C\x47\xA9\xCF\x11\x8E\xE4\x00\xC0\x0C\x20\x53\x65", 16);
-static ByteVector streamPropertiesGuid("\x91\x07\xDC\xB7\xB7\xA9\xCF\x11\x8E\xE6\x00\xC0\x0C\x20\x53\x65", 16);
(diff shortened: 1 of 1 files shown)


TsudaKageyu:
I found Microsoft's official document concerning the ASF spec:
http://download.microsoft.com/download/8/0/5/80506BEB-C95A-47AE-99CF-0D6D6D028ABA/ASF_Specification.pdf

And found the description below in section 7.4:
> **Extended  Content  Description  Object.**  This  object  stores  non-language-specific, 
non-stream-specific attributes of any name for the following types: WCHAR strings, BYTE 
arrays,  Boolean  values,  DWORDs,  QWORDs,  or  WORDs.  Values  are  limited  to  65535
bytes in length.

> **Metadata Library Object.**  This object stores attributes for any name. It allows data to 
be stored in the same types as the Extended Content Description Object as well as a 
GUID  data  type.  Attributes  can  be  stream-specific  and  are  associated  with  a  particular 
language. Values can exceed the 65535-byte limit imposed by the other attribute-storage 
objects because value length is stored in a DWORD.

So I changed the code to store any GUID fields in Metadata Library Object like large sized values.

bromix:
Confirmed...the wma files are working with this PR. Thanks! :+1:

lalinsky:
Thank you for fixing this.

Issue #526, opened 2015-04-27
FileRef: Using IOStream instead of a file name 


chouquette:
Hi,

I'm trying to add taglib support for VLC on WinRT, which has a few "fun" limitations with regards to file.
Basically, we can't fopen the file, which obviously lead to TagLib's parsing failure.

One way to fix this would be to feed taglib using the IOStream interface, which is fairly trivial. The issue lies in using this IOStream afterward. The FileRef class doesn't provide a method for this. So far the ways I see are:
- To not use FileRef, and reimplement the file extension factory somewhere else, which also drops the ability to use potential FileTypeResolver 
- Extend the FileRef class with a create method that takes an IOStream*, however, in order to use FileTypeResolver with an IOStream, we need to add another method to the interface, which would (most likely) break the ABI. I assume this to be a problem :)
- Extend the FileRef class but not expose FileTypeResolver for files accessed through IOStream

I'd be happy to provide a pull request, though I'd appreciate any insight/advices on how to achieve this!

Thanks in advance,

lalinsky:
Well, the only added value of `FileRef` and `FileTypeResolver` is filename-based type guessing. A `IOStream` subclass still needs to return some kind of filename. So my suggestion would be to add another constructor to `FileRef` that accepts a `IOStream` pointer and uses `stream->filename()` to find the right type.

chouquette:
Wouldn't that be an issue with regard to the part of FileRef::create() using FileTypeResolver ? (point 2 above)

lalinsky:
Ah, I knew I'm forgetting something. You are right, we can't add a new virtual method there. Well, for now we can probably just not use `FileTypeResolver` in `master` and add only support for that in the `taglib2` branch.

chouquette:
That's fine by me. I'll try to propose something this week.
Thanks for your answers!

scotchi:
Sorry to have jerked you around a bit here, but after reading through this, I'm not sure that I agree with Lucas that this patch makes sense.  It feels like a lot of extra code to support a special case in one project.  I'm not convinced that it makes sense to support using a FileRef in cases where the file can't be opened.

If the block of code that would need to be copied into your project were ginormous or especially tricky, it'd make sense to me, but at present, it's basically just a map from extensions to namespaces.  (Which honestly, shouldn't generally be used in major applications anyway since you'd want to do mime type detection.)

[...]

PR #528, opened 2015-04-28
FileRef: Allow an IOStream to be used

chouquette:
Refs issue #526 

I haven't tested it on WinRT (which is the main reason for this, but it works fine on linux)
I'm submitting this for early review and will post the WinRT results as soon as I have a build & it's integrated.

Cheers,

### DIFF
--- a/taglib/fileref.cpp
+++ b/taglib/fileref.cpp
@@ -54,6 +54,10 @@
 
 using namespace TagLib;
 
+template <typename T>
+static File* createInternal(T arg, const String& filename, bool readAudioProperties,
+                            AudioProperties::ReadStyle audioPropertiesStyle);
+
 class FileRef::FileRefPrivate : public RefCounter
 {
 public:
@@ -81,6 +85,17 @@ FileRef::FileRef(FileName fileName, bool readAudioProperties,
                  AudioProperties::ReadStyle audioPropertiesStyle)
 {
   d = new FileRefPrivate(create(fileName, readAudioProperties, audioPropertiesStyle));
+}
+
+FileRef::FileRef(IOStream* stream, bool readAudioProperties, AudioProperties::ReadStyle audioPropertiesStyle)
+{
+  d = new FileRefPrivate(createInternal(stream, 
+#ifdef _WIN32
+  stream->name().toString()
+#else
+  stream->name()
+#endif
+  , readAudioProperties, audioPropertiesStyle));
 }
 
 FileRef::FileRef(File *file)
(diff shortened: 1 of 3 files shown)


taglib/fileref.cpp
```diff
@@ -284,3 +289,77 @@ File *FileRef::create(FileName fileName, bool readAudioProperties,
...
+      const int pos = s.rfind(".");
+      if(pos != -1)
+        ext = s.substr(pos + 1).upper();
+    }
+
+    // If this list is updated, the method defaultFileExtensions() should also be
+    // updated.  However at some point that list should be created at the same time
+    // that a default file type resolver is created.
+
+    if(!ext.isEmpty()) {
```
lalinsky: Please do not make a copy of this whole block. Create an internal template function instead.

chouquette:
Updated with a template function to avoid duplicating code

taglib/fileref.cpp
```diff
@@ -54,6 +54,62 @@
 
 using namespace TagLib;
 
+template <typename T>
+static File* createInternal(T arg, const String& ext, bool readAudioProperties,
```
lalinsky: The extension extraction code could be here as well. Just pass it `FileName fileName` instead of `String ext`.

lalinsky:
Can you please add at least one test of this new API to `tests/test_fileref.cpp`?

taglib/fileref.cpp
chouquette: I thought of that, but it makes me pass filename twice for one version, and iostream, filename for the other... That feels weird, so I refrained myself.
There's a way to guess the filename based on the type that is provided, but I'm not sure that would improve readability...
If you're fine with passing filename twice I'll add it

lalinsky: I'm fine with passing the filename twice, of course. That's why I suggested it. :)

chouquette: Just wanted to check we had the same understanding :)
I just pushed that change as well, and updated the tests so all tests are also using the IOStream constructor.

chouquette:
I forgot to mention it before, but this works fine for VLC on WinRT :)

taglib/fileref.cpp
```diff
@@ -213,74 +278,22 @@ File *FileRef::create(FileName fileName, bool readAudioProperties,
...
-    if(ext == "XM")
-      return new XM::File(fileName, readAudioProperties, audioPropertiesStyle);
-  }
-
-  return 0;
+File *FileRef::create(IOStream* ioStream, bool readAudioProperties, AudioProperties::ReadStyle audioPropertiesStyle)
+{
+  return createInternal(ioStream,
+#ifdef _WIN32
+  ioStream->name().toString();
```
scotchi: This almost certainly is wrong.  `;` without a closing paren?

taglib/fileref.cpp
```diff
@@ -54,6 +54,67 @@
 
 using namespace TagLib;
 
+template <typename T>
+static File* createInternal(T arg, const String& filename, bool readAudioProperties,
```
scotchi: TagLib aligns `*` and `&` to the right, so `static File *create...`

taglib/fileref.cpp
```diff
@@ -54,6 +54,67 @@
...
+                            AudioProperties::ReadStyle audioPropertiesStyle)
+{
+    String ext;
```
scotchi: Please use 2 space indents.

TsudaKageyu:
@chouquette, sorry for making you wait for a long time. This PR has been merged now.

PR #533, opened 2015-05-01
Fix the wrong padding of ByteVector::resize().

TsudaKageyu:
The expanded area will be filled with garbage instead of correct padding in some corner cases.

Related to #531, I found another corner case bug of ```ByteVector```. It won't affect badly as far as I tested, but bug is bug.

### DIFF
--- a/taglib/toolkit/tbytevector.cpp
+++ b/taglib/toolkit/tbytevector.cpp
@@ -702,6 +702,7 @@ ByteVector &ByteVector::resize(uint size, char padding)
 {
   if(size != d->length) {
     detach();
+    d->data->data.resize(d->offset + d->length);
     d->data->data.resize(d->offset + size, padding);
     d->length = size;
   }
--- a/taglib/toolkit/tbytevector.cpp
+++ b/taglib/toolkit/tbytevector.cpp
@@ -703,11 +703,15 @@ ByteVector &ByteVector::resize(uint size, char padding)
   if(size != d->length) {
     detach();
 
-    if(size > d->data->data.size() - d->offset)
-      d->data->data.resize(d->offset + size);
+    const size_t bufferSize = d->data->data.size();
 
-    if(size > d->length)
-      ::memset(DATA(d) + d->offset + d->length, padding, size - d->length);
+    if(size > bufferSize - d->offset) {
+      d->data->data.resize(d->offset + size, padding);
+      ::memset(&*end(), padding, bufferSize - (d->length + d->offset));
+    }
+    else if(size > d->length) {
+      ::memset(&*end(), padding, size - d->length);
+    }
 
     d->length = size;
   }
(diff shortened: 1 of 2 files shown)


lalinsky:
I don't get this, what does resizing the vector twice fix?

TsudaKageyu:
```
ByteVector k = ByteVector("0123456789").mid(3, 4);
```
After this line, ```k``` is like ```"3456"``` but the internal buffer is like ```"0123456789"```, ```d->offset == 3``` and ```d->length == 4```. So if we do ```k.resize(6, 'A')```, the internal buffer will be like...
```
// Current
d->data->data.resize(d->offset + size, padding); // "012345678"
``` 
```
// Fixed
d->data->data.resize(d->offset + d->length);     // "0123456"
d->data->data.resize(d->offset + size, padding); // "0123456AA"
```

gogglesguy:
Calling resize twice is extremely ugly imho. So this only happens in the cases where offset>0 and refcount == 1. Aka this works:

```
TagLib::ByteVector a = TagLib::ByteVector("0123456789").mid(3,4).resize(6,'A');
```
But this doesn't work:

```
TagLib::ByteVector a = TagLib::ByteVector("0123456789").mid(3,4);
a.resize(6,'A')
```
All this abstraction really hides the fact that we have a simple situation of a 10 byte buffer, where the only thing we really want to do is turn '78' into 'AA'. So perhaps instead of a resize(), a simple memset could do the trick here. Another option would be to have detach() replace the original data with a 0 offset data (in cases offset>0 and refcount==1). Technically you'd still end up with 2 (r)allocations, but at least any of unused bytes in front of the internal buffer are released. Something the proposed patch doesn't do.

TsudaKageyu:
I agree with you that using ```vector::resize()``` twice is ugly. I was hesitant to write code almost the same as ```vector::resize()``` myself... 
However, as you say, it's definitely inefficient. I will rewrite it.

gogglesguy:
So the last patch doesn't reduce the memory ever. However for taglib's case,  I think this may be fine . Some of the string conversions (UTF8) will allocate larger buffers and reduce it afterwards, or the truncation in IDV1 tags, but in all cases these are short-lived buffers and I generally think it's just a waste to even bother resizing the buffers since many will go away very shortly after. So I'd be ok with this change.

gogglesguy:
I think you could further optimize the memset to only touch the bytes that weren't already set by data.resize()

TsudaKageyu:
I see your point. Will try it.

gogglesguy:
As you've pointed out that st::vector::resize doesn't actually reallocate any memory, perhaps the original patch is cleaner after all. Just add a giant comment explaining the behavior.

TsudaKageyu:
Yes, if I added some appropriate comment, it wouldn't waste your time...

PR #534, opened 2015-05-09
Avoid duplicate version information.

ufleisch:
Instead of having the version information both in CMakeLists.txt and
taglib/toolkit/taglib.h, it is now generated from CMake into a header file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,9 +52,20 @@ if (MSVC AND ENABLE_STATIC_RUNTIME)
   endforeach(flag_var)
 endif()
 
-set(TAGLIB_LIB_MAJOR_VERSION "1")
-set(TAGLIB_LIB_MINOR_VERSION "9")
-set(TAGLIB_LIB_PATCH_VERSION "1")
+# Read version information from file taglib/toolkit/taglib.h into variables
+# TAGLIB_LIB_MAJOR_VERSION, TAGLIB_LIB_MINOR_VERSION, TAGLIB_LIB_PATCH_VERSION.
+foreach(version_part MAJOR MINOR PATCH)
+  set(version_var_name "TAGLIB_${version_part}_VERSION")
+  file(STRINGS taglib/toolkit/taglib.h version_line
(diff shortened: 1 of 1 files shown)


sbooth:
I like the idea of eliminating redundancy but my concern with this pull request is that if people don't use CMake as a build system their library won't have any embedded version info.

bromix:
I agree with @sbooth :+1:
We also try to avoid such mechanism. We for example have a build system besides CMake and could have a problem with this PR.

ufleisch:
Would you prefer if the version information is only stored in taglib.h and read by CMake to avoid duplication?

bromix:
From my point of view (and if possible) yes. Because the header is the at least primitive system (information) we can absolutely count on - even if we change the build system (e.g. premake, nmake and so on).

TsudaKageyu:
I like it @ufleisch.

ufleisch:
I've changed the commit 	accordingly.

FestusHagen:
To make it clear since I am frequently misunderstood ...

I offered this snippet of a different technique just as another technique no better no worse ... 

Maybe my use of the word 'suggest' should be changed 'offer'!

If I offended anybody ... Please accept my apology for I never intend to offend anyone!

-Enjoy

ufleisch:
The purpose of this pull request is to improve TagLib's maintainability. If the version information is stored in two places, it is easy to forget one (as it happened in the 1.8 release). If the version information is changed in taglib.h and then the project is built, the version information must be updated automatically in all dependent files (taglib.pc, taglib_c.pc, taglib-config, taglib-config.cmd). Your suggestion fails to do this, a manual cmake rerun would be necessary. Besides this, it seems more complicated than my proposed commit. If you find an error or an improvement, I will gladly fix my commit. The current offer, however, is in my judgment not helpful and only stalls the processing of this pull request.

FestusHagen:
I didn't offer it as a replacement to your patch, If I did I would of created a pull request, it is a snippet of a different technique for reading/parsing the version from file ... That is all, no more!

This is not holding up this pull request.

-Enjoy

PR #537, opened 2015-05-15
Bytevector truncate

gogglesguy:
Get rid of the additional buffer resize() in String.data(String::UTF8) by doing a cheap truncate on the ByteVector. The returned bytevector is mostly likely a short lived transitional buffer anyhow.

### DIFF
--- a/taglib/toolkit/tbytevector.cpp
+++ b/taglib/toolkit/tbytevector.cpp
@@ -709,6 +709,16 @@ ByteVector &ByteVector::resize(uint size, char padding)
   return *this;
 }
 
+ByteVector &ByteVector::truncate(uint offset)
+{
+  if (offset<d->length) {
+    detach();
+    d->length = offset;
+  }
+
+  return *this;
+}
+
 ByteVector::Iterator ByteVector::begin()
 {
   detach();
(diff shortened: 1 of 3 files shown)


TsudaKageyu:
It looks like PR #533 achieves the same thing, doesn't it?

gogglesguy:
Yes, it would. However, PR #533 currently removes any chance of reducing the buffer to a smaller size. Not sure if this is desired.

TsudaKageyu:
AFAIK, ```std::vector::resize()``` doesn't reduce the actual buffer when decreasing the size. So ```ByteVector::resize()``` doesn't reduce the buffer even if #533 is merged.
If we need to reduce a buffer, we have to add a function works like ```std::vector::shrink_to_fit()```.

gogglesguy:
You're probably right, although that may be implementation dependent regarding what happens when you call std::vector::resize(). I think the new resize() is really what we want, so no need for this then.

TsudaKageyu:
So can you suggest something about #533?

TsudaKageyu:
FYR, according to C++ reference (http://www.cplusplus.com/reference/vector/vector/resize/):
> In case the container shrinks, all iterators, pointers and references to elements that have not been removed remain valid after the resize and refer to the same elements they were referring to before the call.

It means ```std::vector::resize()``` can't reallocate memory when shrinking.

gogglesguy:
Yeah, looks like you are correct. I may have been over concerned.

PR #538, opened 2015-05-16
 bytevector.mid(pos,4).toUInt() => bytevector.toUInt(pos)

gogglesguy:
Replace use of bytevector.mid(pos,4).toUInt() with more optimized bytevector.toUInt(pos)


TsudaKageyu:
Looks good. But could you please stick to our coding style?
Specifically, please have a white space before the second parameter like:
```
data.toUInt(pos, true);
```
Additionally, don't have a space between ```if``` and a brace, and have spaces around an operator like:
```
if(offset < d->length) {
```

gogglesguy:
Sure, I was trying to :)

TsudaKageyu:
:+1:

Issue #336, opened 2013-12-13
Cover art support for Xiph comments


sbooth:
This would apply to Ogg Vorbis, Speex, Opus, and FLAC.

http://mail.kde.org/pipermail/taglib-devel/2013-December/002609.html
http://wiki.xiph.org/VorbisComment#Cover_art

This wouldn't entail too much work other than adding a base 64 encoder/decoder and the accompanying support in `Ogg::File` and friends.

FestusHagen:
#540 closes this?

-Enjoy

TsudaKageyu:
You're right. Thanks!

PR #540, opened 2015-05-16
Implement XiphComment coverart support #336

gogglesguy:
This patch set implements FLAC::Picture support for xiphcomments. Not only makes it a lot easier on the end user, it also saves memory since we no longer have to keep the base64 encoded strings (a freaking 4 bytes per character on Linux) in memory and by decoding it you lose the overhead of base64 as well.

As requested by ticket #336


TsudaKageyu:
I think it's better to make ```ByteVector::fromBase64()``` static like other ```from*``` functions for API consistency.

gogglesguy:
Yeah, I think you're right. entry.mid().fromBase64() will create an extra copy of the data anyway, so you'll lose the in-memory decoding advantage.

taglib/ogg/xiphcomment.cpp
```diff
@@ -26,17 +26,22 @@
...
 #include <tdebug.h>
 
+#include <flacpicture.h>
 #include <xiphcomment.h>
 #include <tpropertymap.h>
 
 using namespace TagLib;
 
+
+typedef List<FLAC::Picture*> PictureList;
```
lalinsky: There should be a space between the class name and `*`.

taglib/ogg/xiphcomment.cpp
```diff
@@ -363,20 +408,41 @@ void Ogg::XiphComment::parse(const ByteVector &data)
...
-    pos += commentLength;
-    if(pos > data.size()) {
-      break;
-    }
+    ByteVector entry = data.mid(pos, commentLength);
 
-    int commentSeparatorPosition = comment.find("=");
-    if(commentSeparatorPosition == -1) {
+    // Don't go past data end
+    pos+=commentLength;
```
lalinsky: Spaces around operators. There are many more cases of this in the patch.

taglib/ogg/xiphcomment.cpp
```diff
@@ -363,20 +408,41 @@ void Ogg::XiphComment::parse(const ByteVector &data)
...
+    // Handle Pictures separately
+    if(entry.startsWith("METADATA_BLOCK_PICTURE=")) {
+
+      // Decode base64 picture data
+      ByteVector picturedata = ByteVector::fromBase64(entry.mid(23));
+
+      if(picturedata.size()==0) {
+        debug("Empty picture data. Discarding content");
+        continue;
+        }
```
lalinsky: This looks indented incorrectly.

lalinsky:
There are many style differences from what we use in TagLib. Please have a look at the existing code and use the same style.

taglib/toolkit/tbytevector.h
```diff
@@ -567,6 +567,16 @@ namespace TagLib {
      */
     ByteVector toHex() const;
 
+    /*!
+     * Returns a base64 encoded copy of the byte vector
+     */
+    ByteVector toBase64() const;
```
lalinsky: We should have many explicit tests for the base64 code, both encoding and decoding.

gogglesguy:
Thanks for the feedback. Bear with me while I get used to the taglib coding style. I'll add some more tests as well.

gogglesguy:
So suppose there's some invalid base64 encoded data, or at least something the base64 decoder can't handle. Right now we'd be throwing that data away and on f->save() it would disappear from the file itself. 

Another way to handle bad data is to store them as key/values as before. They won't be showing up in the pictureList() but they will be preserved (albeit in a different order) when f->save() is called.

Thoughts?

FestusHagen:
Data loss and I don't get along ...

-Enjoy

TsudaKageyu:
It would be considered a fatal bug when some data is gone even if the data is corrupt. We'd better leave corrupted data untouched.

gogglesguy:
Consider the current xiphcomment text implementation, if it's missing a '=' in the string (yay corrupted data), its content will be ignored and upon render() it wont be written back to the file. Same if the key or content happens to be empty. So it isn't such a black and white situation as it may appear.

TsudaKageyu:
Yes. Either removing or writing corrupt data would be considered a bug... We should choose one of them and it's better not to change the current behavior.
Please forget what I said above.

lalinsky:
I think that all fields should end up in the field list like before. Developers should be able to access the raw data. Cover art support should operate on the existing field list and just read/modify it. It should not "steal" fields from it. METADATA_BLOCK_PICTURE is just a recommendation and COVERART might contain just about anything.

scotchi:
As a style related note, I have an .astylerc file that gets very close to TagLib's style that I've run over the sources before and it's caught a lot of stuff.  I'd like to at some point run that over the sources and commit the output and ideally add a post-commit hook to check stuff in the future.  (Out of `indent`, `clang-format` and `astyle`, `astyle` was the one that had the smallest number of false positives on source checks.)  The time to do that though would probably be just before or after a release so that it won't make doing reviews more difficult.  (Though now that we're using git, you can get a long way by doing blames / diffs that ignore whitespace changes.)

FestusHagen:
On the Style side ...

@scotchi think I could get you to post or email it to me?
I have been working on a CONTRIBUTING doc and it would be awesome to use it for examples.
Yahoo.com ... festushagen2002.

On the cleanup side ...
I would love it and am very willing to do the work, testing, ... if we could zap all the white space and style inconsistencies at once and be done with them!
It is very hard to review pull requests with simple changes and dozens of style / white space changes.

I believe it should be done immediately following a release before any merges.

Thanks
-Enjoy

scotchi:
I just added the style file to the root of the repo.  The changes still need to be hand reviewed (I actually did that for the whole repo, but the change set is huge, so I'll just keep the patch sitting around for later), but it still is useful for checking obvious mess-ups.

gogglesguy:
Latest patch will preserve any picture data we can't decode in the field list.

Issue #48, opened 2012-08-21
O_CLOEXEC support (if available)


gogglesguy:
Would be nice if taglib can open its files with O_CLOEXEC on supported platforms (linux / glibc 2.7). This is extremely usefull for multithreaded programs which can fork at a moments notice...

For implementation details:
http://udrepper.livejournal.com/20407.html
Also see "man fopen"

PR #541, opened 2015-05-17
Use O_CLOEXEC on supported glibc

gogglesguy:
See issue #48


lalinsky:
To be honest, I do not see why we should do this. It's just more platform-specific code in TagLib to maintain. If application developers want some non-standard way of opening files, they can provide their own `IOStream` instance.

gogglesguy:
Oh right, I had forgotten you can now provide your own upstream. Never mind then.

PR #543, opened 2015-05-17
Replace cmake ENABLE_STATIC with more generic BUILD_SHARED_LIBS

gogglesguy:
This replaces the #444 pull request (which had a typo). In addition it builds a shared library by default, something the previous patch didn't do.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules)
 
 option(BUILD_SHARED_LIBS "Build as shared library"  ON)
 
-if(DEFINED ENABLE_STATIC)
+if(ENABLE_STATIC)
   message(FATAL_ERROR "ENABLE_STATIC no longer supported. Use BUILD_SHARED_LIBS to control linkage")
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,12 +8,14 @@ endif()
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules)
 
-option(ENABLE_STATIC "Make static version of libtag"  OFF)
-if(ENABLE_STATIC)
+option(BUILD_SHARED_LIBS "Build as shared library"  ON)
+
+if(DEFINED ENABLE_STATIC)
+  message(FATAL_ERROR "ENABLE_STATIC no longer supported. Use BUILD_SHARED_LIBS to control linkage")
+endif()
+
+if(NOT BUILD_SHARED_LIBS)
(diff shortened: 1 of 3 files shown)


gogglesguy:
I believe @lalinsky prefered just one or the other. Then again it would be hard to support both, because which one would take precedence? Wouldn't the majority of distributions build shared libraries anyway?

I did make a small change that allows -DENABLE_STATIC=OFF to still work.

TsudaKageyu:
You should revise ```INSTALL``` as well.
However, I think this is a kind of compatibility breakages and such changes should be made in TagLib2.0.

FestusHagen:
First ... PLEASE DON'T LET MY POOR WRITING SKILLS OFFEND YOU!
Everything I say comes from the heart ... No offense meant!

@gogglesguy, History shows @lalinsky is the owner of
```
option(ENABLE_STATIC "Make static version of libtag" OFF)
if(ENABLE_STATIC)
add_definitions(-DTAGLIB_STATIC)
set(BUILD_SHARED_LIBS OFF)
else()
set(BUILD_SHARED_LIBS ON)
endif()
```
I'll not speak for @lalinsky

I agree with @TsudaKageyu this is a change for 2.x not 1.x.

I am not judge jury and executioner but I cannot agree with this change for 1.x.

-Enjoy

gogglesguy:
So what's the next version 1.9.2 or 1.10? Because if it's 1.10 this kind of "compatibility" is a bit over the top imho.  We already changed the cmake version requirement in master. Perhaps we shouldn't do that either, because we break compatibility. Just promising binary compatibility between stable releases should be enough.

Besides, package maintainers usually do have to change at least the version number in the build scripts, so it's not like building a new package of taglib is completely "free". And considering the cmake version requirements, it may also require updating the build dependencies. What's one more extra flag to change?

gogglesguy:
@TsudaKageyu I did

TsudaKageyu:
I remember that I changed ```cmake_minimum_required``` value, but it was just a follow up of an actual requirement. TagLib had actually been required CMake2.8 before I changed ```cmake_minimum_required```.

Also I wonder if ```BUILD_SHARED_LIBS``` will really help somebody (maintainer of Linux packages?).

gogglesguy:
Not sure. I'm definitely not a cmake expert. Just fixed up the #444 pull request. Don't shoot the messenger :P

TsudaKageyu:
Hmm, it might be better to ask whether ```BUILD_SHARED_LIBS``` is really preferable in the mailing list.

FestusHagen:
One could also argue that using project specific variables (ie: ENABLE_STATIC) and setting CMake variables with them is the fact that when CMake changes the variable name it does not break the public interface ...

The CMake crowd has a history doing so!

-Enjoy

TsudaKageyu:
I merged an old pull request #444 equivalent to this.
Sorry for closing this one after bothering you for a long time.

Issue #344, opened 2014-02-10
Complete quoting for parameters of some CMake commands


elfring:
Some parameters (like "[${CMAKE_CURRENT_BINARY_DIR}/taglib.pc](https://github.com/taglib/taglib/blob/2f7af42092cefac12536ce7ecf5bc519bb2e24fa/CMakeLists.txt#L79)" and "[${CMAKE_CURRENT_SOURCE_DIR}/toolkit](https://github.com/taglib/taglib/blob/a6f759cc9a8e2529ab15fc41790621b2af658ec0/taglib/CMakeLists.txt#L2)") are passed to CMake commands in your build scripts without enclosing them by quotation marks. I see that these places will result in build difficulties if the contents of the used variables will contain special characters like spaces.

I would recommend to apply [advices from a Wiki article](http://cmake.org/Wiki/CMake/Language_Syntax#CMake_splits_arguments_unless_you_use_quotation_marks_or_escapes. "CMake splits arguments unless you use quotation marks or escapes.").

elfring:
Are there [a few update candidates](https://github.com/taglib/taglib/blob/b509106031e6ae3675ad9601e5ae8bf5f3844109/tests/CMakeLists.txt#L1 "Update candidate") left over?

TsudaKageyu:
It looks like we have tens of candidates in some files.

PR #551, opened 2015-05-20
Quote path names including user-provided variables.

TsudaKageyu:
As @elfring pointed out at #344.


FestusHagen:
Quoting is obviously a hot topic in many circles, this is my personal conclusion after my own rigorous testing ...

__*Quote them when you create them, then quote them no more!*__

Or modify them ...

-Enjoy

elfring:
Would you like to distinguish if a variable manages a single string or several list items in your build scripts?

PR #554, opened 2015-05-20
Don't look for CppUnit unless BUILD_TESTS, general cleanup.

FestusHagen:


TsudaKageyu:
Looks good and works well.

Issue #360, opened 2014-03-30
Missing AudioProperties::bitDepth() and sampleFormat()


risingballs:
Two essential audio properties cannot be queried by TagLib:
- bit depth (8, 16, 32, 64)
- sample format (pcm, float)

It would be highly appreciated to have these available.

kfieldi5:
I would love to see these as well (bitdepth moreso than sample format)

TsudaKageyu:
I think I can add this feature (at least bit depth) to #506 and it wouldn't be so hard.

TsudaKageyu:
Currently, some classes have their own "bit depth" features in different names. For example:
```
APE::Properties::bitsPerSample(); // APE, MP4, WavPack
AIFF::Properties::sampleWidth();  // AIFF, FLAC, WAV
```
So I suggest that we choose one of them and mark other one deprecated.

TsudaKageyu:
I took a brief look at this issue.

AIFF, APE, FLAC, MP4, TrueAudio, WAV and WavPack classes already have the feature.
MPC, MPEG, Opus, Speex, Vorbis and MOD related formats don't have the bit depth information.
ASF is the only format we can actually add the feature.

I added ```bitsPerSample()``` property to ```ASF::Properties```. But I wonder if we need to add it to other classes.

Any thougts?

TsudaKageyu:
Additionally, I already added ```WAV::Properties::format()``` to get the sample format.

kfieldi5:
Very interesting! Thanks Tsuda! Can I still just use
TagLib::RIFF::AIFF:Properties for wav though?

PR #556, opened 2015-05-21
(wishlist) ASF: AudioProperties improvements

TsudaKageyu:
This replaces a part of #506.

Add bitsPerSample() property. (#360)


gogglesguy:
Does the bitsPerSample even mean anything in asf? Can we distinguish between a lossy codec vs lossless codec?

TsudaKageyu:
WMA supports lossless encoding. How about adding a property to check if the file is lossless?

gogglesguy:
Having a codec() property like mp4 would be helpful. If its lossy I'm interested in the bitrate while with lossless files I tend more to care about the bitspersample.

TsudaKageyu:
According to the document (https://www.microsoft.com/en-us/download/details.aspx?id=14995), we can detect five kinds of codecs. Namely:
* Windows Media Audio
* Windows Media Audio 9 Professional
* Windows Media Audio 9 Lossless
* GSM-AMR (Fixed bit rate, no SID)
* GSM-AMR (Variable bit rate, including SID)

But does somebody need to distinguish them completely?

TsudaKageyu:
Hell yeah! Now we can get the codec name and encoder settings in String like "Windows Media Audio 9.1" and "VBR Quality 100, 44 kHz, 2 channel 16 bit 1-pass VBR".

TsudaKageyu:
Hi @gogglesguy, how about ```codecName()``` and ```codecDescription()``` functions?

taglib/asf/asffile.cpp
```diff
@@ -365,6 +386,61 @@ ByteVector ASF::File::HeaderExtensionObject::render(ASF::File *file)
...
+    const int descLength = data.toUShort(pos, false);
+    pos += 2;
+
+    const uint descPos = pos;
+    pos += descLength * 2;
+
+    const int infoLength = data.toUShort(pos, false);
+    pos += 2 + infoLength * 2;
+
+    if(type == 2) {
```
gogglesguy: Magic Number, maybe use enum/const for readability?

taglib/asf/asffile.h
```diff
@@ -134,6 +134,7 @@ namespace TagLib {
       class ContentDescriptionObject;
       class ExtendedContentDescriptionObject;
       class HeaderExtensionObject;
+      class CodecListObject;
```
gogglesguy: So why do these forward declarations even need to be here? Kind of defeats the purpose of hiding everything under FilePrivate.

taglib/asf/asffile.cpp
TsudaKageyu: Sounds reasonable. I'll do it soon.

gogglesguy:
Does codecName() provide anything extra that you couldn't figure out from codec()? I guess codecDescription() could be useful for some people, assuming it contains stuff like the input settings of the encoder.

TsudaKageyu:
For example, when ```codec()``` returns ```WMA2```, ```codecName()``` may vary like Windows Media Audio 7, 8, 9, 9.1, 9.2, and so forth.

taglib/asf/asffile.h
TsudaKageyu: I think we can hide them, and it won't affect the binary compatibility.
However, it will be a huge refactoring and should be another try.

TsudaKageyu: I realized it's not so complicated. I just moved the forward declarations to ```FilePrivate```, and it works fine.
Actually, we can do that in this pull request.

TsudaKageyu:
I hid some very "under the hood" things from the public header as @gogglesguy pointed out.

PR #565, opened 2015-05-21
(wishlist) WAV: AudioProperties improvements

TsudaKageyu:
This replaces a part of #506.

Add bitsPerSample() property besides sampleWidth(). (#360)
Add format() property. (#360)
Read the number of samples from "fact" chunk if available.


taglib/riff/wav/wavproperties.h
```diff
@@ -52,35 +52,107 @@ namespace TagLib {
...
         Properties(const ByteVector &data, uint streamLength, ReadStyle style);
 
         /*!
+         * Create an instance of WAV::Properties with the data read from the
+         * ByteVector \a data and the length calculated using \a streamLength
+         * and \a totalSamples.
+         *
+         * \note totalSamples can be zero if the file doesn't have a 'fact' chunk.
+         */
+        Properties(const ByteVector &data, uint streamLength, uint totalSamples, ReadStyle style);
```
scotchi: Hmm.  I'm not really excited about adding more constructors with [magic parameters](http://doc.qt.digia.com/qq/qq13-apis.html#theconveniencetrap) here.  I realize I'm responsible for a lot of them in the code, but that's mostly in code that I wrote more than 10 years ago.  In new code (and in TagLib 2.x) I'd prefer to move to something more along the lines of:

    Properties &setTotalSamples(uint samples);

Since there you can chain these things like:

    Properties p = Properties(data).setStreamLength(length).setTotalSamples(samples);

scotchi: Looking at this more closely, another option is creating a thin wrapper type that we require usage of here, e.g.:

    class SampleLength
    {
    public:
      explicit SampleLength(int length);
      operator int() const;
      // ...
    };

That would probably make sense if this is a pattern we see across most of the file types.

TsudaKageyu: I have considered another option that we create a new constructor like this:
```
Properties(File *file, ReadStyle style);
```
and move some code from ```WAV::File``` to ```WAV::Properties```. 
It looks straightforward, but it requires to add a ```friend class``` to ```RIFF::File``` and I don't like abusing ```friend```.

TsudaKageyu: Come to think of it, it's common to use ```friend``` and pass ```File*``` to a constructor of ```Properties```.
So I'd like to change the constructor to take ```WAV::File*```.

lalinsky:
:+1:

Issue #520, opened 2015-04-08
TrueAudio(.tta)  bitrate "arithmetic overflow"


billlin0904:
/*
d->streamLength = 503215768;
d->length = 4337;

((503215768 * 8) / 4337) / 1000
*/

d->bitrate = d->length > 0 ? ((d->streamLength * 8L) / d->length) / 1000 : 0;

d->bitrate => 0xFFFFFFC2 (-62)

So... 

My solution below ... : )

d->bitrate = d->length > 0 ? (((long long)d->streamLength * 8L) / d->length) / 1000 : 0;

TsudaKageyu:
I think that changing ```8L``` to ```8LL``` is a simpler solution.
Actually, pull request #506 has already fixed it by calculating it in ```double```. However, it's a big change to review and merge soon...

billlin0904:
Awesome ~

TsudaKageyu:
Is it better to fix those possible arithmetic overflows apart from #506? Any thoughts?

lalinsky:
No, I'll try to merge the pull request soon. Sorry for not getting to it sooner.

PR #566, opened 2015-05-21
(wishlist) TrueAudio: AudioProperties improvements

TsudaKageyu:
This replaces a part of #506.

Avoid possible arithmetic overflows. (#520)
Remove some data members which are not needed to carry.


lalinsky:
:+1:

Issue #92, opened 2013-01-01
WavPack files with more than 2 channels are recognized as having only 2


mineo:
The WavPack format supports more than 2 channels but taglib claims there are only two even when the audio has 6.

TsudaKageyu:
I can support multichannel WavPack immediately, but it requires #484 first.

FestusHagen:
#567 Closes this?

-Enjoy

TsudaKageyu:
The .wav (dts) issue is still there. But I'm not sure if we should support it.

ufleisch:
Looks like this is no longer the case:

```
$ ffmpeg -f lavfi -i "sine=frequency=440:duration=5" -ac 6 -c:a wavpack output.wv
(..)
$ examples/tagreader output.wv 
******************** "output.wv" ********************
-- TAG (basic) --
title   - ""
artist  - ""
album   - ""
year    - "0"
comment - ""
track   - "0"
genre   - ""
-- TAG (properties) --
ENCODER - "Lavf61.7.102"
-- AUDIO --
bitrate     - 245
sample rate - 44100
channels    - 6
length      - 0:05


$ ffmpeg -f lavfi -i "sine=frequency=440:duration=5" -ac 6 -c:a dca -strict -2 output_dts.wav
(..)
$ examples/tagreader output_dts.wav 
******************** "output_dts.wav" ********************
-- TAG (basic) --
title   - ""
artist  - ""
album   - ""
year    - "0"
comment - ""
track   - "0"
genre   - ""
-- TAG (properties) --
ENCODING - "Lavf61.7.102"
-- AUDIO --
bitrate     - 1412
sample rate - 44100
channels    - 6
length      - 0:05
```

Please reopen if I am wrong.

PR #567, opened 2015-05-21
(wishlist) WavPack: AudioProperties improvements

TsudaKageyu:
This replaces a part of #506.

Add isLossless() property.
Support multi channel. (#92)
Avoid infinite loop when reading a fuzzed file. (#482)
Remove some data members which are not needed to carry.


lalinsky:
:+1:

PR #568, opened 2015-05-21
(wishlist) MOD: AudioProperties improvements

TsudaKageyu:
Add lengthInSeconds(), lengthInMilliseconds() properties. (always return 0)

### DIFF
--- a/taglib/it/itproperties.cpp
+++ b/taglib/it/itproperties.cpp
@@ -79,6 +79,16 @@ int IT::Properties::length() const
   return 0;
 }
 
+int IT::Properties::lengthInSeconds() const
+{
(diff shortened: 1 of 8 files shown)


scotchi:
I realize that this wasn't previously implemented, but is it really not possible to get a length out of a MOD file?  (I know it's a set of instructions for synthesizer playback, but it seems like that should still have a nominal length?)

TsudaKageyu:
As far as I checked the spec, a MOD file doesn't have its length in its header.
http://www.aes.id.au/modformat.html
I think it requires to parse an entire file to calculate the length.

lalinsky:
:+1:

PR #571, opened 2015-05-22
Avoid an infinite loop when reading fuzzed WavPack files. (#482)

TsudaKageyu:
That's a fatal bug and deserves to fix immediately.


TsudaKageyu:
Thanks, all the reported bugs related to broken files are fixed now.

PR #573, opened 2015-05-24
Use iterators rather than indices to loop through a List.

TsudaKageyu:
Make use of List::setAutoDelete().


TsudaKageyu:
Not urgent and will conflict with other ones.

PR #575, opened 2015-05-24
Stop calculating the offset in RIFF::File::chunkData().

TsudaKageyu:


sbooth:
This looks good to me but I'd like for @lalinsky or @scotchi to weigh in before I merge the PR.  It appears that `offset` is calculated when the chunk is read so the code is equivalent but I'm wondering if there is a reason it wasn't written this way initially.

TsudaKageyu:
Yes, I think so.

TsudaKageyu:
Non-urgent. Close for now.

TsudaKageyu:
I checked the history of ```rifffile.cpp``` again. And I now think it's just an omission in  a2aaf8a8e17372caef3388cb06a88d65f7be52ea.

PR #579, opened 2015-05-27
Skip both ID3v1 and APE tags when seeking the last MPEG frame.

TsudaKageyu:
Related to #577. I missed out on the fact that we can also skip an APE tag.


TsudaKageyu:
Merged into #601.

TsudaKageyu:
I merged this one since it's part of a bug fix already merged, and a very trivial change.

PR #580, opened 2015-05-27
Fix inconsistent negative seek behavior between Linux and Windows.

TsudaKageyu:
Not an urgent bug. It will affect in so unlikely cases that actually do no harm. For example, seeking an ID3v1 tag in a file smaller than 128 bytes.


TsudaKageyu:
Oops! I forgot to describe the bug itself.

For example, when we do a negative seek like this:
```
seek(100, Begin);
seek(-200, Current);
```
The file pointer doesn't move on Linux, but moves to 0 on Windows.

TsudaKageyu:
Non-urgent. Close for now.

PR #582, opened 2015-05-28
Remove config.h requirement.

FestusHagen:
Surround `#include "config.h"` with `#ifdef HAVE_CONFIG_H`

### DIFF
--- a/taglib/mp4/mp4item.cpp
+++ b/taglib/mp4/mp4item.cpp
@@ -23,7 +23,9 @@
  *   http://www.mozilla.org/MPL/                                           *
  ***************************************************************************/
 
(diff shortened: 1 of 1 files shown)


TsudaKageyu:
I think we can remove ```#include "config.h"``` itself from ```mp4item.cpp```.
It had to be removed in 116b1fcff9b686333fdd6d2607c6781e03d708a1.

FestusHagen:
Looking into this further shows `config.h` is included in `tutils.h`, `tutils.h` is included in `mp4item.cpp` thus it is a duplicate and can be removed.

Thanks @TsudaKageyu for pointing it out!

PR #583, opened 2015-05-29
Resolved tests build failures on Win32 MinGW using shared libs.

FestusHagen:
I believe this would break ABI on master

### DIFF
--- a/taglib/mp4/mp4atom.h
+++ b/taglib/mp4/mp4atom.h
@@ -74,7 +74,7 @@ namespace TagLib {
 
     typedef TagLib::List<AtomData> AtomDataList;
 
-    class Atom
+    class TAGLIB_EXPORT Atom
     {
     public:
         Atom(File *file);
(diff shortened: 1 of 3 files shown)


scotchi:
Hmm, I'm actually reasonably sure this would be fine on master.  For the compilers that need this these symbols would have been hidden anyway, so there's no change, and for the compilers that ignore it, this would also be no change.  Anyone else know better off the top of their heads?

FestusHagen:
That decision is above my pay grade ... And yes, I did raise it some in the process!

In master `toolkit/tstringlist.h` operator `<<` needs to be exported as well.
```TAGLIB_EXPORT std::ostream &operator<<(std::ostream &s, const TagLib::StringList &l);```

I tried to get abi-compliance-checker working, too much dead meat goes along with it!

I would be much happy to redo it for master!
In fact I'll just go brew it, upload it, and let y'all with higher pay grades test it out!

FestusHagen:
Moved to master ...

Issue #576, opened 2015-05-24
Multiple Xiph comments in a FLAC file


TsudaKageyu:
If a FLAC file has multiple Xiph comments, TagLib only reads the first one. However, it overwrites all Xiph comments with the same contents.
I wonder if it is a consistent way of handling multiple tags. MPEG, AIFF or WAV can also have duplicate tags, and they just ignore duplicate ones. I think we should leave duplicate tags untouched as for FLAC files too.

gogglesguy:
According to the FLAC format: "There may be only one VORBIS_COMMENT block in a stream". 

Keeping multiple xiph comments in a stream could result in unpredictable behaviour. If you update the xiphcomments using taglib, but a third party client may read a different xiph comment block (or vice versa). I don't think the MPEG,AIFF, WAV method is necessary the best way to go. I think we should remove redundant vorbiscomment blocks and keep the first one found (most likely the one read by other software).

sbooth:
I agree with @gogglesguy that FLAC files with more than one vorbis comment block should be handled as noncompliant.  TagLib definitely shouldn't write noncompliant files; I think the best course of action is to coalesce the multiple blocks into one to avoid data loss.

TsudaKageyu:
I agree with @sbooth that we should stick to the specifications. However...

As for FLAC, the specfication clearly states that VORBIS_COMMENT cannot be redundant. So removing redundant tags would be compliant. However, the specifications of MPEG, AIFF or WAV don't mention ID3v2 tags and the specification of ID3v2 doesn't clarify how a tag should be stored in each audio file as far as I know.

scotchi:
I think there are times where we don't have to have something spelled out in the spec to decide what's reasonable and unreasonable behavior.  In the particular case of multiple ID3v2 tags in the same file, I personally think it's pretty clear that those are created by buggy software and should be treated as such.

Dumping ID3v2 into AIFF and WAV is a bit of a weird thing to do, but for AIFF I just copied what Apple does since it's an Apple format.  I found some other software that was doing that for WAV and decided to copy the behavior there as well.  If we were to find multiple tags in multiple differently named RIFF chunks, then I think the issue of merging them could be relevant, but is that actually something we've seen in the wild enough to bother worrying about it?  Unless we're actually seeing something like this happen more than in some extremely rare cases, I'd say it's safe enough to just throw out the second tag.

gogglesguy:
Yes, what @scotchi says, is this actually a real world problem?

[...]

PR #584, opened 2015-05-29
Skip duplicate ID3v2 tags by taking them into the first one as padding.

TsudaKageyu:
Related to the discussion at #576. This patch will erase duplicate ID3v2 tags by treating them as a part of the first one's padding.

Any thoughts?


TsudaKageyu:
I just became aware of the fact that ```ID3v2::Tag::render()``` may render a data larger than 256MB and it will break a file. I think it worth it to fix.
Which is better to fix it here or another place?

TsudaKageyu:
Merged into #601.

PR #586, opened 2015-05-29
Skip duplicate comment blocks in FLAC files.

TsudaKageyu:
Related to #576.

Duplicate tags will be treated like padding blocks.
This leads to erasing duplicate tags when saving.

### DIFF
--- a/taglib/flac/flacfile.cpp
+++ b/taglib/flac/flacfile.cpp
@@ -197,33 +197,21 @@ bool FLAC::File::save()
 
   // Replace metadata blocks
 
-  bool foundVorbisCommentBlock = false;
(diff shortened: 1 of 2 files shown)


gogglesguy:
Looks good.

TsudaKageyu:
I found that ```paddingLength``` can be greater than 2^24 and it will break a file. It is extremely unlikely but possible and worth it to check.

TsudaKageyu:
Merged into #596 to avoid conflict.

TsudaKageyu:
@gogglesguy, sorry for closing this after your review. But it's better to do at the same time with fixing #606.

PR #587, opened 2015-05-29
Fill string buffers with zero before using snprintf().

TsudaKageyu:
```snprintf()``` doesn't necessarily terminate a string properly.
http://www.dwheeler.com/secure-programs/Secure-Programs-HOWTO/dangers-c.html

### DIFF
--- a/taglib/toolkit/tutils.h
+++ b/taglib/toolkit/tutils.h
@@ -160,7 +160,7 @@ namespace TagLib
       va_list args;
       va_start(args, format);
 
-      char buf[BufferSize];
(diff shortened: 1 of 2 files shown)


scotchi:
This doesn't actually fix the issue.  The issue is that if you do:

    char foo[6];
    snprintf(foo, sizeof(foo), "foobar");

Some implementations will have afterwards `fooba\0` and some will have `foobar`.  Initializing the array to zeros doesn't change that.  I thought about that when adding it, but since it's only in the tests where we have predictable input (and OS X and Linux libcs do the right thing), I didn't worry about it.

PR #589, opened 2015-05-29
Resolved tests failure on Win32 MinGW using shared libs.

FestusHagen:
I just realized I forgot to boot a box with MSVS and test this, @TsudaKageyu you are a native MSVS user?, if you are, would you please test, Thank you!

-Enjoy

### DIFF
--- a/taglib/mp4/mp4atom.h
+++ b/taglib/mp4/mp4atom.h
@@ -74,7 +74,7 @@ namespace TagLib {
 
     typedef TagLib::List<AtomData> AtomDataList;
 
-    class Atom
+    class TAGLIB_EXPORT Atom
     {
     public:
         Atom(File *file);
@@ -92,7 +92,7 @@ namespace TagLib {
     };
 
     //! Root-level atoms
-    class Atoms
+    class TAGLIB_EXPORT Atoms
     {
     public:
         Atoms(File *file);
@@ -109,3 +109,4 @@ namespace TagLib {
 #endif
 
 #endif
+
--- a/taglib/toolkit/tstringlist.h
+++ b/taglib/toolkit/tstringlist.h
@@ -110,6 +110,6 @@ namespace TagLib {
  * \related TagLib::StringList
  * Send the StringList to an output stream.
  */
-std::ostream &operator<<(std::ostream &s, const TagLib::StringList &l);
+TAGLIB_EXPORT std::ostream &operator<<(std::ostream &s, const TagLib::StringList &l);
 
 #endif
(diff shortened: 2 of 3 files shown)


TsudaKageyu:
Sure. All the tests works well on MSVC. No errors nor warnings. :+1: 

By the way, I wonder if ```MP4::Atom``` is designed to be public. 
```mp4atom.h``` has a comment: "This file is not part of the public API!". Additionally, the implementation of ```MP4::Atom``` is different from other classes. It doesn't adopt the d-pointer idiom (a pattern like ```class XxxPrivate; XxxPrivate *d;```) and it implies the class is not designed to be public.

I barely remember @lalinsky mentioned something about that, but I can't recall what he said.

**Comment by FestusHagen** [comment]
Yea, I see the comment now ...

I am confused, it must already be visible in at least FreeBSD that I know of (I suspect all Unix based OS's), the tests work with shared or static library.

All the `__declspec(dllexport)` does it make it visible in Windows ... As it already is in Unix based OS's.

 To far above my C++ skills!

**Comment by lalinsky** [comment]
The tests should not use the shared library, but unfortunately at the moment they do, so if you have a DLL build, you can't really run the tests properly. The proper fix would be to change the build system to either build an utility static library that can be used in the tests and then used to build the DLL or to build the tests completely separately.

**Comment by lalinsky** [comment]
Sorry about closing it, I see that @scotchi was considering merging the other patch, so I guess it would be ok.

**Comment by scotchi** [comment]
I didn't actually look much at the patch or what part of the API it was touching -- I was mostly just responding to the notion of if it was possible to export non-exported symbols without breaking the ABI.  If the classes aren't meant to be exported, then naturally they shouldn't be.

**Comment by lalinsky** [comment]
The original intention was that the classes are private, but maybe it's not worth it and this fix might be the best thing to do.

**Comment by FestusHagen** [comment]
Not questioning or arguing ... Don't know enough about the subject to do so!

I would like to know why the tests should not use a shared library or cannot be run properly on a DLL build and have been unsuccessful on finding a reason, it appears to be normal to do so.

Any pointers to info regarding this matter would be much appreciated!

Like y'all, All I want is whats best for TagLib ... If it should not be done, then we should change the build system to not build tests when building a shared library as well as making note of this in the docs.

**Comment by lalinsky** [comment]
The tests test more than the public TagLib API exposes. They also test internal classes that DLL users shouldn't have access to. Your change works this around by making the private APIs public.

**Comment by FestusHagen** [comment]
That makes perfect sense. Thank you for the explanation @lalinsky!

**Comment by scotchi** [comment]
Yeah, after looking back at this, this isn't the right way to fix this issue.  If it's determined at some point that it'd be valuable to be able to run the tests with a shared lib, then it'd probably make sense to either `#ifdef` these tests out, or have a special shared library build target that exposes them.  I'm not sure what we'd really gain by that though.

**Comment by FestusHagen** [comment]
Nor do I, so should we do something like:
```
if(BUILD_TESTS AND NOT BUILD_SHARED_LIBS)
  enable_testing()
  add_subdirectory(tests)
endif()
```

PR #590, opened 2015-06-01
Reduce CMake checks for the byte order.

TsudaKageyu:
This patch removes byte order checks from CMake checks.

I realized that we don't have to do them by CMake since modern compilers are smart enough to do them at compile time.
I tested this patch on x86/x86_64, ARM (Raspberry Pi 2) and PowerPC (QEMU) systems.


TsudaKageyu:
Though I added those checks, I now wonder if GLIBC's ```__bswap_*```, Mac's ```OSSwapInt*()``` and OpenBSD's ```swap*()``` are really necessary.
As far as I know, GCC, Clang and Intel Compiler supports ```__builtin_bswap*()``` and MSVC has its own built-in functions. Isn't it enough to check these built-in functions?

TsudaKageyu:
Non-urgent. Close for now.

TsudaKageyu:
I close this since it will heavily conflict with the current master.

PR #592, opened 2015-06-05
Implementation for reading and writing Acidized loop information to a WAV file.

FestusHagen:
This is the work of @rupertdaniel not I, It's been almost a year, it appears to be abandoned, thought I could fix it~~, I did~~.

If this is wrong PLEASE TELL ME!

### DIFF
--- a/taglib/riff/wav/acidinfo.cpp
+++ b/taglib/riff/wav/acidinfo.cpp
@@ -1,0 +1,187 @@
+/***************************************************************************
+ copyright            : (C) 2012 by Rupert Daniel
+ email                : rupert@cancelmonday.com
+ ***************************************************************************/
+
+/***************************************************************************
+ *   This library is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU Lesser General Public License version   *
+ *   2.1 as published by the Free Software Foundation.                     *
+ *                                                                         *
+ *   This library is distributed in the hope that it will be useful, but   *
+ *   WITHOUT ANY WARRANTY; without even the implied warranty of            *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU     *
+ *   Lesser General Public License for more details.                       *
+ *                                                                         *
+ *   You should have received a copy of the GNU Lesser General Public      *
+ *   License along with this library; if not, write to the Free Software   *
+ *   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA         *
+ *   02110-1301  USA                                                       *
+ *                                                                         *
+ *   Alternatively, this file is available under the Mozilla Public        *
+ *   License Version 1.1.  You may obtain a copy of the License at         *
+ *   http://www.mozilla.org/MPL/                                           *
+ ***************************************************************************/
+
+#include "acidinfo.h"
+
+#include <tstring.h>
+#include <tbytevector.h>
+#include <tdebug.h>
+
+using namespace TagLib;
+
+class RIFF::WAV::AcidInfo::AcidInfoPrivate
+{
+public:
+  AcidInfoPrivate() :
+    flags(0),
+    rootNote(0),
+    unknown1(0x8000u),
+    unknown2(0.0),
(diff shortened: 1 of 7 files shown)


taglib/riff/wav/acidinfo.cpp
```diff
@@ -0,0 +1,187 @@
...
+  return data;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// private members
+////////////////////////////////////////////////////////////////////////////////
+
+void RIFF::WAV::AcidInfo::read(const ByteVector &data)
+{
+  d->flags = data.mid(0, 4).toUInt(false);
```
TsudaKageyu: It's more efficient to do ```data.toUInt(0, false)``` rather than using ```mid()```. Following ```to*```s can be replaced.

taglib/riff/wav/wavfile.h
```diff
@@ -126,6 +127,11 @@ namespace TagLib {
         Info::Tag *InfoTag() const;
 
         /*!
+         * Returns the Acid loop information for this file
+         */
+        AcidInfo *acidInfo() const;
```
TsudaKageyu: It looks like this function lacks its implementation.

taglib/riff/wav/acidinfo.cpp
FestusHagen: Thank you, Thank you very much ... Thy shall be done.

taglib/riff/wav/wavfile.cpp
```diff
@@ -47,7 +48,8 @@ class RIFF::WAV::File::FilePrivate
     properties(0),
     tagChunkID("ID3 "),
     hasID3v2(false),
-    hasInfo(false)
+    hasInfo(false),
+    acidInfo(NULL)
```
TsudaKageyu: I wonder if ```NULL``` is compatible with all compilers. I think it's better to initialize it with zero like other fields.

taglib/riff/wav/wavfile.h
FestusHagen: Bummer, I was unaware it was incomplete ... I may have bitten off more then I can chew on this one!

taglib/riff/wav/wavfile.cpp
FestusHagen: That I did not know was an issue ... Thank you.
I did test on MinGW, MSVC2010, FreeBSD-8.2 (GCC 4.2.1).

I will follow your suggestion and change it to Zero.

TsudaKageyu:
I think we should add a test handling an actual acidized WAV file. Do you have some sample file @FestusHagen?

FestusHagen:
Agreed, an actual test file is needed. No I do not have one that can be included.

I grabbed a free sample (and royalty free) off the web, I do not know the legalities of it so was not going to include it. I am still looking for something that is clearly legal to include ...

scotchi:
Is Acid Loop metadata used by any software other than Acid?  Acid is basically dead these days and isn't a major player in the sequencer market.  If this is just their proprietary format, I'm not convinced it's useful to add to TagLib.

taglib/riff/wav/wavfile.cpp
scotchi: It's actually a style issue, not a compatibility issue.  As a general rule of thumb, C uses `NULL`, C++ uses `0` and C++11 uses `nullptr`.

FestusHagen:
Your call @scotchi ...
I'm just trying to move old PR's and Issues along to closure, be it merged or dumped!

scotchi:
From the cursory bit that I could find, I searched to see if any of the top sequencers could actually read this metadata, and of Ableton Live, Cubase, Logic, FL Studio, Protools and Sonar, Sonar seems to be the only one that's able to.  Since ACID hasn't released a new version in 5 years and has significantly waned in popularity, unless there's some information that's missing here (i.e. that this metadata is more widely supported), I don't think this makes sense to merge.  I'm open to the idea of reopening this one though if it turns out that the assumptions above are incorrect.

FestusHagen:
@scotchi, Should #396 be closed as well?

PR #593, opened 2015-06-08
Fix checking for RIFF chunk names.

TsudaKageyu:
The check overlooks characters larger than 0x7f when char is signed.
I already fixed this one in ```taglib2``` branch but forgot to do it in ```master```.
 
The test log below implies that this bug actually occurs on OSX. I expect this PR fixes those warnings as well.
http://build.oxygene.sk/job/taglib-osx/58/console

### DIFF
--- a/taglib/riff/rifffile.cpp
+++ b/taglib/riff/rifffile.cpp
@@ -67,6 +67,20 @@ RIFF::File::~File()
   delete d;
 }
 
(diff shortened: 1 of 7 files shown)


TsudaKageyu:
Resolved a conflict with current ```master```.

TsudaKageyu:
Non-urgent. Close for now.

PR #596, opened 2015-06-10
Avoid corrupting files when saving them twice.

TsudaKageyu:
Oops! I realized that TagLib itself may write duplicate ID3v2 tags when saving an MPEG file.

### DIFF
--- a/taglib/asf/asffile.cpp
+++ b/taglib/asf/asffile.cpp
@@ -23,6 +23,8 @@
  *   http://www.mozilla.org/MPL/                                           *
  ***************************************************************************/
 
+#include <climits>
+
 #include <tdebug.h>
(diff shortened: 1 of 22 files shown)


TsudaKageyu:
I found some other formats (at least FLAC and ASF) has similar bugs. I'll check them soon.

TsudaKageyu:
Hmm, this issue is more serious than I expected. I added some tests calls ```File::save()``` twice, and the result is terrible. ASF, FLAC, MP4, Ogg Vorbis, corrupt files when calling ```save()``` twice. Ogg FLAC and Ogg Opus may cause a segfault.

I think we definitely should fix them all.

TsudaKageyu:
I did it but this PR has become too large... I close this and create separate PRs.

PR #598, opened 2015-06-11
Avoid corrupting a file when calling MP4::File::save() twice without …

TsudaKageyu:
…closing it.

This tries to keep the data structure of ```MP4::Tag``` up-to-date to avoid creating the same atom twice.


TsudaKageyu:
Sorry for disturbing you @lalinsky, but I think you are the most appropriate person to check if this measure is correct.

PR #604, opened 2015-06-11
Add tests calling TrueAudio::File::save() twice.

TsudaKageyu:
```TrueAudio::File``` doesn't corrupt a file when calling ```save()``` twice.
Added some tests.


TsudaKageyu:
In fact, ```TrueAudio::File``` sometimes fails to move an ID3v1 tag correctly. Fixed it.

PR #608, opened 2015-06-12
Add a method to check if an MP4 file on disk actually has a tag.

TsudaKageyu:
We can check if an MP4 file on disk actually has a tag. I was totally wrong in #252. Sorry @joelverhagen!


TsudaKageyu:
Non-urgent. Close for now.

TsudaKageyu:
@lalinsky, though I don't want to merge this soon, I'd like to ask you if this PR will work.
Basically, this checks if the "ilst" atom exists.

PR #612, opened 2015-06-18
Avoid overwriting the audio stream when adding an ID3v1 tag to a FLAC…

TsudaKageyu:
… file.

```FLAC::File``` overwrites the last 128 bytes of the audio stream when creating an ID3v1 tag.
Actually, this is part of #600, but I think it's an urgent bug, so I created a small PR to make it able to be reviewed and merged quickly.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -36,6 +36,7 @@ TagLib 1.10 (??? ??, 2015)
  * New API for the audio length in milliseconds.
  * Marked FLAC::File::streamInfoData() deprecated. It returns an empty ByteVector.
  * Marked FLAC::File::streamLength() deprecated. It returns zero.
+ * Fixed possible file corruptions when adding an ID3v1 tag to FLAC files.
  * Many smaller bug fixes and performance improvements.
 
 TagLib 1.9.1 (Oct 8, 2013)
--- a/taglib/flac/flacfile.cpp
+++ b/taglib/flac/flacfile.cpp
@@ -258,8 +258,16 @@ bool FLAC::File::save()
   }
 
   if(ID3v1Tag()) {
-    seek(-128, End);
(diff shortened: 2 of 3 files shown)


TsudaKageyu:
FLAC has adopted Vorbis Comments since v1.0.3 released after a year since v1.0. They needed something to fill in the blank.

TsudaKageyu:
Well, it would be useful for some people if ```FLAC::File``` has ```strip()``` function like APE or MusePak.

gogglesguy:
I'd say that it should automatically convert the idv1 tags to vorbiscomments by default. Since that's the officially supported tag anyhow (and discard id3v1 if vorbiscomments already exists).

TsudaKageyu:
I agree. IMHO, it's totally useless to keep ID3 tags in a FLAC file. But I'm not sure we may remove them without asking.

TsudaKageyu:
I found the following answer in FLAC's official FAQ: https://xiph.org/flac/faq.html

> FLAC has it's own native tagging system which is identical to that of Vorbis. They are called alternately "FLAC tags" and "Vorbis comments". It is the only tagging system required and guaranteed to be supported by FLAC implementations.
Out of convenience, the reference decoder knows how to skip ID3 tags so that they don't interfere with decoding. But you should not expect any tags beside FLAC tags to be supported in applications; some implementations may not even be able to decode a FLAC file with ID3 tags.

Following the description, it would be the safest option to remove ID3 tags automatically as @gogglesguy suggested.

TsudaKageyu:
I reported several similar bugs concerning the ```save()``` functions, but this one is somewhat special.
Since this one will occur on the first call to the function, unlike all the other bugs occur when calling them twice,

Issue #613, opened 2015-06-18
Allow generating taglib.pc on Windows


noscript:
Due to this condition ```taglib.pc``` is not generated:
```
if(NOT WIN32 AND NOT BUILD_FRAMEWORK)
  configure_file("${CMAKE_CURRENT_SOURCE_DIR}/taglib.pc.cmake" "${CMAKE_CURRENT_BINARY_DIR}/taglib.pc")
  install(FILES "${CMAKE_CURRENT_BINARY_DIR}/taglib.pc" DESTINATION "${LIB_INSTALL_DIR}/pkgconfig")
endif()
```
`pkg-config` works on Windows just fine. Please remove ```NOT WIN32``` condition to allow generating ```taglib.pc``` on Windows.

TsudaKageyu:
Sounds good.
There will be no harm for someone doesn't have ```pkg-config```, except for the fact that useless ```taglib.pc``` is generated.

TsudaKageyu:
We can close this one.

noscript:
Thanks!

PR #614, opened 2015-06-18
Enable pkg-config on Windows.

TsudaKageyu:
This is in response to #613. ```pkg-config``` works on Windows too.


TsudaKageyu:
Thanks @scotchi!

I think #612 and #593 are severe bugs worth it to be fixed immediately.
If you have some time to look at other pull requests, please look at them first.

PR #615, opened 2015-06-18
Calculate the CRC32 checksum in Ogg::Page instead of ByteVector.

TsudaKageyu:
```ByteVector::checksum()``` uses an uncommon variant of CRC32 specializes in Ogg. I guess it was designed to be used with the Ogg format, and (AFAIK) it cannot be used for any other purposes. 

So I suggest to calculate the checksum in ```Ogg::Page``` instead of ```ByteVector```.
I think that it's very unlikely that marking the function as deprecated will give disadvantage to users.


gogglesguy:
This would break apps that use the checksum api from bytevector. Granted, it'd be unlikely anyone is using it. I don't think it's worth the effort for next release. For 2.0, just remove the checksum() from bytevector altogether.

TsudaKageyu:
Right... I admit it doesn't deserve to take a risk. Close it.

PR #617, opened 2015-06-21
Remove an useless function call.

TsudaKageyu:
Just a small refactoring. That's definitely useless since ```MPC::File::findAPE()``` doesn't touch any internal data members.


gogglesguy:
Looks good.

PR #628, opened 2015-06-30
return const correct reverse iterator for newer Sun compiler

RyanLucchese:
The Solaris Studio 12.4 compiler chokes on const conversion for ByteVector::rbegin() and ByteVector::rend() with the following errors:

"taglib/toolkit/tbytevector.cpp", line 749: Error: Cannot return std::reverse_iterator<char*, std::random_access_iterator_tag, char, char&, char*, int> from a function that should return std::reverse_iterator<const char*, std::random_access_iterator_tag, char, const char&, const char*, int>.
"taglib/toolkit/tbytevector.cpp", line 760: Error: Cannot return std::reverse_iterator<char*, std::random_access_iterator_tag, char, char&, char*, int> from a function that should return std::reverse_iterator<const char*, std::random_access_iterator_tag, char, const char&, const char*, int>.

The following patch corrects the constness of the reverse iterators to help Solaris Studio out

### DIFF
--- a/taglib/toolkit/tbytevector.cpp
+++ b/taglib/toolkit/tbytevector.cpp
@@ -746,6 +746,7 @@ ByteVector::ReverseIterator ByteVector::rbegin()
 
 ByteVector::ConstReverseIterator ByteVector::rbegin() const
 {
+  // we need a const reference to the data vector so we can ensure the const version of rbegin() is called
   const std::vector<char> &v = d->data->data;
   return v.rbegin() + (v.size() - (d->offset + d->length));
 }
@@ -758,6 +759,7 @@ ByteVector::ReverseIterator ByteVector::rend()
 
 ByteVector::ConstReverseIterator ByteVector::rend() const
 {
+  // we need a const reference to the data vector so we can ensure the const version of rbegin() is called
   const std::vector<char> &v = d->data->data;
   return v.rbegin() + (v.size() - d->offset);
 }
--- a/taglib/toolkit/tbytevector.cpp
+++ b/taglib/toolkit/tbytevector.cpp
@@ -746,7 +746,11 @@ ByteVector::ReverseIterator ByteVector::rbegin()
 
 ByteVector::ConstReverseIterator ByteVector::rbegin() const
 {
+#if __SUNPRO_CC >= 0x5130
   return ConstReverseIterator(static_cast<const char*>(&*(d->data->data.rbegin() + (d->data->data.size() - (d->offset + d->length)))));
+#else
+  return d->data->data.rbegin() + (d->data->data.size() - (d->offset + d->length));
+#endif
 }
 
 ByteVector::ReverseIterator ByteVector::rend()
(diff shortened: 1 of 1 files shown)


TsudaKageyu:
I'm not familiar with the Sun compiler but, don't these work with it?
```
ByteVector::ConstReverseIterator ByteVector::rbegin() const
{
  return ConstReverseIterator(d->data->data.rbegin()) + (d->data->data.size() - (d->offset + d->length));
}

ByteVector::ConstReverseIterator ByteVector::rend() const
{
  return ConstReverseIterator(d->data->data.rbegin()) + (d->data->data.size() - d->offset);
}
```
If they work, we can remove #ifdefs since they are compatible with other compilers.

RyanLucchese:
That was the first thing I attempted, unfortunately the Sun compiler still errors trying to initialize a type where template arguments differ by constness, thus the (silly looking) dereference/address of/cast juggle:

"../external/taglib-1.9.1/taglib/toolkit/tbytevector.cpp", line 655: Error: Could not find a match for std::reverse_iterator<const char*, std::random_access_iterator_tag, char, const char&, const char*, int>::reverse_iterator(std::reverse_iterator<char*, std::random_access_iterator_tag, char, char&, char*, int>) needed in TagLib::ByteVector::rbegin() const.
"../external/taglib-1.9.1/taglib/toolkit/tbytevector.cpp", line 671: Error: Could not find a match for std::reverse_iterator<const char*, std::random_access_iterator_tag, char, const char&, const char*, int>::reverse_iterator(std::reverse_iterator<char*, std::random_access_iterator_tag, char, char&, char*, int>) needed in TagLib::ByteVector::rend() const.

TsudaKageyu:
Hmm... If your compiler supports ```std::vector::crbegin()```, it might help you.
http://www.cplusplus.com/reference/vector/vector/crbegin/

```
#if defined(__SUNPRO_CC) && (__SUNPRO_CC >= 0x5130)  
  return d->data->data.crbegin() + (d->data->data.size() - (d->offset + d->length));
#else
```

RyanLucchese:
I think crbegin() is actually the write answer, but it's C++11 only. I'm not familiar with the rules for C++11 use in taglib, but we can't use it in our implementation because the standard library that ships with Solaris doesn't support C++11. The following is C++11 portable:
```C++
return d->data->data.crbegin() + (d->data->data.size() - (d->offset + d->length));
```
If C++11 is required, I suppose that's the best answer. If C++11 is acceptable but not required, maybe something like:
```C++
#if __cplusplus >= 201103L
return d->data->data.crbegin() + (d->data->data.size() - (d->offset + d->length));
#elif defined(__SUNPRO_CC) && (__SUNPRO_CC >= 0x5130)
// crazy const hack
#else
return d->data->data.rbegin() + (d->data->data.size() - (d->offset + d->length));
```

TsudaKageyu:
I see. After all, it looks like your solution is the simplest one.

TsudaKageyu:
Excuse me for suggesting similar things repeatedly. How about this one?
I think all we have to do is to ensure that the const version of ```std::vector<char>:::rbegin()``` is called.
```
ByteVector::ConstReverseIterator ByteVector::rbegin() const
{
  const std::vector<char> &v = d->data->data;
  return v.rbegin() + (v.size() - (d->offset + d->length));
}
```

RyanLucchese:
Confirmed and updated. That is much better thank you.

TsudaKageyu:
It would be better to add some comments about this workaround.
It's apparently useless at a glance, so someone would remove them in the future.

RyanLucchese:
Yes good point, thanks.

TsudaKageyu:
Even though it won't work without #627, this one itself is worth it to merge.

PR #629, opened 2015-06-30
check SUNPRO_CC version before using Map workaround

RyanLucchese:
The workaround for the old Solaris Studio compiler causes compilation errors on the newer version

### DIFF
--- a/taglib/ape/apetag.cpp
+++ b/taglib/ape/apetag.cpp
@@ -23,7 +23,7 @@
  *   http://www.mozilla.org/MPL/                                           *
  ***************************************************************************/
 
-#ifdef __SUNPRO_CC
+#if __SUNPRO_CC < 0x5130
 // Sun Studio finds multiple specializations of Map because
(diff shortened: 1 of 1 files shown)


TsudaKageyu:
Looks good. But I wonder if such old versions of the Solaris compiler are still in use, or someone needs to build TagLib with them.

RyanLucchese:
Okay I have pushed your recommended change, thank you. The release date of Solaris Studio 12.4 is November 2014, so I would imagine there are still several instances of 12.3 in the wild. We just recently upgraded from 12.2 at my company, as we have to tread lightly with C++ standard library versions. Until our upgrade yesterday, taglib was building for us on 12.2.

TsudaKageyu:
Sorry I had a bad misconception in the value of __SUNPRO_CC.

TsudaKageyu:
Even though it won't work without #627, this one itself is worth it to merge.

PR #630, opened 2015-06-30
fixed endianness bug in ASF metadata

RyanLucchese:


TsudaKageyu:
This isn't right. I tested it on my PowerPC emulator, and it failed.

Although some of the ```String``` constructors have a comment like this:
```
     * /note If \a t is UTF16LE, the byte order of \a s will be swapped regardless
     * of the CPU byte order.  If UTF16BE, it will not be swapped.  This behavior
     * will be changed in TagLib2.0.
```
```ByteVector``` version of the constructor handles ```UTF16LE``` literally.

lalinsky:
The change doesn't make sense to me, because strings in ASF files are always in UTF16LE. That's a part of the file format.

PR #639, opened 2015-07-31
Run-time check for byte order rather than CMake check.

TsudaKageyu:
It's much safer to check the byte order at run time rather than using a self-made CMake script.
Additionally, the modern compilers are smart enough to do the check at compile-time.


TsudaKageyu:
The main aim of this PR is to remove the self-made CMake script I added after v1.9.1 was released.
I now think it was not a good idea to use an unofficial script in a public release.

TsudaKageyu:
A flaw related to this PR was reported at #641.
I think we should merge this before the next release.

TsudaKageyu:
I self-reviewed this PR again, and now think it's fine to merge it. Any thoughts?

FestusHagen:
Yup, I do ... Merge it!

PR #640, opened 2015-07-31
CMake check for vsprintf_s/vsnprintf rather than sprintf_s/snprintf.

TsudaKageyu:
I become aware of the fact that some systems lacks ```vsnprintf``` despite that they have ```snprintf```.
So I'd like to change the CMake tests to check ```vsnprintf``` directly rather than ```snprintf```.

### DIFF
--- a/ConfigureChecks.cmake
+++ b/ConfigureChecks.cmake
@@ -209,22 +209,22 @@ check_cxx_source_compiles("
   int main() {
     char buf[20];
     va_list args;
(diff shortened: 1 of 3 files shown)


TsudaKageyu:
Hmm, it's weird... anyway, it's still better to check ```vsnprintf``` rather than ```snprintf```.

FestusHagen:
Works ... Thank you!

PR #641, opened 2015-08-01
Use volatile for TestFloatFormat

davispuh:
Just use `volatile` and it won't be optimized away. Current code will compile incorrectly on GCC 5.x and 6.0 with O2/O3 because of different pointer type typecast, see [GCC bug 67090](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67090) 

Tested PR with Clang 3.6.2, GCC 5.2 and 6.0, it compiles correctly with max optimizations.

### DIFF
--- a/cmake/TestFloatFormat.c
+++ b/cmake/TestFloatFormat.c
@@ -2,16 +2,14 @@ int main(int argc, char **argv)
 {
   int ret = 0;
 
-  double bin1[] = {
+  volatile double bin1[] = {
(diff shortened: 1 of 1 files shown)


TsudaKageyu:
You are right. 
I created PR #639 that removes the check itself. This can be another reason to do it.

TsudaKageyu:
This has been fixed via #639. Thanks!

PR #643, opened 2015-08-04
Avoid writing duplicate tags when saving ASF files.

### LINKED ISSUE #621: ASF::File::save() writes duplicate tags any number of times.
```ASF::File::save()``` may write duplicate tags any number of times.
For example:
```
  void testSaveTags()
  {
    ScopedFileCopy copy("silence-1", ".wma");

    {
      ASF::File f(copy.fileName().c_str());
      f.tag()->setTitle(std::string(128 * 1024, 'X').c_str());
      f.save();
      f.save();
      f.save(); // The file has three duplicate tags.
    }
  }
```
I already found a way to fix it. But it may interfere with #556, so I will fix it after it will be merged.

### PR DESCRIPTION
**Pull request by TsudaKageyu**
This fix #621.
I reported several similar bugs, but this one is relatively easy to fix. I think we can merge it to v1.10.

The points of this fix are:
* Clear all ```attributeData```s before adding things to them.
* Update ```d->headerSize``` after writing a tag.
 
This also reduces memory reallocations and copies when saving ASF files.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,6 +5,7 @@
  * Better handling of duplicate ID3v2 tags in all kinds of files.
  * Better handling of duplicate tags in WAV files.
  * Fixed crash when calling File::properties() after strip().
+ * Fixed possible file corruptions when saving ASF files.
  * Marked ByteVector::null and ByteVector::isNull() deprecated.
  * Marked String::null and ByteVector::isNull() deprecated.
  * Many smaller bug fixes and performance improvements.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by lalinsky** [comment]
This looks like a problem that we have also with other formats and it's avoidable by changing user code. It might be easy, but there will always be easy bugs to fix. We should define what fixes we want to have in 1.10 and stick to that. No unnecessary changes.

**Comment by TsudaKageyu** [comment]
I agree with you. So I think we'd like to add some warnings to the comments of ```save()``` functions.

**Comment by TsudaKageyu** [comment]
I'm thinking of adding warnings like this to ```save()``` functions:
```
         * \warning In the current implementation, it's dangerous to call save()
         * repeatedly.  At worst it will corrupt the file.
```
Is it OK?

lalinsky:
Yes, that's fine.

PR #645, opened 2015-08-06
Silence uint ambiguity error in wavepackproperties.

FestusHagen:
FreeBSD fails to build with uint ambiguity error.


TsudaKageyu:
Thanks!

FestusHagen:
Thank you!

TsudaKageyu:
By the way, I think you have an expertise in CMake. How about #644? Is it safe to merge it?

Issue #651, opened 2015-08-29
XiphComment::removeField() refers to String::null which is not exported.


TsudaKageyu:
A user reported a bug to me directly.

```XiphComment::removeField()``` refers to ```String::null``` as its default parameter, and  ```String::null``` lacks ```TAGLIB_EXPORT```. So it leads to a linkage error when using the function with a Windows DLL. 
We can change the default parameter to ```String()``` or ```""```, or add ```TAGLIB_EXPORT``` to ```String::null```, but it may break the ABI.

Any thoughs?

**Comment by CyberSinh** [comment]
Changing the def

**Comment by CyberSinh** [comment]
Changing the default parameter to an empty value ( String() or "" ) would change the behavior of the method. Passing an empty string means "remove all empty tag named xxx" and passing null argument means "remove all tags with name xxx". These two legitimate intentions should be possible.

Exporting String::null seems to be the best way to fix that issue. But I don't know it it could break the ABI.

**Comment by TsudaKageyu** [comment]
It looks like the situation is more complicated than I expected.
I found the following thread at Stackoverflow: http://stackoverflow.com/questions/2479784/exporting-static-data-in-a-dll
According to the discussion, static data member can't be exported... and I confirmed that we can't add ```TAGLIB_EXPORT``` to ```String::null```. It means that ```String::null``` can't be used in the public header files.

So I suggest that we add some alternative functions like below and mark ```XiphComment::removeField()``` deprecated.
```
void removeFields(const String &key);
void removeFields(const String &key, const String &value);
void removeAllFields();
```

CyberSinh:
This solution seems clean and elegant.

scotchi:
It's probably be cleaner to make `String()` and `String::null` equivalent by making `String::String()` copy the null string.  (I'm not sure why that didn't happen, historically.)  Then we can just phase out `String::null` in taglib2.  The behavior is copying something from an old version of Qt, which has now also phased out `QString::null`.  You can change the default arguments without breaking the ABI.

CyberSinh:
According to the tstring.h, String::String() constructs an empty string. If we change the constructor to copy a null string, that would be a breaking change.

TsudaKageyu:
Both ```String::isEmpty()``` and ```isNull()``` are widely used, and they are used in subtle situations. So changing the behavior of ```String::String()``` may cause unpredictable issues.

scotchi:
[...]

PR #652, opened 2015-08-30
Add XiphComment::removeFields() functions to avoid a linkage error ca…

TsudaKageyu:
…used by removeField().
Fixes the bug reported at #651. I'd like to ask someone if it's worth it to merge immediately.


TsudaKageyu:
We should consider this issue in a larger scope. Closed it.

PR #654, opened 2015-09-03
Fix XiphComment::setComment() for the case that a Vorbis comment has …

### LINKED ISSUE #653: Ogg::XiphComment::setComment() doesn't remove field when empty
Hi,

I've just compiled Taglib 1.10 beta and it seems that I cannot remove standard field Comment on Opus Files with XiphComment block.

I've tried all others and they're working (title, artist, album, track, year and genre).

Regards,

TsudaKageyu:
…the "COMMENT" field.

This fixes #653. 
Cause: ```d->commentField``` is not initialized until ```comment()``` is called, and ```setComment()``` updates the "DESCRIPTION" field even if the tag has the "COMMENT" field.

### DIFF
--- a/taglib/ogg/xiphcomment.cpp
+++ b/taglib/ogg/xiphcomment.cpp
@@ -137,7 +137,14 @@ void Ogg::XiphComment::setAlbum(const String &s)
 
 void Ogg::XiphComment::setComment(const String &s)
 {
-  addField(d->commentField.isEmpty() ? "DESCRIPTION" : d->commentField, s);
+  if(d->commentField.isEmpty()) {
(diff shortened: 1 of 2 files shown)


TsudaKageyu:
I changed the behavior of this patch.

Now ```XiphComment``` always uses "COMMENT" instead of "DESCRIPTION" fields when updating them, since "COMMENT" is recommended.
It's the same behavior as "DATE" or "TRACKNUMBER" field.

TsudaKageyu:
Hmm, I found that there are some inconsistent documents online about the usage of "COMMENT" and "DESCRIPTION". I can't decide which field we should prefer.
So I reverted some changes so as not to change the current behavior. It may require further discussion on another issue.

PR #397, opened 2014-07-16
Adds a unified picture interface

MaxLeb:

PR #655, opened 2015-09-07
Fix #94: add a unified interface for pictures

### LINKED ISSUE #94: Request for discussion: unified picture interface
Hi,
I am thinking about implementing something similar to properties() / setProperties() for embedded cover art or other related pictures. Before starting to code I'd like to collect some opinions about the optimal interface for that.

The main problem seems to be that pictures are handled differently by different formats. 
- Apparently in both ID3v2 and FLAC each picture has a TYPE from a fixed enum AND an arbitrary string description.
- For APE there are items of type "BINARY". Can they have both a key and a Stringlist of values as well? The class is not documented. Is there any convention of how to put cover art in APE? Binary items could be everything not only images ...
- Similar problems in MP4: MP4::Item has "toCoverArtList()" but this probably does not work for all items!?

As for the interface, I imagine something like pictures() / setPictures() in analogy to the properties. What should the class returned from pictures() look like?
- a "PictureList" of "Picture" objects, where a "Picture" is something with a type, (optional?) description, and actual picture (plus perhaps mime type, ...)
- a "PictureMap", mapping picture type to (a list of?) picture(s) of that type. This might be overkill.

Then the question is of course how to deal with e.g. APE which does not know picture types. How should they be inferred?

Also some more implementation in the backends is needed which I'm afraid I cannot provide. For example, Ogg::Vorbis::File does not support pictures but according to  http://wiki.xiph.org/VorbisComment#Cover_art pictures are actually supported in OGG.

MaxLeb:
This is an update for PR: 397( https://github.com/taglib/taglib/pull/397 )


taglib/toolkit/tpicture.h
```diff
@@ -0,0 +1,145 @@
...
+
+    /*!
+     * Performs a shallow, implicitly shared, copy of \a p, overwriting the
+     * Picture's current data.
+     */
+    Picture& operator=(const Picture& p);
+
+private:
+    PicturePrivate* p() const{ return _p; }
+    PicturePrivate* _p;
```
scotchi: Two things:

* Everywhere else in TagLib's code, it's a `d`-pointer, following the Qt convention.
* There's no reason to have an accessor for this; the only things that can see the private type also have access to the pointer directly.

taglib/toolkit/tpicture.h
```diff
@@ -0,0 +1,145 @@
...
+            Type type = Other,
+            const String& mime = "image/",
+            const String& description = String());
+
+    /*!
+     * Destroys this Picture instance.
+     */
+    virtual ~Picture();
+
+    const String& mime() const;
```
scotchi: TagLib, by convention, prefers to return these by (copy-on-right in the background) values.  That allows for more flexibility in changing the internal mechanics later.

taglib/toolkit/tpicture.h
```diff
@@ -0,0 +1,145 @@
...
+      //! Logo of the publisher (record company)
+      PublisherLogo      = 0x14
+    };
+
+    /*!
+     * Constructs an empty Picture.
+    */
+    Picture();
+
+    /**
```
scotchi: This is mixing Javadoc and Doxygen comment styles.  TagLib uses Doxygen style (though both are supported by Doxygen).

taglib/toolkit/tpicture.h
```diff
@@ -0,0 +1,145 @@
+/***************************************************************************
+    copyright            : (C) 2002 - 2008 by Scott Wheeler
```
scotchi: This should reflect your contact info, since these files obviously weren't written by me 13 years ago.  ;-)

taglib/toolkit/tpicture.h
```diff
@@ -0,0 +1,145 @@
...
+    const String& mime() const;
+    const String& description() const;
+    Type type() const;
+    const ByteVector& data() const;
+
+    /*!
+     * Performs a shallow, implicitly shared, copy of \a p, overwriting the
+     * Picture's current data.
+     */
+    Picture& operator=(const Picture& p);
```
scotchi: TagLib's coding style puts the `&` along with the variable, not along with the type.

scotchi:
It's probably worth noting that there's a `.astylerc` file in the TagLib source directory.  It's not perfect, but hitting this code with it would reveal a number of issues.

MaxLeb:
Sorry I did not see that file, it is my first time contributing to a project like this.

PR #661, opened 2015-09-08
Support for Apple podcast frames

ufleisch:
I would like to bring up again the case of the proprietary Apple podcast frames because I have received a [feature request](http://sourceforge.net/p/kid3/feature-requests/60/).

Three years ago, Hasso Tepper created a fork for the WFED (a strange hybrid being a text information frame with the name of a URL link frame) and some less suspicious text information frames. He announced it on the [mailing list](https://mail.kde.org/pipermail/taglib-devel/2012-September/002335.html), but maybe missed to create a pull request for his [apple-frames branch](https://github.com/hasso/taglib/commit/bdf06fb335759576521b816cceb6ddefa07c08ed).

I have taken his branch and put an implementation of the PCST frame (another strange creation being a frame which has just four zero bytes and its existence marks the audio file as a podcast) on top of it to make it complete. I have compared the created frames with those created by `eyed3` and the Pa-software ID3 tagger.

### DIFF
--- a/taglib/CMakeLists.txt
+++ b/taglib/CMakeLists.txt
@@ -83,6 +83,7 @@ set(tag_HDRS
   mpeg/id3v2/frames/urllinkframe.h
   mpeg/id3v2/frames/chapterframe.h
   mpeg/id3v2/frames/tableofcontentsframe.h
+  mpeg/id3v2/frames/podcastframe.h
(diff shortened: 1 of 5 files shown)


TsudaKageyu:
I think this one is worth it to merge.

PR #662, opened 2015-09-15
Add support for M4V tag parsing

robUx4:
In VLC we use taglib a lot to get title/author/etc of audio and video files. M4V is among the common file formats that taglib can parse easily.


TsudaKageyu:
This patch looks good. However, please wait for us to check if TagLib can handle M4V files correctly.

TsudaKageyu:
It works fine. Thanks!

PR #663, opened 2015-09-15
Allow create FileRef from an already opened IOStream

robUx4:
In Windows Store apps, it's not possible to access the files directly using the regular file operations. The only way to open files (asynchronously) is through managed code. Using the raw IOStream class it's possible to go over this limitation in taglib.


TsudaKageyu:
Hmm, it's almost the same as #528, isn't it?

robUx4:
Yes, I just noticed that. This is actually the same code, updated to the latest master.

robUx4:
#528 was updated with a better version of this code. Discard this patch.

PR #664, opened 2015-09-23
add options R, I, D for replace/insert/delete of arbitrary tags

pbhd:


sbooth:
This looks great and will be useful! However, the TagLib coding style does not use spaces before parentheses for function calls.  The file https://github.com/taglib/taglib/blob/master/.astylerc covers the main coding style elements.

Also, do you think it makes sense to check for the presence of an additional argument for `R` and `D` options before dereferencing `argv`?

pbhd:
Changed it according to your suggestions...

TsudaKageyu:
:+1:

PR #669, opened 2015-10-10
Fixes access violation

### LINKED ISSUE #670: Setting empty properties on MP4 tag segfaults
rogram received signal SIGSEGV, Segmentation fault.
0x00007fffedf95fc0 in std::basic_string<wchar_t, std::char_traits<wchar_t>, std::allocator<wchar_t> >::size() const () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
(gdb) bt
#0  0x00007fffedf95fc0 in std::basic_string<wchar_t, std::char_traits<wchar_t>, std::allocator<wchar_t> >::size() const () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x00007ffff254580c in TagLib::String::toInt (this=0x33e4f68, ok=0x0) at /home/erwinj/src/taglib/taglib/toolkit/tstring.cpp:483
#2  0x00007ffff25457dd in TagLib::String::toInt (this=0x33e4f68) at /home/erwinj/src/taglib/taglib/toolkit/tstring.cpp:476
#3  0x00007ffff257a783 in TagLib::MP4::Tag::setProperties (this=0x33d8d10, props=...) at /home/erwinj/src/taglib/taglib/mp4/mp4tag.cpp:936
#4  0x0000000001e80787 in TestTagParser_EmptyProperties_Test<TagLib::MP4::Tag>::TestBody (this=0x33d8d00) at TestTagLoaderTagLib.cpp:111

This is due to the following code in mp4tag.cpp:

else if(it->first == "COMPILATION") {
         bool value = (it->second.front().toInt() != 0);
         d->items[name] = MP4::Item(value);

Bad things happen when it->second is empty, and .front() returns garbage.

pokowaka:
Fixes #670

- Fixes access violation when setting empty stringlist on integer
  properties in mp4 tag
- Add a unit test that validates the fix.

### DIFF
--- a/taglib/mp4/mp4tag.cpp
+++ b/taglib/mp4/mp4tag.cpp
@@ -917,7 +917,7 @@ PropertyMap MP4::Tag::setProperties(const PropertyMap &props)
   for(; it != props.end(); ++it) {
     if(reverseKeyMap.contains(it->first)) {
       String name = reverseKeyMap[it->first];
-      if(it->first == "TRACKNUMBER" || it->first == "DISCNUMBER") {
+      if((it->first == "TRACKNUMBER" || it->first == "DISCNUMBER") && it->second.size() > 0) {
         int first = 0, second = 0;
         StringList parts = StringList::split(it->second.front(), "/");
         if(parts.size() > 0) {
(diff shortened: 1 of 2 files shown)


TsudaKageyu:
Good catch and fix @pokowaka. :+1:

PR #675, opened 2015-11-05
Add self-assignment check in operator=

garima-g:
Method 'operator=' should check its argument with 'this' pointer.


TsudaKageyu:
Looks good. Thanks!

PR #685, opened 2015-11-26
Added ccache support for faster compilation

g-coder:
ccache is a compiler cache. It speeds up recompilation by caching previous compilations and detecting when the same compilation is being done again. It provides support for C/C++ languages.
It makes compilation fast if it is available on the machine.
https://ccache.samba.org/

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,8 +139,8 @@ add_custom_target(docs doxygen)
 
 find_program(CCACHE_FOUND ccache)
 if(CCACHE_FOUND)
-    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
-    set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
+  set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
+  set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
 endif(CCACHE_FOUND)
 
 # uninstall target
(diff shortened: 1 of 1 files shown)


TsudaKageyu:
I like this feature. :+1: 
However, we use two spaces for an indent instead of four. We appreciate if we could stick to our style.

g-coder:
Thanks for your review. I made the required change. :)

FestusHagen:
Personally, I don't see the need for the trouble this will cause ... It takes all of a minute and a half to rebuild taglib from scratch!

And I'm on old hardware ...

Just my opinion for what it's worth!

TsudaKageyu:
Hi @FestusHagen, first, Thank you for pointing out the issue of the style. I will fix it soon.

Privately, I like this feature. I sometimes rebuild and test TagLib several tens of times a day without exaggeration. This patch and ccache really helps me.
Of course, we could remove this if it will cause some trouble. I think we have enough time to try it.

FestusHagen:
I do get it, I do understand the build time,  Many of us build things back to back to back over and over and over again and again and again ... And again and again and again!

I also get the torment caches can cause, it is of my opinion that developers most of all should not be building and testing with(against) cached objects!

Like I said, Nothing personal, Just my opinion ... No big deal.

-Enjoy

sbooth:
I'm with @FestusHagen on this one- I'm generally opposed to cached objects in a build tree.  I'd prefer to see this disabled by default.  But, if the developer has ccache installed they must think it's a benefit.

TsudaKageyu:
I can understand some people don't like such feature.
So how about adding a build option like ```ENABLE_CCACHE``` which is disabled by default?

sbooth:
@TsudaKageyu that sounds like a great idea to me!  It keeps the current build behavior but allows developers to opt-in to ccache usage.

FestusHagen:
Super Awesome @TsudaKageyu.

No unwitting use!

-Enjoy

PR #690, opened 2015-11-30
fix pkgconfig file for cross compiling

stefansaraev:
as title says. libdir / includedir shall never be hardcoded in generated pkgconfig file.


TsudaKageyu:
Sorry, I'm not familiar with pkgconfig, so I'm not very eligible to review this PR.
I'd like to wait for someone to review it.

lalinsky:
I'm not too familiar with pkg-config either, but this uses the same approach as this example, so I guess it's how pkg-config files should normally look like.

http://people.freedesktop.org/~dbn/pkg-config-guide.html#concepts

PR #691, opened 2015-11-30
Silence ushort ambiguity errors

FestusHagen:


TsudaKageyu:
Thank you for your catch and fix @FestusHagen.

TsudaKageyu:
Honestly, I'd like to remove such typedefs and use standard integer types in taglib2 branch.

sbooth:
I agree with that! Anything we can do to avoid unnecessary duplication is
worthwhile.

FestusHagen:
I'd like to see it consistent ...

-Enjoy

TsudaKageyu:
It looks like the types like ```ushort``` are defined in some standard header.
Assuming that, the ambiguity errors may occur in any header. I think we'd better add ```TagLib::``` to all such types in the public headers.

TsudaKageyu:
Well, I will create another issue to discuss it.

TsudaKageyu:
I have eliminated all the custom integer types from the source. I hope this kind of issues will never happen again.

PR #698, opened 2015-12-08
More robust checks for invalid MPEG frame headers.

### LINKED ISSUE #697: Bad allocation on mp3 file load
We bumped Kodi to taglib 1.10 and it seems, like we found a regression.

When we try to access this file: http://saraev.ca/01%20-%20Eternity.mp3
via 
```
file = mpegFile = new MPEG::File(stream, ID3v2::FrameFactory::instance());
```
it fails with a ```bad allocation```
If you need any more info, please let me know.

image

TsudaKageyu:
This fixes #697.

Cause:
The first MPEG frame of the file has invalid MPEG version bits. They make TagLib to miscalculate the length of MPEG frames and allocate huge memory when seeking VBRI/Xing headers.

Solution:
Check for the validity of MPEG frames more strictly, and skip invalid MPEG frames when gathering the audio properties.


lalinsky:
Is this something that should be released in a hot-fix 1.10.1 release? If yes, we will probably need to create a separate branch, because there have been a lot of changes already in master for a patch release. Or we can release master as 1.11 if you prefer that.

scotchi:
Also, as a side note, is there any specific reason to rip out the bitset code and replace it with binary operators?  At least to me, code using bitset is much easier to read.

TsudaKageyu:
Privately, it looked inconsistent for me that both ```bitset``` and logical operations were used mixedly.

TsudaKageyu:
Anyway, I'm willing to work on it in either way.

PR #703, opened 2015-12-21
Fix .pc file configuration.

inizan-yannick:
.pc file use ${} as variables. so use @var@ as substitutions and set @ONLY parameter in CMakeLists.txt


FestusHagen:
:+1:

-Enjoy

PR #706, opened 2015-12-29
Remove "FORCE" in CMake path specifications to allow installation to a non-standard directory structure

TwoFX:
This change originated from porting this library to [Haiku OS](https://www.haiku-os.org/). There has also been a post on the mailing list regarding this patch. All it does is allow the user to overwrite the installation directories for the different components of taglib, which is useful, when working with a different general folder structure than the one found on Linux. This change does not break any existing build process, it only enables the user to overwrite the directories if they so desire.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,10 +42,10 @@ set(TESTS_DIR "${CMAKE_CURRENT_SOURCE_DIR}/tests/")
 
 ## the following are directories where stuff will be installed to
 set(LIB_SUFFIX "" CACHE STRING "Define suffix of directory name (32/64)")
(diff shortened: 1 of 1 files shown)


TsudaKageyu:
I think it will work fine. However, I'm not an expert in CMake. So I'd like to hear someone else's opinion.

FestusHagen:
This should be fine, the users it could affect are those that build TagLib within their projects and set those variables before TagLib.

`set(... CACHE ...)` ...
If variable exists, `set()` will not reset variable without FORCE
if variable does not exist, `set()` will set variable as usual.

-Enjoy

TwoFX:
Thank you very much everyone!!

PR #716, opened 2016-02-10
Album artist api

pdgendt:


taglib/tag.h
```diff
@@ -97,6 +97,12 @@ namespace TagLib {
     virtual String album() const = 0;
 
     /*!
+     * Returns the album artist name; if no album artist name is present in the
+     * tag String::null will be returned.
+     */
+    virtual String albumArtist() const = 0;
```
TsudaKageyu: As @scotchi metioned, we can't add virtual functions due to ABI compatibility.

To merge this pull request to our `master` branch, you need to change the function like this: https://github.com/taglib/taglib/blob/master/taglib/audioproperties.cpp#L108-L154
I know it's really ugly, but it's necessary to emulate a virtual function without breaking the ABI.

taglib/ape/apetag.cpp
```diff
@@ -136,6 +136,13 @@ String APE::Tag::album() const
   return d->itemListMap["ALBUM"].values().toString();
 }
 
+String APE::Tag::albumArtist() const
+{
+  if(d->itemListMap["ALBUMARTIST"].isEmpty())
+    return String::null;
```
TsudaKageyu: `String::null` has been marked deprecated for some reasons. Please use an empty string like other functions of this class.
Also, please revise the related comments.

taglib/ogg/xiphcomment.cpp
```diff
@@ -154,6 +165,11 @@ void Ogg::XiphComment::setAlbum(const String &s)
   addField("ALBUM", s);
 }
 
+void Ogg::XiphComment::setAlbumArtist(const String &s)
+{
+  addField("ALBUMARTIST", s);
```
TsudaKageyu: How about when `ALBUM ARTIST` or `ALBUM_ARTIST` field exists?

taglib/mod/modtag.cpp
```diff
@@ -68,6 +68,11 @@ String Mod::Tag::album() const
   return String();
 }
 
+String Mod::Tag::albumArtist() const
+{
+  return String::null;
```
0xcaff: Do mod tags not have album artists?

taglib/ogg/xiphcomment.cpp
```diff
@@ -99,6 +99,17 @@ String Ogg::XiphComment::album() const
   return d->fieldListMap["ALBUM"].toString();
 }
 
+String Ogg::XiphComment::albumArtist() const
+{
+  if(!d->fieldListMap["ALBUMARTIST"].isEmpty())
```
0xcaff: `ALBUMARTIST` isn't in the [Xiph spec](https://www.xiph.org/vorbis/doc/v-comment.html#fieldnames), but `PERFORMER` is. Why `ALBUMARTIST`?

sbooth:
I don't think in practice `PERFORMER` is used by most players or taggers. 
 `ALBUMARTIST` is used by [Picard](https://picard.musicbrainz.org/docs/mappings/) and is listed on [HydrogenAudio](http://wiki.hydrogenaud.io/index.php?title=Tag_Mapping).  Performer seems to typically refer to a musician playing a specific instrument.

PR #727, opened 2016-03-06
TAGLIB_EXPORT on the class causes warnings for std::string

MordredKLB:
Setting TAGLIB_EXPORT on the class instead of the individual functions generates compiler warnings for both std::string when trying to build taglib as a .dll. See [discussion here](http://stackoverflow.com/questions/4145605/stdvector-needs-to-have-dll-interface-to-be-used-by-clients-of-class-xt-war). Moving the TAGLIB_EXPORT to just the exported functions fixes this, even if it isn't as pretty.

### DIFF
--- a/taglib/toolkit/tiostream.h
+++ b/taglib/toolkit/tiostream.h
@@ -33,21 +33,21 @@
 namespace TagLib {
 
 #ifdef _WIN32
-  class TAGLIB_EXPORT FileName
+  class FileName
   {
   public:
-    FileName(const wchar_t *name);
-    FileName(const char *name);
+    TAGLIB_EXPORT FileName(const wchar_t *name);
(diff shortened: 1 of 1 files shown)


FestusHagen:
This should be two pull requests ...

However, 98ceaab is remarkably similar to #716, with the same issues!

MordredKLB:
Screwed up and accidentally committed #716 inside this PR. Should be good now.

TsudaKageyu:
We have been aware of the issue for years, but untouched due to the binary compatibility.
Are you sure this one won't break the ABI?

These guidelines would help you:
https://techbase.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B

MordredKLB:
Yeah, that'd help. I'll close it out.

PR #739, opened 2016-04-22
c: fix a typo

frgm:
readble => readable


frgm:
++ [taglib_file_new()](https://github.com/taglib/taglib/blob/d52e97dfcd1c4c7385c1ca9c334e9571da330162/bindings/c/tag_c.h#L110-L111) returns:
> NULL if the file type cannot be determined or the file cannot  be opened.

It returns NULL only if the file without extension.

PR #748, opened 2016-07-23
Adds a function for dynamic version information retrieval

supermihi:
The current way of exposing TagLib's version only through #define's
makes it impossible for clients (e.g. language bindings) to reliably
determine the TagLib version that is currently in use: using the
define's in client code will statically copy the compile-time values
into the client's library, but if TagLib is dynamically bound the
version (at least minor and patch version) can change after building
client code.

### DIFF
--- a/taglib/CMakeLists.txt
+++ b/taglib/CMakeLists.txt
@@ -321,6 +321,7 @@ set(matroska_SRCS
 )
 
 set(toolkit_SRCS
+  toolkit/taglib.cpp
   toolkit/tstring.cpp
   toolkit/tstringlist.cpp
   toolkit/tstringhandler.cpp
--- a/taglib/toolkit/taglib.cpp
+++ b/taglib/toolkit/taglib.cpp
@@ -4,31 +4,34 @@
 
 using namespace TagLib;
 
-String TagLib::GetVersionString()
+
+String TagLib::Version::string()
 {
   return String::number(TAGLIB_MAJOR_VERSION)
       + "." + String::number(TAGLIB_MINOR_VERSION)
       + "." + String::number(TAGLIB_PATCH_VERSION);
 }
 
-unsigned int TagLib::GetMajorVersion()
-{
-  return TAGLIB_MAJOR_VERSION;
-}
-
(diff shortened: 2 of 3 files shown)


TsudaKageyu:
Basically, I agree with you. However, I think that TagLib should not use a C++11 class like ```std::tuple```.
Please note that we have kept it compatible with non-C++11 compilers. I like to return a primitive type like ```const char *``` or ```int```. ```TagLib::String``` would be a good option as well.

I'd like to hear someone's opinion. What way is better?

supermihi:
Thanks for your response. Actually, I wasn't aware that `std::tuple` is C++11. One could of course debate whether support of C++11/14 features can be considered standard by the time taglib2 is released, but that's a different issue. :)

My objection agains version strings is that they it is rather cumbersome to compare them automatically because e.g. `"1.11.2" > "1.9.0"` returns `false`. One could alternatively encode the three numbers in an integer, something like
```
unsigned version = (TAGLIB_MAJOR_VERSION << 4) | (TAGLIB_MINOR_VERSION << 2) | TAGLIB_PATCH_VERSION
```
but that's not entirely user-friendly either ...

sbooth:
I think this is a good idea also.  We could add several functions:

```
// Returns the version as a string in the form (Major Version).(Minor Version).(Patch Version), e.g. "4.2.0"
TagLib::String TagLib::GetVersionString();

// Returns the version as an unsigned integer in the form (Major Version << 16) | (Minor Version << 8) | (Patch Version), e.x. 0x040200
uint TagLib::GetVersion();

// Returns the major version, e.g. 4
uint TagLib::GetMajorVersion();

// Returns the minor version, e.g. 2
uint TagLib::GetMinorVersion();

// Returns the patch version, e.g. 0
uint TagLib::GetPatchVersion();
```

supermihi:
Souns good; my updated PR implements this.

supermihi:
Sounds reasonable; I've updated the PR correspondingly. Note I had to introduce parentheses around the `minor` and `major` function to avoid a name clash with `#define`'s from `sysmacros.h` ... :-/

Issue #758, opened 2016-10-18
Support for classical music tags introduced in iTunes 12.5


ufleisch:
With iTunes 12.5, Apple introduced a few new atoms, or frames, respectively, to better support classical music. It would be great if TagLib supports such tags.
- M4A:
  - ©wrk: Work (string)
  - ©mvn: Movement Name (string)
  - ©mvi: Movement Number (number)
  - ©mvc: Movement Count (number)
  -  shwm: Show Work & Movement (0/1)
- ID3 (2.3, 2.4; MVN, MVI for 2.2):
  - MVNM: Movement Name
  - MVIN: Movement Number/Count

PR #759, opened 2016-10-18
Support new classical music frames introduced with iTunes 12.5, #758.

ufleisch:
M4A:
    ©wrk: Work (string)
    ©mvn: Movement Name (string)
    ©mvi: Movement Number (number)
    ©mvc: Movement Count (number)
    shwm: Show Work & Movement (0/1)

ID3 (2.3, 2.4; MVN, MVI for 2.2):
    MVNM: Movement Name
    MVIN: Movement Number/Count

### DIFF
--- a/taglib/mp4/mp4tag.cpp
+++ b/taglib/mp4/mp4tag.cpp
@@ -71,10 +71,10 @@ MP4::Tag::Tag(TagLib::File *file, MP4::Atoms *atoms) :
       parseIntPair(atom);
     }
     else if(atom->name == "cpil" || atom->name == "pgap" || atom->name == "pcst" ||
-            atom->name == "hdvd") {
+            atom->name == "hdvd" || atom->name == "shwm") {
       parseBool(atom);
     }
-    else if(atom->name == "tmpo") {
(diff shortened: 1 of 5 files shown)


scotchi:
I'm a little bit conflicted about supporting writing these tags since they're non-standard extensions (when the ID3v2 standard has a mechanism for adding custom frames via either `TXXX`, `PRIV` or `GEOB`).

ufleisch:
Yes, with MVNM and MVIN it's the same story as with WFED, they are simple text identification frames, and Apple could use TXXX (or COMM or PRIV) or at least let them start with a 'T'. And also the other frames they invented (PCST, TCAT, TDES, TGID, TCMP, probably TSO2, TSOC) could have been put into TXXX frames without pain (as opposed to other non-standard frames like CHAP, CTOC which do have too much structured information to fit into a simple text frame). Maybe they have never read the standard.

But there are so many people who use iTunes that it is not easy to just ignore such frames (or the wishes of people who would like to have them supported).

TsudaKageyu:
IMO, this PR is worth merging and I'd like to do it soon.

PR #761, opened 2016-10-23
WinRT compatibility

albertofustinoni:
Some simple changes to allow TagLib to compile as a WinRT assembly.


taglib/toolkit/tfilestream.cpp
```diff
@@ -51,10 +51,18 @@ namespace
...
+	  return CreateFile2(path.wstr().c_str(), access, FILE_SHARE_READ, OPEN_EXISTING, NULL);
+#else
+	  return CreateFileW(path.wstr().c_str(), access, FILE_SHARE_READ, NULL, OPEN_EXISTING, 0, NULL);
+#endif
+	else if(!path.str().empty())
+#if (_WIN32_WINNT >= 0x0602)
+	  return CreateFile2(path.toString().toCWString(), access, FILE_SHARE_READ, OPEN_EXISTING, NULL);
+#else
+	  return CreateFileA(path.str().c_str(), access, FILE_SHARE_READ, NULL, OPEN_EXISTING, 0, NULL);
+#endif
```
TsudaKageyu: IMO, this way is better than having two `#if` blocks. 
```
#if (_WIN32_WINNT >= 0x0602)
    return CreateFile2(path.toString().toCWString(), ...);
#else
    // The original code goes here.
#endif
```

taglib/toolkit/tfilestream.cpp
```diff
@@ -437,7 +445,13 @@ long FileStream::length()
 #ifdef _WIN32
 
   SetLastError(NO_ERROR);
+#if (_WIN32_WINNT >= 0x0602)
```
TsudaKageyu: It's safer to consider the case that `_WIN32_WINNT` is undefined.

taglib/toolkit/tfilestream.cpp
albertofustinoni: I understand the point about protecting against _WIN32_WINNT being undefined, but don't you still end up having two code paths? In both cases you have to consider wchar and char strings being used.

TsudaKageyu: Your code and mine are exactly equivalent, since `path.toString()` has the same branch as `if(!path.wstr().empty()) ...` in it. My intention is to keep our codebase maintainable by having simpler `#if ~ #endif` structures.

TsudaKageyu:
Thanks! It looks good.

albertofustinoni:
Likewise, thank you for merging :)

TsudaKageyu:
@albertofustinoni,
Out of interest, isn't `SetFilePointer()` superseded by `SetFilePointerEx()`?

albertofustinoni:
Possibly. To be honest, I did the bare minimum of edits needed to get the source to compile as a WinRT assembly. SetFilePointer didn't cause an error in Visual Studio so I didn't even see it.

In any case, turns out that CreateFile2 only works for files in the app's own install folder or data folder: anything else has to be accessed using the new async WinRT APIs, so while my edits allow the taglib's source to compile it won't be very useful on its own.

I have created an alternative implementation of TIOStream that does just that, but I don't think Taglib would want it because:
a. It is quite invasive: tfilestream needs to be rewritten, file paths cannot be used anymore (filestream equivalents are created from WinRT's IRandomAccessStream)
b. It is written in C++/CX (I don't have the stomach to use COM code to invoke the WRL directly, sorry)
c. As such, it requires Taglib to be compiled as a WinRT component
d. I'm afraid the unit tests can't be linked against Taglib compiled this way: callable API surface needs to be exported as WinMD metadata and not as header files+stub libs
e. Even if they could, Travis CI doesn't really support Windows, much less UWP

On account of all that, I think I'll have to fork: the plan would be to keep in sync with Taglib's source code, replace the build system with a native Visual Studio tool chain and add my file access implementation.
Should taglib be open to figure a way to add WinRT support upstream I'd be happy to help, of course.

TsudaKageyu:
Thanks! I understood your intention.

PR #776, opened 2016-11-23
Small improvements

mathbunnyru:
Made some small changes to slightly improve cppcheck analyser output.


TsudaKageyu:
I think that those changes are reasonable.

PR #779, opened 2016-11-27
Fixes #743 by not overwriting existing Xiph comment in FLAC::File::save

### LINKED ISSUE #743: Inconsistent properties behavior on FLAC file with ID3 tag
Hi all,
according to https://github.com/supermihi/pytaglib/issues/19, a user experiences the following problem: A FLAC file has an ID3 tag. When he deletes the COMMENT entry with setProperties(), the entry is still there upon the next call to properties() after a save() and subsequent read().

After having a look at the implementation,
1. properties() forwards TagUnion::properties(), which returns the first existing tag of (Xiph, ID3v2, ID3v1)
2. setProperties() always writes a Xiph comment (creates if necessary).

Hence, what _should_ happen is that, after deleting the COMMENT tag, a Xiph comment that contains the remaining tags is written. Any subsequent properties() calls should return that Xiph comment, which does not have a COMMENT entry anymore.

Any ideas what could be the problem here? Not being able to remove certain tags with the generic interface is somewhat unsatisfactory.

Michael

supermihi:
IMHO this fix is okay but I'd like someone with better understanding of the `FLAC::File` internals to review it before merging.

Thanks,
Michael

### DIFF
--- a/taglib/flac/flacfile.cpp
+++ b/taglib/flac/flacfile.cpp
@@ -168,8 +168,9 @@ bool FLAC::File::save()
   }
 
   // Create new vorbis comments
-
-  Tag::duplicate(&d->tag, xiphComment(true), false);
+  if (!hasXiphComment())
+    Tag::duplicate(&d->tag, xiphComment(true), false);
+  
 
   d->xiphCommentData = xiphComment()->render(false);
 


TsudaKageyu:
IMO, it looks good.
ID3 tags are non-standard for FLAC files, so I think we have no good reason to copy ID3 tags into Vorbis comments.

sbooth:
One concern I have with this PR is that it introduces the potential for tag information to be lost if a FLAC file has an ID3v2 tag.  Granted, ID3v2 in FLAC is very nonstandard but this has been the behavior since 7fe6647435856631c8351a34d04d7e01acc79d1e

supermihi:
I don't see how this PR could lead to ID3v2 tags to be lost. It is effective only when there already exists a Xiph comment, in which case the ID3v2 tag will remain untouched. Moreover, the simple interface (like `Tag::title()`)  will continue to return the ID3v2 values if the respective tag does not exist in Xiph, because of the `TagUnion` implementation.

The only behavioral change of this PR should occur when an existing Xiph comment is overwritten with a new one, and ID3 contains tags that do not exist in the to-be-written Xiph comment. I cannot imagine a client delibaretely depending on this extremely counterintuitive behavior ...

sbooth:
I looked again and I'm not sure what I was thinking.  I reviewed the code and it looks good to me.

Issue #793, opened 2017-01-25
Non free files unicode.{h,cpp}


cypriani:
Hi there,

The automated testing tools in Debian warned me that the files `unicode.h` and `unicode.cpp` from the `taglib/toolkit/` directory are not released under a free software license. See the corresponding lintian tag description for more information: [license-problem-convert-utf-code](https://lintian.debian.org/tags/license-problem-convert-utf-code.html).

With the upcoming Debian release, I had to act quickly, so I repacked the taglib tarball to remove the incriminated files, and wrote a patch that uses ICU (as lintian suggests). Here is the link to that patch:
https://sources.debian.net/src/taglib/1.11.1+dfsg.1-0.1/debian/patches/icu.patch/

The downside is obviously a new external dependency, but libicu is a well recognised library and should be safe to use. I made sure the unit tests still pass with the patch.

As you can see, I used a naive character by character copy to convert between `wchar_t` and `UChar` buffers, which is certainly not ideal. For some reason, the ICU functions that perform this conversion did not give the same results, and made the unit tests fail (see comments in the patch). I did not investigate for too long, because I think the best way to solve this would be to use ICU types and functions in the whole taglib code base, which would suppress the need for these conversions altogether. In my opition, it would be far better than relying on the size of `wchar_t` (which, as you know, can be either bigger or smaller than 16 bits) as is currently the case. ICU works on Windows, and using it would also allow you to get rid of the windows-specific Unicode code.

As is, my patch requires cmake >= 3.7, because the FindICU.cmake module was not shipped with prior cmake releases. It would be trivial to vendor that file if needed, or you could just document that cmake >= 3.7 is needed to build.

Please let me know your thoughts. If you're willing to integrate my patch, I'd be happy to turn it into a pull request if more convenient for you.

  Matteo

scotchi:
The license seems very permissive:

     * Unicode, Inc. hereby grants the right to freely use the information
     * supplied in this file in the creation of products supporting the
     * Unicode Standard, and to make copies of this file in any form
     * for internal or external distribution as long as this notice
     * remains attached.

I can't see how that could be labeled "non-free".  It doesn't use a standard license, but it does meet the Free Software definition.

cypriani:
Hi Scott,

[...]

PR #794, opened 2017-01-27
Replace unicode.h/unicode.cpp by the UTF8-CPP library.

TsudaKageyu:
`unicode.h/unicode.cpp` are no longer maintained and incompatible with Debian's guideline as reported at #793.
[UTF8-CPP](https://github.com/nemtrif/utfcpp) is another library which offers virtually the same feature as `unicode.h/unicode.cpp` with the STL-friendly style. It is maintained on GitHub and published under the Boost Software License which is compatible with either LGPL or MPL and should go along with Debian's guideline.


FestusHagen:
Can't believe I'm going to say this to a third a party addition ... (just don't tell anybody)

Simply Awesome!

-Enjoy
fh  :  )_~
RIP Hammer

TsudaKageyu:
Additionally, I realized that `unicode.h/unicode.cpp` accepts invalid UTF-8 sequences which contain surrogate characters.

PR #797, opened 2017-02-02
Enable ByteVectorStream to work with FileRef

### LINKED ISSUE #796: Use FileRef constructor passing IOStream
I'm using TagLib in an Android project, building it in arm mode and it's all working.

I'm trying to manipulate audios without passing file paths, but passing to the native code a the `byte[]` which is then converted to `char *`.
I double checked the integrity of the bytes while passing them to the native code, but when I initialize the `FileRef`, it always fails again the `isNull()` check.

```
char* buf = new char[ numBytes ];
env->GetByteArrayRegion(data, 0, numBytes, reinterpret_cast<jbyte*>(buf));

ByteVector byteVector = ByteVector::fromCString(buf, numBytes);
TagLib::IOStream *vectorStream = new TagLib::ByteVectorStream(byteVector);
TagLib::FileRef f(vectorStream);

if(f.isNull())
     return;
```

TsudaKageyu:
This will fix #796.
`FileRef` requires a file extension to determine the file type, but `ByteVectorStream` can't hold a file name.
This PR enables `ByteVectorStream` to take and hold a file name via its constructor.

### DIFF
--- a/taglib/toolkit/tbytevectorstream.cpp
+++ b/taglib/toolkit/tbytevectorstream.cpp
@@ -23,38 +23,40 @@
  *   http://www.mozilla.org/MPL/                                           *
  ***************************************************************************/
 
-#include "tbytevectorstream.h"
+#include <cstring>
+#include <cstdlib>
+
 #include "tstring.h"
 #include "tdebug.h"
-
-#include <stdio.h>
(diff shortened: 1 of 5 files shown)


fillobotto:
For me it's working now

PR #799, opened 2017-02-04
Enable FileRef to detect file types by the actual content of a stream.

TsudaKageyu:
`FileRef` doesn't work with `ByteVectorStream` as reported at #796, since `ByteVectorStream` is not associated with a file name and `FileRef` detects file types based on file extensions.
This commit makes `FileRef `to work with `ByteVectorStream` by enabling it to detect file types based on the actual content of a stream.

This is not a plugable and extensible solution, but I think it's impossible to implement it not changing the current interface and behavior. At least we have to add a stream-based replacement of `FileRef::FileTypeResolver` since we can't add a virtual function to it.

The overview is:
* Every `File` type (not literally) has a static method `static bool isSupported(IOStream *stream);`  to do a quick check if a stream can be opened as the type itself.
* `FileRef` calls `isSupported()` of each type to detect file types.
* The file extension has priority over the contents. It's necessary in order not to change the current behavior.


fillobotto:
Once done, si it going to be merged to branch 2?

TsudaKageyu:
No, this is based on the master branch.
We would have to devise cleaner, plugable and extensible solution for the taglib2 branch.

Issue #795, opened 2017-02-01
File::


TsudaKageyu:
If we have an ASCII text file like "0123456789", file.find("56", 0, "34") return 5 even thoug it should return -1.

However, I'm not sure if it's worth fixing. The parameter before is not used at all

TsudaKageyu:
Additionally, I realized that `rfind()` can't find a pattern lies across two buffers.
I think that it should be fixed.

PR #800, opened 2017-02-04
Fix File::find() and rfind() to handle the parameter 'before' properly.

TsudaKageyu:
As reported at #795, the parameter `before` of `File::find()` and `rfind()` work in some cases and `rfind()` can't find a pattern lies across two buffers.
Though those bugs are virtually ignorable, I think it's better to fix them.

This also enables `File::find()` and `rfind()` to handle patterns which are greater than 1024 bytes.


TsudaKageyu:
Sorry, this one is still buggy.

PR #813, opened 2017-03-27
Added setProperties implementation to TagUnion

albertofustinoni:
Added setProperties implementation to TagUnion.

It works by calling setProperties  on each tag making up the union and returns only the properties unsupported by all member tags as unsupported.

This allows many commonly supported properties that are not part of the base Tag class accessors (AlbumArtist for example) to be read and written.

### DIFF
--- a/taglib/tagunion.cpp
+++ b/taglib/tagunion.cpp
@@ -125,7 +125,8 @@ namespace TagLib
     std::vector<PropertyMap> returnCandidates;
     for (size_t i = 0; i < COUNT; ++i) {
       if (d->tags[i])
-        returnCandidates.insert(returnCandidates.end(), d->tags[i]->setProperties(properties));
+        returnCandidates.insert(returnCandidates.end(),
+          d->tags[i]->setProperties(properties));
     }
 
     if (!returnCandidates.empty()) {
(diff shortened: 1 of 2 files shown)


taglib/tagunion.cpp
```diff
@@ -119,6 +119,40 @@ namespace TagLib
...
+        returnCandidates.insert(returnCandidates.end(), d->tags[i]->setProperties(properties));
+    }
+
+    if (!returnCandidates.empty()) {
+      //Only one tag present, return its unassigned properties
+      if (returnCandidates.size() == 1) {
+        return returnCandidates.front();
+      }
+
+      //Multiple tags in union: if a property has been assigned in any member tag, remove it from ignored properties to return
```
scotchi: I haven't looked at this in detail at all, but just from a quick glance the super long lines need to be truncated.  TagLib generally tries to keep things to 80-ish columns.

claudiuslollarius: Broke up long lines as suggested

albertofustinoni:
@scotchi I have broken up the long lines to be about 8 chars long

supermihi:
Looks good to me. Any objections to merging this PR?

Issue #814, opened 2017-03-27
Missing taglib_default_extensions to C bindings


redrabbit:
Hello.

I've been using the C bindings and discovered that `TagLib::FileRef::defaultFileExtensions()` has not been implemented yet.

scotchi:
The C API is only a small wrapper around the most important functions in the C++ API.  Most functions from the C++ API aren't represented there; this is just one of hundreds where that's the case.

PR #815, opened 2017-03-27
Add taglib_default_file_extensions() to C bindings

redrabbit:
This PR add `taglib_default_file_extensions()` to the C bindings (see #814). I think this function may  be useful to check if a file is supported by TagLib without having to open and read the file.


sbooth:
Would it make more sense to store the array in a static variable instead of using `malloc` on each invocation?

ufleisch:
The repository has been deleted, it there still interest in such a feature?

Issue #822, opened 2017-06-05
Suggestion: Move to another unit test framework


TsudaKageyu:
**CppUnit** haven't been able to catch up the latest compilers. I found that it doesn't work well with the latest AppleClang without an option and I made a [quick workaround](https://github.com/taglib/taglib/commit/6000a19f70ef57f993b7f5594a2dcf9651e5801d). 
Additionally, I have thought all along that CppUnit is not very handy. For example, it requires useless type conversions like:
```
    // Why I can't compare TagLib::String to a string literal?
    CPPUNIT_ASSERT_EQUAL(f.tag()->title(), String("Title1"));

    // I'd like to simply do:
    CPPUNIT_ASSERT_EQUAL(f.tag()->title(), "Title1");
```

So I'd like to suggest that we move from CppUnit to another up-to-date unit test framework.
I privately tried [Boost.Test](http://www.boost.org/doc/libs/1_64_0/libs/test/doc/html/index.html) and it worked fine. (The diff is [here](https://github.com/TsudaKageyu/taglib/commit/8e544d13443210115f8e2727b757b66e7c02c229).) Other libraries like [Google Test](https://github.com/google/googletest) or [Catch](https://github.com/philsquared/Catch) would be great.

sbooth:
I agree, I'm not a fan of CppUnit.  I always like boost but I don't know enough about Google Test or Catch to propose a selection.

TsudaKageyu:
I have tried another library [Catch](https://github.com/philsquared/Catch). (See 41968bb192a671c040c7be5bff40afcd5a0c8dd2) I find it really handy and love it.
Especially, it doesn't require any preparation (like building static libs, adding paths and blah blah blah) even used with Visual Studio since it's a single-header library. Of course, it works fine on either Linux or OSX too.

I'd like to create a pull request based on the commit 41968bb192a671c040c7be5bff40afcd5a0c8dd2.

PR #825, opened 2017-06-12
Convert unit tests from CppUnit to Catch.

TsudaKageyu:
As I mentioned in #822, **CppUnit** haven't been able to catch up the latest compilers. It requires a workaround to work with the latest AppleClang. (See 6000a19f70ef57f993b7f5594a2dcf9651e5801d)
So I'd like to suggest that we change our unit test framework from CppUnit to [**Catch**](https://github.com/philsquared/Catch), one of the up-to-date and well-maintained unit test frameworks.

**Catch** is a single-header library maintained on GitHub. I find it really handy especially on Windows/MSVC since it doesn't require any preparation like building libs, adding paths and blah blah blah. Of course, it works fine on Linux or OSX without any external dependencies.


sbooth:
I'd like to go ahead with this.  Are there any major objections?

atimholt:
Hey, no biggy, but I was just peeking my head into the taglib repo here and noticed that this pull request seems not to have gone through, but that it looks like it was meant to?

—And maybe this is overthinking things, or bikeshedding, or whatever, but doctest is a similar unit test library to Catch, but much faster (according to [their github page](https://github.com/onqtam/doctest), at least). I think it may even be very nearly directly compatible with Catch—I seem to recall it actually originating as a fork of Catch. Maybe I'd try offering a pull request myself, but again, maybe such a thing is too “bikeshed-y”, especially with this pull request apparently not having gone through.

neheb:
gtest might be better. catch2 recently decided to break compatibility while retaining the catch2 name.

neheb:
@TsudaKageyu can you rebase this?

neheb:
Alright so CMake doesn't do subprojects which makes using Catch2 difficult. That is, OS provided Catch2. Catch2 itself really shouldn't be included here.

OS wise, Catch2 is all over the place. Older Debian/Ubuntu have only Catch1. Newer have Catch2 v2 and no v3. Others only have v3.

gtest is much better OS wise. But it is more verbose than Catch2.

sbooth:
Is this PR still relevant?

ufleisch:
If you ask me, neither.
CppUnit may not be the most modern unit test framework for C++, but its feature set is enough for TagLib (e.g. we do not need to mock anything), and the tests run quickly, which is the most important point for me.

PR #827, opened 2017-06-27
Taglib2: Enhanced support for Wav format

yanxijian:
1. Tag info cannot be displayed on Windows 7.
2. Support non-FORMAT_PCM format


TsudaKageyu:
Hi @yanxijian,
Could you explain the points of this PR in a little more detail? 
Additionally, those changes can be applied to the ```master``` branch since they don't change the interface. I think this PR is more suitable for the ```master``` branch rather than ```taglib2```.

yanxijian:
Hi @TsudaKageyu 
1. If ID3V2 tag ahead of Info tag, the following information cannot be displayed in Explorer on Windows 7. 


2. Some wav files are not FORMAT_PCM format and the "totalSamples" is 0. if a return directly, we cannot get the length(duration time) of these wav files. In fact, if we ignore this error(or exception), we can still  calculate the length with streamLength and byteRate. Of course, my solution here might be lack of consideration.

ufleisch:
In TagLib 1.12, support for WAV files was improved. Can you check if the problems addressed by this PR have been solved with the current master branch?

yanxijian:
I'm sorry, but I'm no longer working in a related field.

PR #830, opened 2017-07-21
Crash when loading an empty mpeg file

yanxijian:
This is not really  a question. When loading an empty (empty.txt -> empty.mp3 ), program will crash.
buffer.size() is 0, then buffer.size() - 1 is undefined (unsigned int).


sbooth:
This is almost an argument for using `int64` instead of `unsigned int` for sizes and lengths.  But this is a good fix.

sbooth:
To keep with TagLib's coding style could you place the opening brace for the `if` on the same line?

yanxijian:
@sbooth Updated : )

PR #832, opened 2017-08-03
Add support for file descriptor to FileStream

safu9:
I added support for file descriptor to FileStream. (only for non-Windows)

### DIFF
--- a/taglib/toolkit/tfilestream.cpp
+++ b/taglib/toolkit/tfilestream.cpp
@@ -58,6 +58,11 @@ namespace
 #endif
   }
 
+  FileHandle openFile(const int fileDescriptor, bool readOnly)
+  {
+    return InvalidFileHandle;
+  }
+
   void closeFile(FileHandle file)
   {
     CloseHandle(file);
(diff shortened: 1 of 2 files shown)


fillobotto:
Why is this PR still on hold? Support for file descriptor would be great

sbooth:
My only concern with this PR is whether a file is more likely to be modified without TagLib's knowledge if a caller has a reference to the FD being used.  That may not be a good argument against merging, though.

safu9:
I didn't expected such usage, but certainly there is a possibility of that. I was thinking of Android when I added this PR because c/c++ library like TagLib can't access files in a external sd card without using file descriptor. I think this PR is useful to Android application developers.

taglib/toolkit/tfilestream.cpp
```diff
@@ -158,6 +168,29 @@ FileStream::FileStream(FileName fileName, bool openReadOnly)
...
+    d->readOnly = false;
+  else
+    d->file = openFile(fileDescriptor, true);
+
+  if(d->file == InvalidFileHandle)
+  {
+# ifdef _WIN32
+    debug("Could not open file using file descriptor");
+# else
+    debug("Could not open file file using file descriptor");
```
sbooth: `file` is repeated twice. Why does this section have an `#ifdef`?

safu9: Thank you. I've fixed it.

PR #843, opened 2017-11-19
Added virtual destructor for FileTypeResolver which is an abstract class

voldyman:
`FileRef::FileTypeResolver` is an abstract class with a pure virtual function but the destructor isn't marked virtual.

I have added a virtual destructor but haven't made it pure virtual so the deriving classes have the option of implementing a destructor and are not forced to define a destructor.


voldyman:
sorry my bad, will raise another reivew

PR #844, opened 2017-11-19
Added virtual destructor for FileTypeResolver which is an abstract class

voldyman:
`FileRef::FileTypeResolver` is an abstract class with a pure virtual function but the destructor isn't marked virtual.

I have added a virtual destructor but haven't made it pure virtual so the deriving classes have the option of implementing a destructor and are not forced to define a destructor.


scotchi:
This can't be done in the master branch for binary compatibility reasons (you can't add or remove virtual functions and maintain binary compatibility) and has already been done in the taglib2 branch.

Issue #508, opened 2015-02-28
Add support for .aac files (AAC in ADTS)


ghost:
For .AAC files, is there a way to create an audio header so I can edit tags and whatnot?

TsudaKageyu:
.AAC files are already supported by ```MP4::File``` class, but ```FileRef``` doesn't recognize the extension. Should we add ```".aac"``` to ```FileRef```?

lalinsky:
`MP4::File` only supports AAC-in-MP4, not AAC-in-ADTS (which is basically just raw AAC with extra synchronization headers).

TsudaKageyu:
I see, I'd like to look for further information. Thanks.

shaforostoff:
i have extended audio attribute parsing, now samplerate is read correctly, and it even provides an estimate for bitrate and length

kluszon:
As I see aac files still aren't support and recognized by `FileRef`. Can we do sth with it?

shaforostoff:
> As I see **.aac** files still aren't support and recognized by `FileRef`. Can we do sth with it?

taglib author seems to be reluctant to add new stuff to mainline, i have a bunch of fixes since 2018 and none of them has been included. so this projects needs to be forked, I guess.

as for aac support, you are welcome to use my patch: https://github.com/taglib/taglib/pull/856/

ufleisch:
Implemented with #856.

PR #856, opened 2018-03-02
Raw AAC (ADTS) reading support

shaforostoff:
samplerate and channels should be detected correctly, and it even provides an estimate for bitrate and length

and tag reading worked out of the box: many players are explicitly using MPEG::File for parsing .aac files
for example: https://mailman.videolan.org/pipermail/vlc-commits/2016-February/034366.html

addresses #508


sbooth:
I believe this PR is essentially good but should the class be called `ADTSHeader` (capitalized acronym)?

shaforostoff:
i have addressed your comment

taglib/mpeg/adtsheader.cpp
```diff
@@ -0,0 +1,207 @@
+/***************************************************************************
+    copyright           : (C) 2018 inMusic brands, inc. All rights reserved.
```
jonaski: This code has a conflicting license. It can't be "All rights reserved" at the same time as GPL.

taglib/mpeg/mpegproperties.cpp
```diff
@@ -155,6 +156,25 @@ bool MPEG::Properties::isOriginal() const
...
+  if(file->containerFormat() == File::ADTS) {
+    const MPEG::ADTSHeader firstHeader(file, 0);
+
+    if(firstHeader.bitrate() > 0) {
+      d->bitrate = firstHeader.bitrate();
+      d->length = file->length() / d->bitrate / 8;
+    }
+
+    d->sampleRate        = firstHeader.sampleRate();
+    d->channels          = firstHeader.channelMode() == ADTSHeader::FrontCenterLeftRightSideLeftRightBackLeftRightLFE ? 8 : static_cast<int>(firstHeader.channelMode());
```
scotchi: Also note that TagLib usually tries to keep lines to about 80 characters wide.  This one is 170 wide.

taglib/mpeg/adtsheader.cpp
sbooth: Good point. We will need @shaforostoff to make this change since he's the author.

shaforostoff: the web-site says LGPL and MPL. Anyways, I will remove that

taglib/mpeg/mpegproperties.cpp
shaforostoff: turned that into if/else

shaforostoff:
Is there anything I can do for this pull request to be merged?

scotchi:
It's possible that one of the other TagLib developers will have some time to look at this before I have a chance to review it in detail, but with the current global situation (and specifically my kids' pre-school closing, reopening, then closing again), it's difficult to say when I'll have a block of time where I can go through outstanding TagLib PRs and merge where appropriate.

ufleisch:
I had a look at your patch. Concerning the style, you should replace the tab characters in *adtsheader.h* by spaces. But there are other things which worry me more:

- Currently, most users of TagLib just take an `MPEG::File` to work with AAC files. They will not get properties, but at least, they can edit ID3v1, ID3v2 and APE tags on the AAC files. With your patch, AAC files become read-only, you get the audio properties, but you cannot edit tags - but the later is the purpose of TagLib.
- For AAC files, you have to explicitly set the `ContainerFormat` in one of the new `MPEG::File` constructors. I would prefer it if the format is automatically detected, so we do not have to set it manually.

ufleisch:
I have improved the code so that ADTS is automatically detected when opening an `MPEG:File` with an AAC file. Therefore AAC files can be tagged in the same way as MP3 files. In standard `Average` read mode, bitrate and length are estimated, it should be accurate enough, but probably it would be good if someone could have a look at it.

PR #857, opened 2018-03-02
avoid freezes reading incomplete mp3 files

### LINKED ISSUE #842: [patch] avoid freezes reading incomplete mp3 files
avoid being stuck in previousFrameOffset() called from lastFrameOffset() when position >= length()

https://pastebin.com/xjc4PK5N

shaforostoff:
closes #842


dr0biwan:
There's an issue with using this patch as-is if the file is CBR & no appropriate header to help determine the file duration can be located.

The following might be more appropriate `if (position > length()) position = length();` as clamping to a ~1MB end position prevents the lastFrameOffset related checks in mpegproperties.cpp from working correctly.

shaforostoff:
> There's an issue with using this patch as-is if the file is CBR & no appropriate header to help determine the file duration can be located.

i have updated the merge request

ufleisch:
Please provide instructions on how to reproduce such a freeze. Then we could find the cause why `position` could become larger than `length()`, maybe because the `APELocation` can be wrong?

Issue #862, opened 2018-03-23
MP4 support for atoms cmID, purl, egid is missing


ufleisch:
TagLib lacks support for some MP4 atoms which are used by iTunes and are supported by mp4v2, mutagen and AtomicParsley:

- The composer ID (cmID) would need explicit parsing as UInt (like "atID").
- Podcast URL (purl) and Podcast GUID (egid) would be text but have the flag 0 (when generated with AtomicParsley or iTunes) which is ignored by TagLib's parseText() because it expects a flag of 1.

PR #863, opened 2018-03-23
Add support for cmID, purl, egid MP4 atoms (#862).

ufleisch:
Add support for the MP4 atoms cmID (Composer ID), purl (Podcast URL) and egid (Podcast GUID).


sbooth:
This may be a silly question, but isn't `cmID` an internal iTunes ID? I'm curious what the use case is outside of iTunes.

ufleisch:
Probably the Composer ID (cmID) is an internal iTunes ID, but so are Artist ID (atID), Catalog ID (cnID) and Genre ID (geID). The latter three are supported by TagLib, so it is not wrong to support cmID too. I was asked by a Kid3 user to support cmID, there may be valid reasons to have these atoms editable, and be it merely to delete them.

PR #873, opened 2018-06-30
Add BPM read-only information to "generic" interface and C bindings

AdamHarries:
This PR adds further functionality to the "generic" tag interface, and (on top of that), the C bindings. 

This may be a contentious addition, as only 3/7 of the tag "types" (i.e. classes that inherit from `Tag`) currently support BPM information. I would, however, that the majority of "modern" media formats for which we care about BPM information (such as ALAC, FLAC, mp3, m4a, etc) are covered by some combination of those 3 tag types. 

I would welcome some discussion on this, in case there's a better way of going about this!

### DIFF
--- a/bindings/c/tag_c.cpp
+++ b/bindings/c/tag_c.cpp
@@ -210,6 +210,12 @@ unsigned int taglib_tag_track(const TagLib_Tag *tag)
   const Tag *t = reinterpret_cast<const Tag *>(tag);
   return t->track();
 }
+ 
+unsigned int taglib_tag_bpm(const TagLib_Tag *tag)
+{
+  const Tag *t = reinterpret_cast<const Tag *>(tag);
+  return t->bpm();
+}
 
 void taglib_tag_set_title(TagLib_Tag *tag, const char *title)
 {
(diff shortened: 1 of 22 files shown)


madmaxoft:
Why is the BPM an int? There's no reason not to have non-integral numbers as BPM.

AdamHarries:
A very good point! My experience with tags is mostly with id3v2, which
stores BPM as an integer, so I defaulted to that as a BPM type. Once I'm
back at a computer I'll have a look through the type that the various tag
formats use (as an aside, is there a good repository of format
documentation out there?) to see if there's a need for it to be a float.

madmaxoft:
If I'm reading the source right, ID3v2 stores the BPM as a string, so anything can go there, really. I've just tried storing a generic string in an MP3 file using Winamp, and it seems to work (not sure if it is ID3v2-compliant, though :)

AdamHarries:
Ah, maybe I'm thinking of how iTunes deals with BPMs then ;)

I'll try extending it to floating point numbers, and see what breaks :P

scotchi:
It's not possible to add things to the generic interface in the current master branch for binary compatibility reasons (you can't add virtual functions to a base class).  In TagLib 2, it's likely that the current generic interface will go away in favor of the property interface, which is more flexible.

AdamHarries:
Thanks scorchi - in the meantime I'll use my own fork of taglib, as BPM information is quite critical for my application.

Do you have a timescale for when taglib2 with be feature compatible with taglib, and whether the property interface will have c bindings?

Issue #882, opened 2018-12-19
RIFF chunks are searched at odd offset after non zero padding


ufleisch:
I came across an AIFF file which is not parsed by TagLib. When reading the file, `TagLib: RIFF::File::read() -- Chunk 'reg' has invalid ID` is printed. After analyzing the situation, I recognized that it is not the invalid ID which causes parsing to fail, but the padding byte which is 1 instead of 0. There is an SSND chunk with an odd size in the file followed by one byte padding with the value 0x01 and then a "regn" chunk follows. TagLib ignores the padding byte because it is not zero. It then searches for the next chunk ID at the odd position of the padding, and that ID is invalid because it starts with the padding byte 0x01, which is not ASCII.

I looked into the specification at [Resource Interchange File Format (RIFF)](https://docs.microsoft.com/en-us/windows/desktop/xaudio2/resource-interchange-file-format--riff-) which states

> Where chunkID is a FOURCC that identifies the data contained in the chunk, chunkSize is a 4-byte value giving the size of the data section of the chunk, and data is zero or more bytes of data. The data is always padded to the nearest WORD boundary. chunkSize gives the size of the valid data in the chunk. It does not include the padding, the size of chunkID, or the size of chunkSize.

Here it is not said that padding has to be zero (however, I have to admit that it is possible to find other documents on the Web which say that the padding should be a zero). I think that the analyzed file is not 100% clean because the padding is not zero. However, the behavior of TagLib is not correct when it is searching for the beginning of a chunk at an odd offset, it is specified that chunks always start at a word boundary. Therefore I would suggest to relax the current scheme and tolerate padding bytes which are not zero.

ufleisch:
I added a pull request #883.

ufleisch:
Fixed with #883.

PR #883, opened 2018-12-19
Do not ignore non zero RIFF padding if leading to parse error (#882)

ufleisch:
Currently padding in RIFF files is ignored if it is not zero. To allow parsing of files with non zero padding bytes, this patch changes the parsing of RIFF files in the following way:

* If a zero padding byte is found, it is accepted and handled as before.
* If the padding byte is not zero, it is accepted if the four bytes following the padding byte contain a valid chunk ID but the four bytes starting at the padding byte would not be a valid chunk ID, i.e. parsing would fail without this patch.

This should make parsing more tolerant in the presence of non zero padding bytes while still providing the feature that chunks at odd offsets can be parsed.

### DIFF
--- a/taglib/riff/rifffile.cpp
+++ b/taglib/riff/rifffile.cpp
@@ -325,9 +325,20 @@ void RIFF::File::read()
     if(offset & 1) {
       seek(offset);
       const ByteVector iByte = readBlock(1);
-      if(iByte.size() == 1 && iByte[0] == '\0') {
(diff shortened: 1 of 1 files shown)


ufleisch:
I have seen that my original patch was not strict enough for some AIFF files, so I have adapted it slightly to respect the padding whenever it does not lead to a wrong chunk name.

PR #887, opened 2019-02-03
Taglib2 Simple Tag and File type identification.

abelmul:
Simple tag type and file type identification for taglib. This allows code like this to work.
```cpp
tl::FileRef file(file_path.c_str());
tl::Tag* tag = file.tag();

if (tag && tag->getType() == tl::TagTypes::ID3v2) {
    m_artist = tag->artist().to8Bit(true);
    m_title = tag->title().to8Bit(true);
    m_album = tag->album().to8Bit(true);
    valid = true;
}
```
Instead of
```cpp
tl::MPEG::File* f = dynamic_cast<tl::MPEG::File*>(file.file());
if (!file.isNull() && f->hasID3v2Tag()) {
    m_artist = file.tag()->artist().to8Bit(true);
    m_title = file.tag()->title().to8Bit(true);
    m_album = file.tag()->album().to8Bit(true);
    valid = true.
}
```

The problem with the second one is that is supposes id3v2 tags are only for mpeg files. But there are many that implement id3v2 tags [see here](https://en.wikipedia.org/wiki/ID3#Non-MP3-implementation_and_alternatives).


sbooth:
Does this PR add functionality that isn't already available using `PropertyMap`?

abelmul:
I think the code may be misleading. The `PropertyMap` doesn't support file and tag identification. For example this is a code from [clementine player](https://www.clementine-player.org).
```cpp
  if (TagLib::MPEG::File* file =
          dynamic_cast<TagLib::MPEG::File*>(fileref->file())) {
// Do something with mpeg file
  } else if (TagLib::FLAC::File* file =
                 dynamic_cast<TagLib::FLAC::File*>(fileref->file())) {
// Do something with flac file
  } else if (TagLib::MP4::File* file =
                 dynamic_cast<TagLib::MP4::File*>(fileref->file())) {
// Do something with mp4 file.
  }
```
With this PR we can do this(without dynamic cast)
```cpp
switch(fileref->file()->getType())
{
     case(FileTypes::MPEG):
         //do something with mpeg file
        break;
      
     case(FileTypes::FLAC):
         //do something with flac file
        break;
     case(FileTypes::MP4):
         //do something with mp4 file
        break;
}
```
[see](https://github.com/clementine-player/Clementine/blob/master/ext/libclementine-tagreader/tagreader.cpp)

sbooth:
Is there a reason to avoid `dynamic_cast`? I am probably missing something but I'm not seeing what functionality is lacking that this PR adds.

supermihi:
As far as I understand it's a convenience feature. Not sure if this belongs into the core library, however. To mee, it seems a rather rare use case to identify the file type other than just looking at the extension.

PR #888, opened 2019-02-08
fileref: Use user defined resolvers on streams

chouquette:
Since the resolve only use the filename, there is no reason not to probe
them, as a filename might be available through the IOStream interface.
When using a ByteVectorStream, the filename will be empty and user
defined resolvers won't be probed.

### DIFF
--- a/taglib/fileref.cpp
+++ b/taglib/fileref.cpp
@@ -27,6 +27,8 @@
  *   http://www.mozilla.org/MPL/                                           *
  ***************************************************************************/
 
+#include <cstring>
+
 #include <tfile.h>
 #include <tfilestream.h>
 #include <tstring.h>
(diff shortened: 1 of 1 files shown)


chouquette:
Hi,

Any chance you could review/apply this?

Thanks a lot!

taglib/fileref.cpp
```diff
@@ -477,7 +481,11 @@ void FileRef::parse(FileName fileName, bool readAudioProperties,
 void FileRef::parse(IOStream *stream, bool readAudioProperties,
                     AudioProperties::ReadStyle audioPropertiesStyle)
 {
-  // User-defined resolvers won't work with a stream.
+    // Try user-defined resolvers.
+
+    d->file = detectByResolvers(stream->name(), readAudioProperties, audioPropertiesStyle);
+    if(d->file)
+      return;
```
ufleisch, line 488: Too much indenting

PR #889, opened 2019-02-08
fileref: Handle .aac files as MPEG::File

chouquette:


chouquette:
This seems to be a (lower quality) duplicate of #856 so I suppose this could be closed

sbooth:
Closed per @chouquette request

PR #892, opened 2019-03-14
Fixed an MP4 AudioProperties parsing error

### LINKED ISSUE #891: TagLib 2: Incorrect parsing of MP4 'Media Header' atom
MP4's encoded with the mdhd version 1 atom incorrectly report the track length as '0'.

When parsing a MP4 Media Header (mdhd) version 1 atom, the 'unit' (timescale) is parsed as a 64-bit integer. This should actually be a 32-bit integer (the timescale is only 4 bytes long, see link below). The 'length' (duration) is then off by 4 bytes as a result.

See:
https://wiki.multimedia.cx/index.php/QuickTime_container

timusus:
The 'unit' (timescale) of 'mdhd' v1 atoms is a 32-bit unsigned integer (4 bytes long), but is parsed as a 64-bit integer (8 bytes long). This also affects the byte offset of the 'length' (duration).  This results in MP4 AudioProperties reporting a track length of zero.

This change addresses both problems.

Resolves #891

See:
https://wiki.multimedia.cx/index.php/QuickTime_container

### DIFF
--- a/taglib/mp4/mp4properties.cpp
+++ b/taglib/mp4/mp4properties.cpp
@@ -195,8 +195,8 @@ MP4::AudioProperties::read(File *file, Atoms *atoms)
       debug("MP4: Atom 'trak.mdia.mdhd' is smaller than expected");
       return;
     }
-    unit   = data.toInt64BE(28);
-    length = data.toInt64BE(36);
+    unit   = data.toUInt32BE(28);
+    length = data.toInt64BE(32);
   }
   else {
     if(data.size() < 24 + 4) {


sbooth:
This looks good to me; it also jives with https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/QTFFChap2/qtff2.html#//apple_ref/doc/uid/TP40000939-CH204-25615

Should the size check on line 194 also be updated to reflect the change?

timusus:
I believe the spec you link to refers to the mdhd version 0 headers. In version 1, the timescale remains a 4 byte integer, but the `creation time`, `modification time` and `duration` are all changed from 4 bytes to 8 bytes. 

|  Entry | Bytes (v0) | Bytes (v1) |
| :---         |     :---:      | :---: |
| size  | 4  | 4 |
| type  | 4  | 4 |
| version | 1 | 1 |
| flags | 3 | 3 |
| creation time* | 4 | **8** |
| modification time* | 4 | **8** |
| time scale | 4 | 4 |
| duration* | 4 | **8** |
| language | 2 | 2 |
| quality | 2 | 2 |

If I understand the code correctly,

`if(data.size() < 36 + 8)..`

The `+ 8` here represents `size + type`, and the `36` is the sum of the remaining bytes in the header? If that is the case, then `36 + 8` is the correct minimum byte size of a v1 mdhd header.

sbooth:
You're right about the documentation from Apple, I looked at the size of 1 byte for *Version* and read that as a value of 1.

Comparing the version 1 size check to the size check at line 202 (presumably for version 0; perhaps it should be made explicit) I wonder if that line should check for  size of `24 + 8` instead of `24 + 4`.

timusus:
I think you're right there. Should we deal with that under a separate PR?

Also, looking at the master branch, I think this mdhd parsing issue probably exists there as well.

Are you able to merge this change Stephen?

sbooth:
I think the version 0 issues should be under a separate PR.

PR #894, opened 2019-03-16
TagLib2: Minor fix for mp4 media header v0 minimum size check

timusus:
Mp4 media header (mdhd) v0 atoms are a minimum of 8 bytes for size & type information, plus 24 bytes for remaining entries (`24 +8`) bytes in total, rather than (`24 + 4`).

Relates to #895 (same issue, different branch)

See https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/QTFFChap2/qtff2.html#//apple_ref/doc/uid/TP40000939-CH204-25615

### DIFF
--- a/taglib/mp4/mp4properties.cpp
+++ b/taglib/mp4/mp4properties.cpp
@@ -199,7 +199,7 @@ MP4::AudioProperties::read(File *file, Atoms *atoms)
     length = data.toInt64BE(32);
   }
   else {
-    if(data.size() < 24 + 4) {
+    if(data.size() < 24 + 8) {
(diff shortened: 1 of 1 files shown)


sbooth:
Along with this fix does it still make sense to assume that if version isn't 1 it should be handled as version 0, or should there be an explicit check for 0 and an error return otherwise?

timusus:
I'm not really sure what the correct approach is here. I don't know a lot about the mp4 spec, and I can't seem to find the authoritative source (I guess Apple keep this information private?). 

I'd say we could explicitly handle v0 and v1, and throw an error if the version is anything else, with a message that this version is unknown to TagLib. I'd again probably say this could be decided under the scope of a separate issue/PR.

Issue #885, opened 2019-01-17
Issue with MP4 rate atom in git


astippich:
Hello,

openSUSE Tumbleweed recently updated to a git version of taglib, and since then the CI tests of KDE's KFileMetaData library fail with the MP4 "rate" atom.
I think the issue is https://github.com/taglib/taglib/pull/818
which changes the internal representation of the rate atom from text to int. However, this causes a regression when reading files that previously saved the rating as text. For example, if the rate atom was written to "80" as text, or 0x3830, taglib will now interpret it as integer, which yields 14384.
I am also now unable to read ratings saved with e.g. MediaMonkey or MP3Tag, which seem to save them as text as well.
Thoughts?

StefanBruens:
I think this could be fixed in a compatible way by respecting the "class" of the atom when reading it. The sample file mentioned above has class 1 == UTF-8 for the "rate" atom.

scotchi:
The fix for this was already merged.

PR #896, opened 2019-03-17
Respect atom type when converting rate tag

StefanBruens:
TagLib prior to #818 (commit ff28cf276c60)read and wrote the "rate" tag as
text, and switched to reading it as integer value even when the atom class
is a text type. This breaks reading existing files, and can be avoided by
taking the atom type into account.

This fixes issue #885.

### DIFF
--- a/taglib/mp4/mp4tag.cpp
+++ b/taglib/mp4/mp4tag.cpp
@@ -74,8 +74,19 @@ MP4::Tag::Tag(TagLib::File *file, MP4::Atoms *atoms) :
             atom->name == "hdvd" || atom->name == "shwm") {
       parseBool(atom);
     }
-    else if(atom->name == "tmpo" || atom->name == "rate" || atom->name == "\251mvi" || atom->name == "\251mvc") {
+    else if(atom->name == "tmpo" || atom->name == "\251mvi" || atom->name == "\251mvc") {
       parseInt(atom);
+    }
+    else if(atom->name == "rate") {
+      AtomDataList data = parseData2(atom);
+      if(!data.isEmpty()) {
+        AtomData val = data[0];
+        if (val.type == TypeUTF8) {
+          addItem(atom->name, StringList(String(val.data, String::UTF8)));
+        } else {
+          addItem(atom->name, (int)(val.data.toShort()));
+        }
+      }
     }
     else if(atom->name == "tvsn" || atom->name == "tves" || atom->name == "cnID" ||
             atom->name == "sfID" || atom->name == "atID" || atom->name == "geID" ||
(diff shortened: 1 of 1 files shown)


dfaure-kdab:
Can this be reviewed soon?
The kfilemetadata unittests are broken because of this regression.
Thanks :)

sbooth:
Since the atom value is always rendered as `int` would it make sense when the atom's existing representation is text to parse the string and store it as an integer value?

astippich:
Any progress here?

sbooth:
> Any progress here?

Per my comment above I'm waiting on a response from @StefanBruens as I am not sure if the PR as it stands now is correct.

StefanBruens:
@sbooth 
Any code developed against older versions of taglib will provide the rating as a String(List), so always writing the MP4::Items integer value will give bad results.
This is even more critical, as the union in MP4::Item::Itemprivate is not initialized, thus a random value will be written.
Probably the right approach would be to check if the MP4::Item has a non-empty Stringlist and write it as TypeUTF8 then, otherwise write it as TypeInteger.

astippich:
Well, you have to change your code anyways if you want write, and I think the only fix for that is to revert the change to int. How about writing the rating value to both in case it is a string?

StefanBruens:
Ping!

sbooth:
What is the expected type for the `rate` atom, text or int? I think TagLib should always write the expected type but be lenient with reading. This is why I suggested parsing text values as ints, to preserve the value but write the expected atom type (assuming it's int as currently implemented). I realize there is a backward compatibility issue with that approach. If there truly is no de facto standard for what the atom contains should it be always handled as text?

StefanBruens:
There is no canonical type.
As all released versions of taglib uses a text MP4 atom for the value, and any software using taglib for the last 12 years had to use text for reading and writing the value, I strongly prefer text.
The patch set restores compatibility with released taglib versions, and avoids writing garbage to the file.

taglib/mp4/mp4tag.cpp
```diff
@@ -74,9 +74,20 @@ MP4::Tag::Tag(TagLib::File *file, MP4::Atoms *atoms) :
             atom->name == "hdvd" || atom->name == "shwm") {
       parseBool(atom);
     }
-    else if(atom->name == "tmpo" || atom->name == "rate" || atom->name == "\251mvi" || atom->name == "\251mvc") {
+    else if(atom->name == "tmpo" || atom->name == "\251mvi" || atom->name == "\251mvc") {
       parseInt(atom);
     }
+    else if(atom->name == "rate") {
+      AtomDataList data = parseData2(atom);
```
sbooth: I believe this should be `parseData` instead of `parseData2`

StefanBruens: Obviously not - parseData returns a ByteVectorList, which loses the Atom type which is required here.

PR #898, opened 2019-03-24
Fix crash in vorbis when opening invalid files

astippich:
When invalid vorbis files are put into taglib, no XiphComment is generated. This leads dereferences on invalid pointers when calling any function on the tags later (e.g. title(), properties() etc.) and crashes.
Fix this by creating a default XiphComment when opening of the file fails.

### DIFF
--- a/taglib/ogg/flac/oggflacfile.cpp
+++ b/taglib/ogg/flac/oggflacfile.cpp
@@ -180,6 +180,7 @@ void Ogg::FLAC::File::read(bool readProperties, Properties::ReadStyle properties
 
   if (!d->scanned) {
     setValid(false);
+    d->comment = new Ogg::XiphComment();
     return;
   }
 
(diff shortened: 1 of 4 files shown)


sbooth:
I think it is better for client code to use the existing mechanism to check if a file was successfully opened using `isValid()` instead of changing the library's behavior.

astippich:
I agree that one should always use file.isValid(). However, I would like to point out that only Vorbis files crash and none of the others as the TagUnion used there handles this gracefully. So this patch actually aligns it more with the other formats.

sbooth:
You bring up a good point about the behavior of `TagUnion`. For files that support only one type of tag (for example DSF) accessing the tag of an invalid file will cause a segfault. I hadn't thought about the inconsistency before.

astippich:
So, is this an appropriate fix?

PR #899, opened 2019-04-13
mp4 properties: handle the case when mp4 file header has zero bitrate

shaforostoff:
calculate it based on filesize and length

i can send you the link to the broken file, if you like

### DIFF
--- a/taglib/mp4/mp4properties.cpp
+++ b/taglib/mp4/mp4properties.cpp
@@ -213,7 +213,12 @@ MP4::Properties::read(File *file, Atoms *atoms)
           pos += 3;
         }
         pos += 10;
-        d->bitrate = static_cast<int>((data.toUInt(pos) + 500) / 1000.0 + 0.5);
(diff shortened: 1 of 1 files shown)


sbooth:
Could you send me the link to the broken file?

shaforostoff:
hey, i did send you the link. did you get it?

scotchi:
Do you happen to know what created these files with broken headers?  Unless it's something very common, my impulse is not to work around this sort of brokenness.

ufleisch:
The calculated bitrate seems to be slightly too high, so I suggest that we incorporate this patch into #981 and use the same calculation:

```
        if(data.toUInt(pos) != 0u || d->length <= 0) {
          d->bitrate = static_cast<int>((data.toUInt(pos) + 500) / 1000.0 + 0.5);
        }
        else {
          d->bitrate = static_cast<int>((calculateMdatLength(atoms->atoms) * 8) / d->length);
        }
```

ufleisch:
Is merged with #981.

PR #900, opened 2019-04-17
Add explicit

jonaski:


scotchi:
This can't be done in the master branch as it changes the API in an incompatible way.

scotchi:
Actually -- I hadn't read closely enough to see that these are all in private classes ... what's the benefit then of adding `explicit` in private classes?

PR #907, opened 2019-05-14
Fix unintentional id3v1 creation

smithjd15:
Fix unintended creation of id3v1 tags when generically saving or setting tag properties.

### DIFF
--- a/taglib/mpeg/mpegfile.cpp
+++ b/taglib/mpeg/mpegfile.cpp
@@ -110,7 +110,7 @@ bool MPEG::File::isSupported(IOStream *stream)
 
   if(buffer.isEmpty())
 	  return false;
-  
+
   const long originalPosition = stream->tell();
   AdapterFile file(stream);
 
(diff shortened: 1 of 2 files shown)


sbooth:
Based on the changes to the documentation in this PR I don't think the ID3v1 tag creation is unintentional as far as `MPEG::File::save()` is concerned. Could you show a snippet exhibiting the problem this PR solves?

smithjd15:
Fileref is created and id3v2 is read.
Tag::Id3v2 is modified.
Fileref::save is called.
File ends up with an id3v1 from the fileref::save method because the tag is duplicated into id3v1. 

Either way, calling into mpeg::file can work around the fileref::save shortcomings, but the generic fileref::save method shouldn't create an obsolete tag type.

scotchi:
This is actually the intended behavior.  There are still sadly a handful of things that either can't read ID3v2 or ID3v2.4, so duplicating the contents into an ID3v1 tag is intentional.

Issue #913, opened 2019-07-17
MP3 strip bug


whatdoineed2do:
Update: Fix in pr https://github.com/taglib/taglib/issues/914

There is MP3 stripping bug if you have a file with both APE and ID3v1 tag.

If you try to strip using `strip(TagLib::MPEG::File::ID3v1 | TagLib::MPEG::File::APE);` or calling `strip(TagLib::MPEG::File::ID3v1)` *before* `strip(TagLib::MPEG::File::APE)` then saving the file results in the ID3v1 untouched.

However if your file does not have APE tag then this works as expected.

Here is the code to demonstrate:
```
#include <iostream>
#include <taglib/mpegfile.h>

int main(int argc, char *argv[])
{
    if (argc != 3) {
        return 1;
    }
    try
    {
        TagLib::MPEG::File  f(argv[2], false);

	switch (*argv[1])
	{
            case '0':
// this is the only item that works
	      f.strip(TagLib::MPEG::File::APE);
	      f.strip(TagLib::MPEG::File::ID3v1);
	      break;

// these leave the ID3v1 in the file when there is APE + ID31
// if there is ONLY id3v1 then this is removes the id3v1 tag
            case '1':
	      f.strip(TagLib::MPEG::File::ID3v1);
	      f.strip(TagLib::MPEG::File::APE);
	      break;

	    default:
	      f.strip(TagLib::MPEG::File::ID3v1 | TagLib::MPEG::File::APE);
	}

        f.save(TagLib::MPEG::File::ID3v2);
    }
    catch (const std::exception& ex)
    {
        std::cerr << ex.what() << std::endl;
    }
    return 0;
}
```

Moonbase59:
Unfortunately, this is also true when using "stripOthers" in f.save() like so:
```c
f.save(TagLib::MPEG::File::ID3v2, true, 3);
```

Moonbase59:
Possible duplicate of #835

whatdoineed2do:
Ah yes. This is indeed the same - I had forgotten that I raised the same issue 2 years previous!

Moonbase59:
Apart from a few disastrous bugs, taglib is a great library and used by many (I think I remember EasyTag, Clementine, Mixxx, loudgain, …). With all this groundwork laid out, it’d be a shame to see it go under—or maintainers having to patch it themselves.

I’d REALLY love to see a new fixed version to be picked up by the distros, because end users simply can’t be expected to fix things and compile it themselves!

[...]

PR #914, opened 2019-07-21
tfilestream truncate/read bug - MP3 stripping fix

whatdoineed2do:
Closes https://github.com/taglib/taglib/issues/913.

The `fread(3C)` function potentially buffers the input stream and needs `fflush` following a `ftruncate(2)` being called; ie
```
FILE* f = fopen(...);
ftruncate(fileno(f), 10);
// fflush(f);  -- this is needed here
fseek(f, 10, SEEK_SET);  // seek to where we've truncated the file
fread(....);   // this will read data that has been truncated but buffered
```
These functions are used by code by `tfilestream` for some file operations, in particular the stripping of MP3 tags.

The missing `fflush` causes a problem with MP3 stripping of APE and ID3v1 tags - these tags reside at the end of the file and in this order if both exist.

Stripping functionality for `APE` tags `seek` to the end of the tag (based on `APELocation` and `APEOriginalSize`) and performs a _blind_ `fread()` until it reaches EOF, assuming it to be the ID3v1 tag.

Striipping functionality for `ID3v1` tag simply performs a `ftruncate()` from the start of the tag position (`ID3v1Location`).

The bug can be observed by:
```
    f.strip(ID3v1);
    f.strip(APE);  // the ID3v1 will be untouched/not stripped
```
This means current master executes:
* truncate file to ID3v1 pos
* seek to end of APE tag pos and `fread()` til end of file; the data returned here is potentially buffered  (ie reads the ID3v1 tag that was just truncated/stripped)
* rewrite all the read data (ie the stripped ID3v1) at the APE tag position

This commit performs a simple `fflush()` following the `ftruncate()` to discard any buffered data to ensure no phantom data being returned.

The [POSIX defintion](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fflush.html) for `fflush()` defines the behavior for input streams, whilst the latest draft [C2x standard](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2346.pdf) leaves it undefined.

Tested on Linux/Fedora 26.

### DIFF
--- a/taglib/toolkit/tfilestream.cpp
+++ b/taglib/toolkit/tfilestream.cpp
@@ -493,6 +493,7 @@ void FileStream::truncate(long length)
 
 #else
 
+  fflush(d->file);
   const int error = ftruncate(fileno(d->file), length);
   if(error != 0)
     debug("FileStream::truncate() -- Coundn't truncate the file.");


sbooth:
Thank you for this fix.

Out of curiosity if `fflush()` is called before `ftruncate()` instead of after do you see a difference in behavior?

whatdoineed2do:
Calling `fflush()`before the `ftruncate()` also works to fix this bug.

Here's a test prog to demo the issue: `a.out /tmp/data 0` for no `fflush`, change 0 to 1 to observe the fix.
```
#include <sys/stat.h>
#include <sys/types.h>
#include <fcntl.h>
#include <unistd.h>
#include <sys/types.h>
#include <string.h>
#include <stdio.h>
#include <errno.h>



int main ( int argc, char *argv[])
{
    char buf[64];

    if (argc != 3) {
	printf("args: <tmpfile> <0 or 1 to call fturncate or not>");
	return -1;
    }

    int fd;

    if ((fd = open (argv[1], O_TRUNC | O_CREAT | O_WRONLY, 0600)) < 0) {
	printf ("failed to create/trunc file '%s' - %s\n", argv[1], strerror (errno));
	return -1;
    }


    const char *data = "0123456789abcdef";

    write (fd, data, strlen (data));
    fsync (fd);
    close (fd);
    fd = -1;

    FILE *f = fopen (argv[1], "rb+");

    if (f == NULL) {
	printf ("failed to open data file\n");
	return -1;
    }

    printf ("data file = '%s'\ntruncating at pos=10, remain = '%s'\n", data, &data[10]);
    fread (buf, 1, 8, f);  // prime the internal stream buffer

    fseek (f, 10, SEEK_SET);
    ftruncate (fileno (f), 10);
    if (*argv[2] == '1') {
	printf ("   flush FILE, will see following read contain NO data \n");
	fflush (f);
    }
    else {
	printf ("NO flush FILE, will see following read contain data\n");
    }

    int r = fread (buf, 1, sizeof (buf), f);
    buf[r] = '\0';
    printf ("fread %ld - '%s'\n", r, buf);
    fclose (f);

    return 0;
}
```

sbooth:
Since `fflush()` before `ftruncate()` works correctly would you mind to amend the commit with the change? Without combing through the uses of `tfilestream` I am not sure if this could happen in TagLib, but I think it's possible that a pending `fwrite()`could have buffered unwritten changes that `ftruncate()` might alter. Switching the order of the calls will eliminate that possibility.

whatdoineed2do:
It makes sense.  I've made the change and pushed.

PR #917, opened 2019-08-09
Calculate Ogg bitrate without comment size (#874)

### LINKED ISSUE #874: The reported bitrate of Ogg files is different after adding metadata
The problem occurs with opus, ogg and spx-files. I am using an opus-file in the following example.

When adding metadata to an opus file, the reported bitrate gets larger. It would be more useful if the bitrate was independent of the overhead. I have used the following script from within a TagLib build directory.

```
#!/bin/sh
set -e
file1=$1
file2=/tmp/test.${file1##*.}
echo "Using file $file1"
cp "$file1" "$file2"
examples/tagreader "$file1" | grep bitrate
comment=$(seq -s '' 10000)
echo "Adding a comment with ${#comment} bytes to $file2"
examples/tagwriter -c $comment "$file2"
examples/tagreader "$file2" | grep bitrate
```

A test file has to be passed, which is copied to /tmp/ and a comment with 37 KBytes is added. Such a comment is not realistic, but it could also be an attached picture, which would have the same effect.

The script reports

```
$ /tmp/bitratetest.sh ../taglib-git-source/tests/data/correctness_gain_silent_output.opus
Using file ../taglib-git-source/tests/data/correctness_gain_silent_output.opus
bitrate     - 37
Adding a comment with 38894 bytes
bitrate     - 77
```

opusinfo reports the following bitrate:

```
$ opusinfo ../taglib-git-source/tests/data/correctness_gain_silent_output.opus | grep bitrate
        Average bitrate: 36.71 kb/s, w/o overhead: 35.53 kb/s
$ opusinfo /tmp/test.opus | grep bitrate
        Average bitrate: 77.1 kb/s, w/o overhead: 35.53 kb/s
```

I have used the script with various formats supported by TagLib and the bitrate is wrong for opus-, speex- and ogg-files. The other formats show the same bitrate before and after adding new metadata.

ufleisch:
This commit fixes #874 by subtracting the size of the XiphComment header data when calculating the bitrate for audio formats which use an Ogg container (Vorbis, Opus, Speex). The bitrate calculated is not as accurate as when decoding the whole stream (and then being able to subtract the whole overhead), but still accurate and fast enough. The reported bitrate will no longer be completely wrong when files contain large metadata, for example when embedding cover art.

### DIFF
--- a/taglib/ogg/opus/opusproperties.cpp
+++ b/taglib/ogg/opus/opusproperties.cpp
@@ -163,8 +163,14 @@ void Opus::Properties::read(File *file)
 
       if(frameCount > 0) {
         const double length = frameCount * 1000.0 / 48000.0;
+        long fileLengthWithoutOverhead = file->length();
+        // Ignore the two mandatory header packets, see "3. Packet Organization"
+        // in https://tools.ietf.org/html/rfc7845.html
+        for (unsigned int i = 0; i < 2; ++i) {
+          fileLengthWithoutOverhead -= file->packet(i).size();
+        }
         d->length  = static_cast<int>(length + 0.5);
-        d->bitrate = static_cast<int>(file->length() * 8.0 / length + 0.5);
+        d->bitrate = static_cast<int>(fileLengthWithoutOverhead * 8.0 / length + 0.5);
       }
     }
     else {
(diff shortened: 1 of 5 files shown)


scotchi:
I just was re-reading the spec for Vorbis for the first time in about 1000 years to review this request, and wondered if there was a reason for just ignoring the comment packet, instead of all three of the setup packets at the beginning of a Vorbis file?  (For now, I'll assume that the other formats also have three header packets.)

ufleisch:
Thanks for following up. I had a look at the Ogg, Vorbis, Opus and Speex format specifications. Vorbis uses the first three packets for initial header, comments and setup whereas Opus and Speex only use the first two packets for overhead. I will amend my commit to ignore this overhead too when calculating the bitrate.

PR #918, opened 2019-08-15
Update mp3 id3 cover art example

kdeng00:
Added an example to update an mp3 song's front cover art. It's a simple program where the two arguments are required - song path and image path. A check is done to find any cover art found the the id3 tag and then remove the frame from the id3 tag. Then the specified image path will be applied to the cover art.

I switched over to TagLib recently and though it might help someone else if they wanted to see an example on how to update the cover art.


ufleisch:
This is now documented in [tag.h](https://github.com/taglib/taglib/blob/master/taglib/tag.h) and [tag_c.h](https://github.com/taglib/taglib/blob/master/bindings/c/tag_c.h). Examples can be found in [tagwriter.cpp](https://github.com/taglib/taglib/blob/master/examples/tagwriter.cpp), [tagreader.cpp](https://github.com/taglib/taglib/blob/master/examples/tagreader.cpp), and [tagreader_c.c](https://github.com/taglib/taglib/blob/master/examples/tagreader_c.c).

PR #930, opened 2019-10-04
MP4 - setAlbum/Artist/Genre/Comment() accepts empty string to remove tag

### LINKED ISSUE #929: M4a - setXXX() does not remove tag when param is String()
A `String()` or `String::null` does not remove tag - similar bug to #911

whatdoineed2do:
Closes #929 

As per documentation/functionality across other tpyes (mp3/flac/...); m4a do not honour removing tags when input param is `String()` or `String::null`


ufleisch:
Thanks for the PR, is now fixed using #984.

PR #932, opened 2019-10-05
FLAC - lookup on non existant tags do not inadvertently create entry in FieldListMap

### LINKED ISSUE #931: FLAC, APE - get on empty tag (title/album/...) will create an empty value in properties map
FLAC:  Querying a common tag value (like title, album, artist etc) that does not exist will create an empty value in the properties map.  This does not happen for mp3,m4a files.

Code to demonstrate:
* force removal of `title` and save file
* print property map shows no `TITLE`
* reopen file and query value of `title`
* show property map before/after call `title()` shows empty `TITLE` entry (**bug**) 
```
#include <iostream>
#include <taglib/mpegfile.h>
#include <taglib/flacfile.h>
#include <taglib/xiphcomment.h>
#include <taglib/tpropertymap.h>

std::ostream& operator<<(std::ostream& os_, const TagLib::PropertyMap&  m)
{
    for (const auto i : m)
    {
        os_ << i.first.toCString() << " {\n";
        for (const auto& j : i.second) {
            os_ << "  " << j << '\n';
        }
        os_ << "}\n";
    }
    return os_;
}


int main(int argc, char *argv[])
{
    if (argc != 2) {
        return 1;
    }
    try
    {
        {
            TagLib::FLAC::File f(argv[1]);
            f.xiphComment()->removeFields("TITLE");
            std::cout << "FIRST (removed title)\n" << f.properties() << "\n";
            f.save();
        }
        {
            TagLib::FLAC::File f(argv[1]);
            std::cout << "SECOND\n" << f.properties() << "\n";
        }
        {

          TagLib::FLAC::File f(argv[1]);
          std::cout << "THIRD\n" << f.properties() << "\n"
                    << "title = '" <<  f.tag()->title() << "'\n"
                    << "FOURTH\n" << f.properties() << "\n";
        }
    }
    catch (const std::exception& ex)
    {
      std::cerr << ex.what() << std::endl;
    }
    return 0;
}
```

whatdoineed2do:
Fixes #931 

``` 
           TagLib::FLAC::File f(argv[1]);
            f.xiphComment()->removeFields("TITLE");
            std::cout << "FIRST (removed title)\n" << f.properties() << "\n";
            f.save();

            f.tag()->title();
```
Whilst `title()` returns empty string, the `Ogg::XiphComment` interface will create an empty entry in the `FieldListMap` for `TITLE`.  This is inconsistent with MP3/MP4 interfaces in the library.

This behaviour is seen for `TITLE`, `ARTIST`, `ALBUM`, `GENRE`, `COMMENTS`/`DESCRIPTION` which are fixed with this PR

### DIFF
--- a/taglib/ogg/xiphcomment.cpp
+++ b/taglib/ogg/xiphcomment.cpp
@@ -78,35 +78,34 @@ Ogg::XiphComment::~XiphComment()
 
 String Ogg::XiphComment::title() const
 {
-  if(d->fieldListMap["TITLE"].isEmpty())
-    return String();
-  return d->fieldListMap["TITLE"].toString();
+  StringList value = d->fieldListMap.value("TITLE");
+  return value.isEmpty() ? String() : value.toString();
 }
 
 String Ogg::XiphComment::artist() const
 {
-  if(d->fieldListMap["ARTIST"].isEmpty())
-    return String();
-  return d->fieldListMap["ARTIST"].toString();
+  StringList value = d->fieldListMap.value("ARTIST");
+  return value.isEmpty() ? String() : value.toString();
 }
 
 String Ogg::XiphComment::album() const
 {
-  if(d->fieldListMap["ALBUM"].isEmpty())
-    return String();
-  return d->fieldListMap["ALBUM"].toString();
+  StringList value = d->fieldListMap.value("ALBUM");
+  return value.isEmpty() ? String() : value.toString();
 }
 
 String Ogg::XiphComment::comment() const
 {
-  if(!d->fieldListMap["DESCRIPTION"].isEmpty()) {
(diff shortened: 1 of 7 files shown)


taglib/ogg/xiphcomment.cpp
```diff
@@ -40,6 +40,19 @@ namespace
   typedef List<FLAC::Picture *> PictureList;
   typedef PictureList::Iterator PictureIterator;
   typedef PictureList::Iterator PictureConstIterator;
+
+  /* returns the tag value for the given map[key] - however it ensure that map
```
whatdoineed2do: NB: `bool Ogg::XiphComment::contains()` will still create entries for non exist elements - to avoid breaking that contract we use `find(...)` to determine existance instead of _fixing_ and using `contains()`

PR #935, opened 2019-10-19
taglib-config.cmake has static libdir and includedir variables

jiblime:
This matches the syntax of the 1.12 beta release and fixes issues where a user's libdir is not at ${exec_prefix}/lib but ${exec_prefix}/lib64 or otherwise.


ufleisch (approved):
Look good to me.

PR #939, opened 2019-11-05
clang-tidy fixes

neheb:
I tried to avoid the ones that require C++11 and up. I think? this has support for C++03 and below.


neheb:
Fixed the test failure. I tried to be clever.

Specifically, clang-tidy throws https://clang.llvm.org/extra/clang-tidy/checks/bugprone-incorrect-roundings.html in several places.

I tried replacing x + 0.5 with round, floor, and ceil, but they all failed the test. No idea what the issue is there.

neheb:
Oh I see my mistake.

PR #942, opened 2019-12-19
Fix spelling of Bebop and update Fusion and Hardcore to match Wikipedia

uqs:
Source: https://en.wikipedia.org/wiki/List_of_ID3v1_Genres

Similar patches will be sent to libid3tag and ffmpeg to harmonize the
genre names and spellings.


uqs:
This is what I sent to FFmpeg
https://github.com/FFmpeg/FFmpeg/pull/328

sbooth:
Is https://en.wikipedia.org/wiki/List_of_ID3v1_Genres the presumptive normative reference for ID3v1 genres? Some of the changes in this PR don't match that page.

ufleisch:
I do not think that a normative reference for ID3v1 genres exists. I would suggest that commonly used genre names should be used so that the names mapped to ID3v1 genre codes match the names used in other tag formats where arbitrary names are possible in order to make it easier for the users of the tagged files to filter them by genre. The genre names in https://en.wikipedia.org/wiki/List_of_ID3v1_Genres do not even match all the genre names used in the corresponding Wikipedia articles about the corresponding genre. I only made changes to names where I think that they are misspelled, but I am not a native English speaker, so please correct them if I am wrong. And I also changed the name of the "racist joke in Swedish" (as so designated in https://en.wikipedia.org/wiki/ID3).

sbooth:
I'm sure you're right about the non-existence of a normative reference. Regarding the spelling, I certainly wouldn't have guessed you weren't a native speaker. They all look good to me.

scotchi:
There's a partial list in the ID3v2 standard, and I think past that Winamp is basically considered to be authoritative.

scotchi:
Something else potentially worth considering:  changing these spellings will change the behavior of current apps since these names are reverse mapped onto ID3v1 genre numbers -- so in places where that's done, if the old text value is used, it now won't be recognized.  I suspect some apps have genre drop downs that this may break.

uqs:
Can you point to such programs and how they use taglib? Please note that the spelling is already fixed in ffmpeg and I'm not sure there was any fallout there.

If programs deal with free-form text like "Bebob", why would they map that to ID3v1 instead of using v2 directly? Makes little sense to me.

scotchi:
There are hundreds of programs that use TagLib, many of them closed source.  I'm not even arguing against this patch -- just noting potential incompatibility that it introduces.  With this patch applied, this code would work differently:

`file.setGenre("Avantegarde");`

I think it would probably make sense to keep the present spellings in a fix-up array that maps them onto the new spellings.

ufleisch:
I have added such a fix-up array to improve compatibility.

PR #947, opened 2020-02-19
Replacing space with other characters for repeated tags

scbash:
Hi all.  I've been working on a [Gerbera media server](https://github.com/gerbera/gerbera) setup for the past few weeks, and one of the things I discovered is a small percentage (but more than I want to retag) of my FLAC files have repeated tags, e.g. `Artist=Artist1`, `Artist=Artist2`, etc.  Since the default behavior of TagLib 1.11.1 is to concatenate repeated tags with spaces I ran into difficulty splitting the artists apart in Gerbera.  Since `StringList::toString()` already takes a separator argument, over this past weekend I added a similar argument to `Tag::genre()`, `Tag::artist()`, `Tag::comment()`, `Tag::title()`, and `Tag::album()` -- honestly genre and artist are the pain points for me, but I can go either way on the others.

Note MP4 was an outlier in that it already passed ", " to `StringList::toString()`, so I kept that as the default value.

I saw several comments about this area of code potentially changing in TagLib2, so I don't know if this change is valuable or not?  Feel free to reject this if it doesn't fit what you're going for.  And to be honest I did all this work against 1.11.1, not master.  I did my best to follow local conventions in the code, but commit messages are 100% my style -- if you'd like changes let me know.  I'm also happy to squash these three into one if it's a better fit.  

Finally there's also the weird question about what's going on with the default argument in `TagUnion` (see the last commit, fc5ac34f9fa7666b9fbfa141ca84e398f9576e33), but it does work for me.

### DIFF
--- a/taglib/ape/apetag.cpp
+++ b/taglib/ape/apetag.cpp
@@ -115,39 +115,39 @@ ByteVector APE::Tag::fileIdentifier()
   return ByteVector::fromCString("APETAGEX");
 }
 
-String APE::Tag::title() const
+String APE::Tag::title(const String &separator) const
 {
   if(d->itemListMap["TITLE"].isEmpty())
     return String();
-  return d->itemListMap["TITLE"].values().toString();
+  return d->itemListMap["TITLE"].values().toString(separator);
 }
 
-String APE::Tag::artist() const
+String APE::Tag::artist(const String &separator) const
 {
   if(d->itemListMap["ARTIST"].isEmpty())
     return String();
-  return d->itemListMap["ARTIST"].values().toString();
+  return d->itemListMap["ARTIST"].values().toString(separator);
 }
 
-String APE::Tag::album() const
+String APE::Tag::album(const String &separator) const
 {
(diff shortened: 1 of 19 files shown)


ufleisch:
This would break binary compatibility. Besides this, I do not think that it is a good idea. If you seriously want to deal with tags having multiple values, I would suggest to use the specialized API of the corresponding tag format. Vorbis Comments, as they are probably used in your FLAC files, can have multiple key-value pairs with the same key whereas things are more restricted for ID3v2: There must be only one title (TIT2), album (TALB), artist (TPE1). There can be multiple comments (COMM), but with different language/content descriptor. The genre (TCON) is defined to be a list, which clearly defines zeros as separators. Therefore, the simple unified tag API is better left in its current state with single values.

scbash:
Thanks for the feedback Urs.  Personally I think tying the general API to an aging standard like ID3v2 is very limiting (ideally I'd prefer to always return a list of tag values and if that list happens to have a single entry so be it), but I'll investigate what it would take to patch Gerbera to use the specialized API instead.

Issue #951, opened 2020-04-05
[flac]Request for Making a copy of flac picture list to the existing tags


khjxiaogu:
_Taglib2_ has an extracted picture interface for most media types. However, getting flac picture list is still a little bit troublesome because the program have to check if the file is flac, then using another method to get pictures. This is confusing. Could there be another Tag to provide picture list from flac?

ufleisch:
Implemented with #1156.

PR #952, opened 2020-04-05
added flac picture tag

khjxiaogu:
As said in issue #951 this is used to add a picture interface for flac picture list


ufleisch:
Implemented with #1156.

PR #958, opened 2020-05-12
Perform a more complete search for MPEG frame sync before declaring file 'not supported'

timusus:
Previously, TagLib only searched the first 1024 bytes (bufferSize()) of the file for the 'frame sync' bytes, which indicate the start of an ID3v2 Tag / MPEG Frame header. In some (arguably corrupt) MPEG files, the frame sync may be located further into the file.

This change instead traverses the entire file in search for the frame sync.


ufleisch:
Can you confirm that this is fixed in the current implementation of `MPEG::File::findID3v2()`?

PR #960, opened 2020-06-04
clang-tidy fixes

neheb:


scotchi:
TagLib doesn't use C++11, which most of these "fixes" assume.

scotchi:
To be clear, even most of the non-C++11 changes I wouldn't commit.  There's no reason to mess up the git history for changes that don't actually improve the code in any meaningful way.

neheb:
Oh I sent the wrong branch. Will post the right one.

neheb:
@scotchi compiling with -std=c++98 fails:

```
  5%] Building CXX object taglib/CMakeFiles/tag.dir/toolkit/trefcounter.cpp.o
In file included from /usr/include/c++/8/atomic:38,
                 from /home/mangix/devstuff/taglib/taglib/toolkit/trefcounter.cpp:33:
/usr/include/c++/8/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
 #error This file requires compiler and library support \
```

Issue #962, opened 2020-06-27
Wavpack DSD files show wrong duration


the-black-eagle:
When attempting to get the duration of a wavpack DSD file, taglib returns a duration that is roughly 4 times what it should be. For instance, for the Genesis track Hairless Heart, the duration should be 2:09 or thereabouts.  Taglib returns a duration of 8:41 for this track. 
This issue was first raised on the Kodi forum (https://forum.kodi.tv/showthread.php?tid=355493) as Kodi uses taglib for the majority of it's tag reading.  I would therefore expect that this could impact a large number of users given that Kodi can now play these files.

I can supply a test file if required.

dbry:
I took a quick look at the taglib WavPack code and the problem is that it only seems to examine the 32-byte WavPack header. This is sufficient for most things, but there are a couple cases where a little more work needs to be done to determine the actual sample rate of the file (and it's the sample rate that's wrong here, which of course affects the reported duration).

The first case is when a "non-standard" sampling rate is used and the other is for DSD audio. I recently provided a [patch for MKVToolNix](https://gitlab.com/mbunkus/mkvtoolnix/-/merge_requests/2204) for this exact issue (plus another issue involving checksums) and all the information and code is there.

I'll see if I can get a chance to create a patch for taglib to do this. As I mentioned in the other thread I'm not a strong C++ programmer, so hopefully someone else can carefully review it. Or, there's enough code in the other patch to allow someone else to do this, but I'm thinking that the WavPack code in taglib is old enough that nobody is familiar with it any more.

Thanks!

ufleisch:
Fixed with #963.

PR #963, opened 2020-07-01
WavPack fixes with test file verifying issue #962

dbry:
Fixes to properly handle WavPack files with leading and trailing non-audio blocks, non-standard sampling rates, and DSD audio (introduced in WavPack 5). Also add WavPack DSD file to tests for issue #962.


sbooth:
Thank you for this! I will try and take a closer look in the next few days and merge the PR.

dbry:
I have about 10 files that I used to test out these changes, although they're about 3 MB each which is why I didn't consider adding them to the regular tests. I might be able to upload them here (not sure about size limits).

dbry:
Is there anything I can do to help this along?

taglib/wavpack/wavpackproperties.cpp
```diff
@@ -212,25 +253,126 @@ void WavPack::Properties::read(File *file, long streamLength)
...
+        sample_rate |= static_cast<int32_t>(*buffer++) << 8;
+        sample_rate |= static_cast<int32_t>(*buffer++) << 16;
+
+        // only use 4th byte if it's really there
+
+        if (!(meta_id & ID_ODD_SIZE))
+          sample_rate |= static_cast<int32_t>(buffer[3] & 0x7f) << 24;
+
+        return sample_rate;
+      }
```
ufleisch: The indentation here is two characters too much.

taglib/wavpack/wavpackproperties.cpp
```diff
@@ -212,25 +253,126 @@ void WavPack::Properties::read(File *file, long streamLength)
...
+      meta_bc += (static_cast<uint32_t>(c1) << 9) + (static_cast<uint32_t>(c2) << 17);
+      bcount -= 2;
+    }
+
+    if (bcount < meta_bc)
+      return 0;
+
+    // if we got a sample rate, return it
+
+    if ((meta_id & ID_UNIQUE) == ID_SAMPLE_RATE && meta_bc == 4) {
```
ufleisch, line 319: I do not know much about WavPack, but this seems a bit suspicious. `bcount` is at least `meta_bc`, which is `4`, but `buffer` is incremented here three times, and finally `meta_bc` is added, which gives 7 and this is larger than the guaranteed size of 4. `buffer[3]` would also access beyond the `buffer` if its size is only 4, because at this time, it was already incremented by `3`. Maybe you wanted to access `buffer[0]`, `buffer[1]`, `buffer[2]` instead of the three `*buffer++`?

taglib/wavpack/wavpackproperties.h
```diff
@@ -139,6 +139,8 @@ namespace TagLib {
 
       void read(File *file, long streamLength);
       unsigned int seekFinalIndex(File *file, long streamLength);
+      int getNonStandardRate(unsigned char const *buffer, int bcount);
+      int getDsdRateShifter(unsigned char const *buffer, int bcount);
```
ufleisch, line 143: These two methods to not access other members of the class. I would make them local to the *wavpackproperties.cpp* file and put them into a `namespace {}` which has to be below the block with the `#define`.

ufleisch (changes requested):
The changes in your last commit "correctly read very high sample rates from WavPack files" confirm my suspicion expressed in one of my comments. Please give me feedback before I can merge this pull request.

taglib/wavpack/wavpackproperties.cpp
dbry: Hi, and thanks for reviewing this! The reason this is okay is that there's an unconditional return at the end of the block. We're guaranteed to have at least 4 bytes and we read the 3 or 4-byte sample rate and return with it because that's all this function does. As you say, the `buffer += meta_bc` would be wrong, but it's never reached. Using `buffer[3]` was also wrong (I copied it from other code without enough context) but fixed it once it failed the test. Of course, that's only for sample rates over 16 MHz, so wouldn't be hit too often!

As for the other fixes, I'll get them in. My C++ is really rusty, but I think I can figure it out (or get someone to help me). Feel free to do it yourself if you like.

ufleisch: You are right, there is a `return` statement following the mentioned pointer arithmetic, so it does not matter. My other comments are only minor details, so I think it is OK as it is, and I could do the minor stuff when merging.

dbry: Okay, sounds good, and thanks again.

It will be great to get this in!

scotchi:
Looking through this, I also find suspect the usage of `char *` pointers and pointer arithmetic instead of `ByteVector` and indexes suspect.  I'd also prefer variables to be initialized on declaration and variable naming (underscores vs. camelCase) to be consistent with other bits of TagLib.

ufleisch (approved):

ufleisch:
I have added two more commits to this PR, the first addressing the things mentioned by @scotchi in https://github.com/taglib/taglib/pull/963#issuecomment-748617259, and also removing some duplicated code. The second commit adds a test for non-standard sample rates as I have seen that this case was not covered by a test yet. I have generated the test file simply using `wavpack tests/data/empty.wav -o tests/data/non_standard_rate.wv`.
@dbry Could you please have a look at the changes and test if everything is still OK?

dbry:
Thanks so much for doing this; I was not looking forward to getting out my C++ textbook!

I will definitely test this out locally.

BTW, I noticed that the [PR fails Travis now](https://travis-ci.org/github/taglib/taglib/builds/751014049). Not sure what's going on there.

ufleisch:
Travis failure: This could have something to do with the newly added support for PPC. It fails on PPC because `uint32_t`, `int32_t` are not found. I tried to fix it including *cstdint*, but then I get a "C++11 required" error. I have now changed it to include *stdint.h*, and it is still running for 20 minutes, *wavpackproperties.cpp* should have passed, maybe it will succeed this time...

dbry:
I pulled in these changes and verified that all of the test files I had created for this effort still work correctly, plus all the files in the standard WavPack test suite.

I also looked over the code changes and everything seems to be correct.

Thanks again @ufleisch and the other reviewers for your help on this!

PR #967, opened 2020-08-07
support for matroska files

sandsmark:


ufleisch:
First I ran the test suite, it crashes in `TestFileRef::testUnsupported()`, constructs `File` with `d=0`, which is then dereferenced in the destructor.

To test interoperability with other tools, I generated tags with *HandBrake*, the tags are all shown in *MKVToolNix* (Title, ARTIST, DIRECTOR, DATE_RELEASED, SUMMARY, GENRE, DESCRIPTION, SYNOPSIS, ENCODER). When I open the mkv file with *examples/tagreader*, it only shows ARTIST, GENRE. Writing with *examples/tagwriter* and reading back with *examples/tagreader*, the fields set are displayed except album. With *MKVToolNix* and *HandBrake*, again only ARTIST, GENRE seem to be "interoperable". But *MKVToolNix* now shows in another tag, three times COMMENT, DATE_RELEASE, PART_NUMBER, TITLE, and in yet another tag again three times TITLE. The Tags section also shows an element "EBML void" with size 280. After setting new values with *tagwriter*, two more "EBML void" elements are shown, (size 404 and size 530). Deleting tags using empty values with *tagwriter* does not work, it just creates even more "EBML void" elements.

This all reminds me of my experience more than two years ago with the Matroska code in Taglib2 (probably, it is the same code). The positive thing is that this time, I did not manage to destroy my MKV file when modifying it with TagLib (or *MKVToolNix* has become more tolerant).

One part of the problem is the implementation which seems to be unfinished and incomplete. It is not easy to integrate because Matroska-Tags are a bit different. They use different target types - to use the same element at different levels (e.g. track, album), and nested tags - to use elements in a different context (e.g. original, sort). Then there can be multiple such element trees.

All this would have to be mapped correctly into the model used by TagLib, but it supports just a few elements correctly and hides all the rest. It also seems to generate duplicate information and is not able to delete elements - they are just set to a void type. I would at least expect that a file can be reverted to its original state when tags are added and later removed, just as a simple test, that nothing is corrupted.

NDDN85:
gh pr checkout 967

neheb:
@whyman exiv2 can be used for gerbera I think.

ufleisch:
Support for Matroska has been added to TagLib #1149, so this PR can be closed.

Issue #920, opened 2019-08-31
Runtime version check?


Moonbase59:
There should be a runtime (API) "get version" call, to determine which taglib we’re using when using shared objects or DLLs.

Since there wasn’t a new version for some time, but some bugs have been fixed, it might even make sense to also return the commit hash from which it has been built, maybe optional. Programs like loudgain and others could thus better work around known bugs (like #864).

scotchi:
Out of curiosity, what's the use case that you imagine runtime version information being useful in? (It is available at build time.) I'm not aware of many cases where a program is built against a different version of TagLib than the one that it's using at runtime (though it is possible, and TagLib maintains binary compatibility to support such).

Moonbase59:
Just take the case above (#864): My loudgain ReplayGain tagger should be sure not to corrupt Ogg Vorbis files. Mixxx already blocks TagLib 1.11.1 for writing tags …

With shared libraries, we must be sure that the use can (and will) update these independently from the application, and some distros still ship buggy versions.

PR #970, opened 2020-09-02
Add simple runtime version check API

hyperquantum:
This adds a simple runtime version check as requested in #920 .

How to use:

```
  #include <taglib.h>

  if (TagLib::isTagLibVersionAtLeast(1, 11, 1)) {
    // running v1.11.1 or higher
  }

  if (TagLib::isTagLibVersionAtLeast(1, 12)) {
    // running v1.12 or higher
  }
```

### DIFF
--- a/taglib/CMakeLists.txt
+++ b/taglib/CMakeLists.txt
@@ -297,6 +297,7 @@ set(xm_SRCS
 )
 
 set(toolkit_SRCS
+  toolkit/taglib.cpp
   toolkit/tstring.cpp
   toolkit/tstringlist.cpp
   toolkit/tbytevector.cpp
--- a/taglib/toolkit/taglib.cpp
+++ b/taglib/toolkit/taglib.cpp
@@ -1,0 +1,43 @@
+/***************************************************************************
+    copyright            : (C) 2020 by Kevin Andre
+    email                : hyperquantum@gmail.com
+ ***************************************************************************/
+
+/***************************************************************************
+ *   This library is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU Lesser General Public License version   *
+ *   2.1 as published by the Free Software Foundation.                     *
+ *                                                                         *
+ *   This library is distributed in the hope that it will be useful, but   *
+ *   WITHOUT ANY WARRANTY; without even the implied warranty of            *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU     *
+ *   Lesser General Public License for more details.                       *
+ *                                                                         *
+ *   You should have received a copy of the GNU Lesser General Public      *
+ *   License along with this library; if not, write to the Free Software   *
+ *   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA         *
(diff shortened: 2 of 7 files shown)


xzyangNB (approved):

hyperquantum:
Why has the label "taglib2" been applied to this PR? These changes were intended for version 1.

ufleisch:
I removed the label "taglib2", since such a feature would not break binary compatibility. The reasoning was that it might be not so useful when introduced in a late 1.x version.

What I do not like about the suggested solution is that it covers exactly one use case, namely "is version at least". What if someone wants to know the exact version? So I would suggest a simpler API: Either a simple `int tagLibVersion()` function, which returns the TagLib version with 8 bits for major, minor and patch version `(TAGLIB_MAJOR_VERSION << 16) + (TAGLIB_MINOR_VERSION << 8) + TAGLIB_PATCH_VERSION` or a function returning the version as an object like [`QLibraryInfo::version()`](https://doc.qt.io/qt-6/qlibraryinfo.html#version), but this might be a bit too much.

hyperquantum:
I initially only provided that one use case because:
1. It seemed sufficient for doing compatibility checks.
2. I was (and still am) reluctant of exposing the 'major', 'minor' and 'patch' components separately, because version number checks are easy to get wrong: `if (major < 3 || minor < 11) { /* error: we require version 3.11 or higher */ }`  [source](https://devblogs.microsoft.com/oldnewthing/20040213-00/?p=40633)

A while after I created this PR I checked out the taglib2 branch and discovered that it already has [some support](https://github.com/taglib/taglib/blob/taglib2/taglib/toolkit/taglib.h#L59) for getting the version numbers. Could this be reused? However I am not in favor of returning the TagLib version as one big number, it seems an inconvenient API to use. Suppose one wants to check if the version is 1.11.1 this way:

- `if (tagLibVersion() == ((1 << 16) | (11 << 8) | 1))` -- bit twiddling, ugh, not easy to read
- `if (tagLibVersion() == 0x010b01)` -- hex, easy to get wrong, I almost wrote 0x011101 which means 1.17.1

What do you think about extending the existing code from this PR to something like this:
```
namespace TagLib {
  namespace Version {
    bool isAtLeast(int major, int minor);
    bool isAtLeast(int major, int minor, int patch);
    bool equals(int major, int minor, int patch);
    String asString(); /* returns <major>.<minor>.<patch>, e.g. "1.12.0" */
  }
}
```
Or alternatively, returning the version as an object would be a nice solution, although it may seem a bit overkill.

hyperquantum:
Looks good to me.

It may be useful to also provide utility functions for converting a `VersionNumber` to a `TagLib::String` or `std::string`. But  `String` probably cannot be used in `taglib.h` because `tstring.h` already includes `taglib.h`.

ufleisch:
It is probably better to move the version number stuff into its own file. I have done this and also added a `toString()` method.

hyperquantum:
I have updated the example usage in the description of this PR.

For clarity I kept the code as similar as possible to the original, so the version numbers used (1.11.1 and 1.12) don't make much sense anymore, as the code will never be run on version 1.x.

PR #975, opened 2020-10-25
Improve speed of metadata track tag search for contemporary x86/64 CPUs

shaforostoff:
use AVX2

### DIFF
--- a/taglib/CMakeLists.txt
+++ b/taglib/CMakeLists.txt
@@ -57,6 +57,8 @@ set(tag_HDRS
   toolkit/tpropertymap.h
   toolkit/trefcounter.h
   toolkit/tdebuglistener.h
+  toolkit/fastsearch.h
+  toolkit/cpuFeaturesCheck.h
   mpeg/mpegfile.h
   mpeg/mpegproperties.h
   mpeg/mpegheader.h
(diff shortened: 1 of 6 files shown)


scotchi:
Hi Nick -- have you benchmarked this with disk caches cleared?  These sorts of things are almost always I/O bound, and this introduces a lot of complexity for something I'm skeptical is faster for real-world use cases.  (I haven't tried this since SSDs became common, so there's a chance that some things which used to be I/O-bound are now CPU bound.)

scotchi:
After trying this on my Linux box (CPU: Intel(R) Core(TM) i5-2500K CPU @ 3.30GHz) I'm not seeing any significant improvement in cached or non-cached performance for scanning a directory of 13 GB of MP3s.  I'd really need to see a use case where this significantly improves non-cached performance to consider merging it.  (There are a number of issues in the patch itself, but for the moment my primary concern is seeing this improve an important metric.)

scotchi:
Also worth noting there is that in my tests (on an SSD) it was also pretty clear that things are still I/O bound:  in non-cached test a single core only hit 42% usage and the non-cached version took 7 times longer.

shaforostoff:
> After trying this on my Linux box (CPU: Intel(R) Core(TM) i5-2500K CPU @ 3.30GHz) I'm not seeing any significant improvement
of course you won't see it, since your CPU is only 2nd gen (2500K), whereas avx instructions are available only starting with 4th gen

scotchi:
That's the sort of info that should really be included in a pull request:

- These are the CPUs this affects
- These are the use cases that it improves
- These are the benchmarks proving such

It's just random chance that I had a little time to test this out -- most of the time a patch like this wouldn't even get reviewed since there's not enough info to evaluate it on.

scotchi:
There's also no measurable performance increase in scanning 83 GB of MP3s on a brand new Macbook Pro (MacBookPro16,2 / Quad-Core Intel Core i7).  Again, for this to be given serious consideration, there needs to be a repeatable benchmark included that demonstrates improved non-cached performance.

Issue #959, opened 2020-06-01
TagLib cannot get AudioProperties with some wav files.


BLumia:
I use TagLib to get audio properties like sample rate and bitrate, it works fine with most of my audio files but some wav files doesn't works.

## To Reproduce:

 1. Download the example code [examples/tagreader.cpp](https://github.com/taglib/taglib/blob/master/examples/tagreader.cpp)
 2. Make sure taglib is installed, then build with ```g++ `pkg-config taglib --cflags --libs` tagreader.cpp```.
 3. Download [test.zip](https://github.com/taglib/taglib/files/4709696/test.zip) and unzip it, and pass the wav file path to the built `a.out` executable.
 4. See the result.

```
...
-- TAG (properties) --
-- AUDIO --
bitrate     - 0
sample rate - 0
channels    - 0
length      - 0:00
```

TagLib Version: 1.11.1-4 under Archlinux. ( I also did test with the master branch under windows, build it manually and it also doesn't works ).

Let me know if you need other details. Thanks.

ufleisch:
When you open the attached file with a debug build of TagLib, you will see the following output:

```
TagLib: RIFF::WAV::Properties::read() - Non-PCM format, but 'fact' chunk not found.
```

The `wFormatTag` in the `fmt ` chunk is 3, which means `WAVE_FORMAT_IEEE_FLOAT`. "New Multimedia Data Types and Data Techniques" (Microsoft, 1994, RIFFNEW.pdf, see also http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/WAVE.html) states

> The fact chunk is required for all new WAVE formats. The chunk is not required for the standard WAVE_FORMAT_PCM files.

As a `fact` chunk is missing in the attached file, it seems that this file is not valid. Editing the tags is still possible. Maybe it would be possible to get the properties correctly in this case without the information in the fact chunk (which is actually only the number of samples).

scotchi:
Just for a little context on why TagLib is typically pretty strict on corrupted or non-conformant files:

Very often when files are non-conformant there's garbage somewhere in the file.  Historically, a lot of crashes and hangs (i.e. parsing the whole file) came from trying to make sense of bogus files, and one of the easiest ways to avoid that was to bail out as soon as a problem was detected.

BLumia:
> ...on corrupted or non-conformant files:...

[...]

PR #980, opened 2020-12-06
Calculate bitrate for IEEE Float WAV files without fact chunk (#959)

ufleisch:
When a WAV file with float format without a `fact` chunk containing a `totalSamples` value is encountered, the bitrate is not calculated. However, since all samples have the same number of bits, the number of samples can be calculated from the size of the `data` chunk and number of channels and bits per sample, as it is done in the PCM case.


sbooth:
This makes sense to me!

Issue #961, opened 2020-06-18
MP4 File's bitrate return 0


TheRealNox:
When a M4A file is written using refalac64, taglib always return a bitrate of 0.
Here is the MediaInfo of the file
```
General
Complete name                            : D:\Music\Electro\Progressive House\Deadmau5\2010 - 4x4=12\01 - Some Chords.m4a
Format                                   : MPEG-4
Format profile                           : Apple audio with iTunes info
Codec ID                                 : M4A  (M4A /mp42/isom)
File size                                : 51.4 MiB
Duration                                 : 7 min 27 s
Overall bit rate mode                    : Variable
Overall bit rate                         : 964 kb/s
Album                                    : 4x4=12
Track name                               : Some Chords
Track name/Position                      : 1
Track name/Total                         : 12
Performer                                : Deadmau5
Genre                                    : Progressive House
Recorded date                            : 2010
Encoded date                             : UTC 2018-11-15 11:49:54
Tagged date                              : UTC 2018-11-15 11:50:02
Writing application                      : refalac 1.64, Apple Lossless Encoder
Cover                                    : Yes
VENDORID                                 : reference libFLAC 1.3.0 20130526

Audio
ID                                       : 1
Format                                   : ALAC
Codec ID                                 : alac
Codec ID/Info                            : Apple Lossless Audio Codec
Duration                                 : 7 min 27 s
Duration_LastFrame                       : -59 ms
Bit rate mode                            : Variable
Bit rate                                 : 960 kb/s
Channel(s)                               : 2 channels
Sampling rate                            : 44.1 kHz
Bit depth                                : 16 bits
Stream size                              : 51.2 MiB (100%)
Encoded date                             : UTC 2018-11-15 11:49:54
Tagged date                              : UTC 2018-11-15 11:50:02
```

If I convert it back using ffmpeg, it's all fine again.
Here is the new MediaInfo when converted with ffmpeg:
```
General
Complete name                            : C:\Users\Nox\Desktop\2010 - 4x4=12\01 - Some Chords.m4a
Format                                   : MPEG-4
Format profile                           : Apple audio with iTunes info
Codec ID                                 : M4A  (M4A /isom/iso2)
File size                                : 52.7 MiB
Duration                                 : 7 min 27 s
Overall bit rate mode                    : Variable
Overall bit rate                         : 989 kb/s
Album                                    : 4x4=12
Track name                               : Some Chords
Track name/Position                      : 1
Track name/Total                         : 12
[...]

PR #981, opened 2020-12-08
Calculate bitrate for ALAC files without it (#961)

ufleisch:
If the "alac" atom of an MP4 file has a zero bitrate, it is calculated.


scotchi:
Again, on patches covering new territory, it's always good to have tests.

Issue #954, opened 2020-04-10
Saving artwork in FLAC file prevents Windows 10 from displaying the tags


CyberSinh:
Hi,

Saving FLAC files with Taglib covers prevents Windows 10 from displaying the tags in the file explorer, if the image size is a bit heavy.

I have the impression that Windows stops looking for the tags if it doesn't find them at the very beginning of the files. However, it seems that Taglib saves the PICTURE block before the VORBIS_COMMENT block. And if the PICTURE block is too long, the VORBIS_COMMENT block is not found by Windows.

Is it possible to save the VORBIS_COMMENT block just after the STREAMINFO block, before the PICTURE block in FLAC files?

Thank you.

ufleisch:
I can confirm this on Windows 10. For small pictures, it is not a problem, but a size of 1 MB is enough to prevent Windows from displaying the tags. We probably all agree that this should be fixed on Windows, but nevertheless I suggest that TagLib writes the Vorbis Comment block before the picture block in order to improve compatibility. I have made a pull request doing this.

ufleisch:
Fixed with #983.

PR #983, opened 2020-12-12
FLAC: Store comment block before picture block (#954)

ufleisch:
When the picture block is large and comes before the comment block, Windows will no longer display the tags. This can be fixed by storing the comment block before the picture block instead of appending it at the end.


scotchi:
It'd be good to add some tests for this

PR #996, opened 2021-02-16
Use CTest, GNUInstallDirs, install examples, some cleanups

a17r:
- BUILD_TESTING is default enabled, which is a good default anyway. BUILD_TESTS is obsolete.
- The need for BUILD_SHARED_LIBS=OFF for testing is not clear, it works on Linux. But I kept it in the instructions for now.
- Move finding deps to root CMakeLists.txt: Small line decrease, but also easier to read what is happening. Now all dependencies can be read from the root CMakeLists.txt file.
- Use GNUInstallDirs: Well-established CMake standard for installation directories.
- ENABLE_STATIC error has been there since 2014, that is long enough.
- Install examples
- Move taglib.png location into html subdir


ufleisch:
Thanks for your contribution. The only problem I see is that the tests are not running on Windows with `BUILD_SHARED_LIBS=ON`. There are some symbols which are used by the unit tests but not exported. With MSVC there is another problem because multiple `Map<String, StringList>` templates are instantiated. I have added another commit to fix this. The binary compatibility on Linux is not affected by this, which is a bit strange, as well as the fact that the tests on Linux are working with shared libraries even if `VISIBILITY_HIDDEN=ON` is set. The reason why it works on Linux is that the option `VISIBILITY_HIDDEN=ON` does not really work because in _taglib/CMakeLists.txt_

```
if(VISIBILITY_HIDDEN)
  set_target_properties(tag PROPERTIES C_VISIBILITY_PRESET hidden)
endif()
```

should have `CXX_VISIBILITY_PRESET` instead of `C_VISIBILITY_PRESET`. If this is fixed, the tests on Linux will not run with shared libraries (without my additional commit). With `CXX_VISIBILITY_PRESET` fixed and `VISIBILITY_HIDDEN=ON`, `abi-compliance-checker` will report binary compatibility problems, because of 14 removed symbols (when run through `c++filt` there are duplicates, probably because of multiple template instantiations).

    TagLib::List<TagLib::ByteVector>::~List()
    TagLib::List<TagLib::ID3v2::Frame*>::~List()
    TagLib::List<TagLib::MP4::Atom*>::~List()
    TagLib::List<TagLib::String>::~List()
    TagLib::Map<TagLib::String, TagLib::StringList>::~Map()

I made a quick test by replacing the existing 1.11.1 library with a fixed one, and could not see any problems. The question is, if the thing with the hidden visibility should be fixed despite the negative `abi-compliance-checker` result.

taglib-config.cmake
```diff
@@ -16,8 +16,8 @@ EOH
 
 prefix=@CMAKE_INSTALL_PREFIX@
 exec_prefix=@CMAKE_INSTALL_PREFIX@
-libdir=@LIB_INSTALL_DIR@
-includedir=@INCLUDE_INSTALL_DIR@
+libdir=@CMAKE_INSTALL_FULL_LIBDIR@
+includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
```
joerg-krause, line 20: To fix cross-compilation issues, this shoud be:

```
libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
```

ufleisch: If I correctly understand https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html, the only difference introduced by your variant would be to prevent some _GNUInstallDirs_ automatism from prepending _/usr_ to the library and include directories when `CMAKE_INSTALL_PREFIX=/`. But I would expect that you would rather set `CMAKE_INSTALL_PREFIX=/some/crossprefix`, which should work. Or could you tell me, what you are setting as `CMAKE_INSTALL_PREFIX`, what you are expecting to get from `taglib-config` and what you are getting from `taglib-config` with the current state of this PR.

a17r: Right. Also, `CMAKE_INSTALL_<dir>` may be relative or absolute value, both is allowed, so it must not be used to assemble paths like that.

joerg-krause: To produce relocatable packages all related files bust be installed relative to a single point, referred as the "prefix". Therefore, build systems which produce relocatable packages use `CMAKE_INSTALL_PREFIX="/usr"` when building the package and set the `prefix` afterwards to a path, referred as the "pcfiledir", which is determined at runtime of the config script file, e.g. `prefix=$(dirname $(readlink -e $0))/../../usr`. The same applies to "exec_prefix".

That's why libdir and includedir should use "exec_prefix" / "prefix" to make the package fully relocatable.

Unfortunately, `CMAKE_INSTALL_<dir>` can be overridden to some path. By default, CMake sets to value to `lib`, `lib64` or some Debian-related value.

ufleisch: Something I do not understand: With your changes applied, the resulting `taglib-config` file will contain shell script code like this (with a default `CMAKE_INSTALL_PREFIX` of `/usr/local`):

```
prefix=/usr/local
exec_prefix=/usr/local
libdir=${exec_prefix}/lib
includedir=${prefix}/include
```

If you now call `taglib-config` with `prefix` and `exec_prefix`environment variables set, this will not have an effect, these variables are assigned in the script, and this is a shell script and not a Makefile where you can override variables with environment variables.

If you want externally set `prefix` and `exec_prefix`environment variables to have and effect, you would have to handle this explicitly in the script, for example using

```
-prefix=@CMAKE_INSTALL_PREFIX@
-exec_prefix=@CMAKE_INSTALL_PREFIX@
-libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
-includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
+if test -z ${prefix+x}; then
+  prefix=@CMAKE_INSTALL_PREFIX@
+  includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
+else
+  includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
+fi
+if test -z ${exec_prefix+x}; then
+  exec_prefix=@CMAKE_INSTALL_PREFIX@
+  libdir=@CMAKE_INSTALL_FULL_LIBDIR@
+else
+  libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
+fi
 ```

Comment (a17r) [reply taglib-config.cmake:20]: > To produce relocatable packages all related files bust be installed relative to a single point, referred as the "prefix". 

As a distribution person I am probably the wrong one to pitch this relocatable package thing. We build it for a specific target in a sandbox.

> Unfortunately, `CMAKE_INSTALL_<dir>` can be overridden to some path. By default, CMake sets to value to `lib`, `lib64` or some Debian-related value.

Being an overrideable standard knob is a feature that every distribution appreciates in a CMake package.

Comment (ufleisch) [reply taglib-config.cmake:20]: @a17r Sorry if I pull your PR too much into the cross compilation direction, but I want to avoid conflicts from changes for different purposes. My last comment should be answered by @joerg-krause.

Comment (joerg-krause) [reply taglib-config.cmake:20]: Sorry for pushing it into the cross-compilation and relocation direction. The issue with building packages on one machine and using them on another machine is for example described here: https://docs.conan.io/en/1.14/integrations/pkg_config_pc_files.html. The issue is not Conan-specific, e.g. I use Buildroot, but an issue for all build and/or package systems which allow to use the packages build an a specific machine to be used on another machine.

The approach Buildroot does is to patch the *-config files after installation of the package by replace the variables `prefix` and `exec_prefix` with a "pcfiledir", e.g.:

```sh
prefix=/usr
exec_prefix=/usr
libdir=${exec_prefix}/lib
includedir=${prefix}/include
```

using (simplified)

```
sed -i -e "s,^\(exec_\)\?prefix=.*,\1prefix=@STAGING_DIR@/usr,g" \
-e 's,@STAGING_DIR@,$$(dirname $$(readlink -e $$0))/../..,g'
```
becomes:

```sh
prefix=$(dirname $(readlink -e $0))/../../usr
exec_prefix=$(dirname $(readlink -e $0))/../../usr
libdir=${exec_prefix}/lib
includedir=${prefix}/include
```

This only works if `libdir` and `includedir` use `exec_prefix` or `prefix` as stated in the pkg-config: https://people.freedesktop.org/~dbn/pkg-config-guide.html. Following this approach only requires build and package systems to override the origin point `*prefix` instead of dealing with every possible variable defined in the config file.

joerg-krause: I admit, that my proposed solution does not work if `CMAKE_INSTALL_LIBDIR` is overridden. There is no easy solution to fix the issue. The issue has been discussed in zstd, see https://github.com/facebook/zstd/issues/1794 and all the follow-up patches which try to fix the creation of the pkg-config file.

ufleisch:
What about the following change to make taglib-config behave unchanged in the "distribution" use case and give Buildroot the possibility to make its replacements in the "cross-compilation" use case?

```
diff --git a/taglib-config.cmake b/taglib-config.cmake
index 9f251178..ec0834ab 100644
--- a/taglib-config.cmake
+++ b/taglib-config.cmake
@@ -14,10 +14,20 @@ EOH
        exit 1;
 }
 
-prefix=@CMAKE_INSTALL_PREFIX@
-exec_prefix=@CMAKE_INSTALL_PREFIX@
-libdir=@CMAKE_INSTALL_FULL_LIBDIR@
-includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
+# Looks useless as it is, but could be replaced with a "pcfiledir" by Buildroot.
+prefix=
+exec_prefix=
+
+if test -z "$prefix"; then
+  includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
+else
+  includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
+fi
+if test -z "$exec_prefix"; then
+  libdir=@CMAKE_INSTALL_FULL_LIBDIR@
+else
+  libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
+fi
 
 flags=""
 
@@ -38,7 +48,7 @@ do
          echo @TAGLIB_LIB_VERSION_STRING@
          ;;
     --prefix)
-         echo $prefix
+         echo ${prefix:-@CMAKE_INSTALL_PREFIX@}
          ;;
        *)
          echo "$0: unknown option $1"
```

joerg-krause:
@ufleisch Sorry for not replying earlier. I've tested your change and it works in Buildroot. Thanks!

PR #998, opened 2021-02-17
Fix taglib-config for cross compiling, again

joerg-krause:
Commit 7470f92a67375d00e53b3785a88fa7b26ad6f1da fixed using taglib-config in a cross compiling context. Commit cd9e6b750206417f155574c78d2551242a779a97 broke using `taglib-config --libs` when cross compiling as the sysroot path is not included in `LIB_INSTALL_DIR`.

Use CMakes `GNUInstallDirs` and its variables `CMAKE_INSTALL_<dir>` to get the directory name `lib` or `lib64` from CMake.

This should hopefully fixe **cross compilation and lib64** issues.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION 3.0.0 FATAL_ERROR)
 
 project(taglib)
 
+include(GNUInstallDirs)
+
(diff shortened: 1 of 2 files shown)


ufleisch:
Was integrated into #996.

PR #1006, opened 2021-03-23
Add CMake option for building with zlib

JuliusBrueggemann:
Since zlib is an optional dependency, it should be possible to build taglib without it, even if zlib is installed. Currently, this is not possible. This PR aims to change that by adding a CMake option `USE_ZLIB`.


ufleisch:
Please squash your two commits. I personally think that `WITH_` is more commonly used than `USE_`, so I would have `WITH_ZLIB` instead of `USE_ZLIB` (but maybe this is a leftover from the good? old autotools days). It would also be good to see some documentation about this option in INSTALL.md.

JuliusBrueggemann:
I don't have any preference regarding `USE_` vs `WITH_`, so I changed that to `WITH_ZLIB`.

I also squashed my commits and added a quick note about the option in the INSTALL.md

ufleisch:
Looks good, I just replaced "wheter" by "whether".

JuliusBrueggemann:
Oh, how embarrassing 😅 . Thank you.

PR #1009, opened 2021-04-25
Fuzzer fixes

bobsayshilol:
While fuzzing another project I started getting notifications from KDE that `kdeinit5` was crashing trying to generate thumbnails for the invalid FLAC files that were being produced. That piqued my curiosity and so I made a rudimentary fuzzer for taglib which discovered the following issues.

`kdeinit5` doesn't appear to use taglib for all file types, so opening a folder containing the interesting inputs only produces the following crash logs as well as causing 1 core to pin at 100% inside taglib:

```
#4  0x00007f8b080959d5 in raise () from /lib64/libc.so.6
#5  0x00007f8b0807e8a4 in abort () from /lib64/libc.so.6
#6  0x00007f8ae8274bd8 in std::__replacement_assert(char const*, int, char const*, char const*) () from /lib64/libtag.so.1
#7  0x00007f8ae8280beb in TagLib::ByteVector::operator[](int) const () from /lib64/libtag.so.1
#8  0x00007f8ae8279fb3 in TagLib::MPC::Properties::readSV7(TagLib::ByteVector const&, long) () from /lib64/libtag.so.1
#9  0x00007f8ae827a4a6 in TagLib::MPC::Properties::Properties(TagLib::MPC::File*, long, TagLib::AudioProperties::ReadStyle) () from /lib64/libtag.so.1
#10 0x00007f8ae827d1ae in TagLib::MPC::File::read(bool) () from /lib64/libtag.so.1
#11 0x00007f8ae827d37b in TagLib::MPC::File::File(char const*, bool, TagLib::AudioProperties::ReadStyle) () from /lib64/libtag.so.1
```
```
#4  0x00007f8b080959d5 in raise () from /lib64/libc.so.6
#5  0x00007f8b0807e8a4 in abort () from /lib64/libc.so.6
#6  0x00007f8ae8274bd8 in std::__replacement_assert(char const*, int, char const*, char const*) () from /lib64/libtag.so.1
#7  0x00007f8ae8280beb in TagLib::ByteVector::operator[](int) const () from /lib64/libtag.so.1
#8  0x00007f8ae828a574 in TagLib::FLAC::File::scan() () from /lib64/libtag.so.1
#9  0x00007f8ae828a8ec in TagLib::FLAC::File::read(bool) () from /lib64/libtag.so.1
#10 0x00007f8ae828ab7c in TagLib::FLAC::File::File(char const*, bool, TagLib::AudioProperties::ReadStyle) () from /lib64/libtag.so.1
```
```
#4  0x00007f8b080959d5 in raise () from /lib64/libc.so.6
#5  0x00007f8b0807e8a4 in abort () from /lib64/libc.so.6
#6  0x00007f8ae8274bd8 in std::__replacement_assert(char const*, int, char const*, char const*) () from /lib64/libtag.so.1
#7  0x00007f8ae8280beb in TagLib::ByteVector::operator[](int) const () from /lib64/libtag.so.1
#8  0x00007f8ae8266e3f in TagLib::ID3v2::OwnershipFrame::parseFields(TagLib::ByteVector const&) () from /lib64/libtag.so.1
#9  0x00007f8ae82670ae in TagLib::ID3v2::OwnershipFrame::OwnershipFrame(TagLib::ByteVector const&, TagLib::ID3v2::Frame::Header*) () from /lib64/libtag.so.1
#10 0x00007f8ae825ac6d in TagLib::ID3v2::FrameFactory::createFrame(TagLib::ByteVector const&, TagLib::ID3v2::Header const*) const () from /lib64/libtag.so.1
#11 0x00007f8ae825e483 in TagLib::ID3v2::Tag::parse(TagLib::ByteVector const&) () from /lib64/libtag.so.1
#12 0x00007f8ae825e831 in TagLib::ID3v2::Tag::read() () from /lib64/libtag.so.1
#13 0x00007f8ae825e984 in TagLib::ID3v2::Tag::Tag(TagLib::File*, long, TagLib::ID3v2::FrameFactory const*) () from /lib64/libtag.so.1
#14 0x00007f8ae825ea22 in TagLib::MPEG::File::read(bool) () from /lib64/libtag.so.1
#15 0x00007f8ae825ec6c in TagLib::MPEG::File::File(char const*, bool, TagLib::AudioProperties::ReadStyle) () from /lib64/libtag.so.1
```

Note that there's also a few UBSan issues that fuzzing picked up which I haven't fixed (mostly invalid enum values, but also unspecified `float` to `int` conversions for values that don't fit inside `int`s), however I would strongly recommend setting up something like [oss-fuzz integration](https://github.com/google/oss-fuzz) to catch any remaining issues.

### DIFF
--- a/taglib/ape/apetag.cpp
+++ b/taglib/ape/apetag.cpp
@@ -419,9 +419,9 @@ void APE::Tag::parse(const ByteVector &data)
     }
 
     const unsigned int keyLength = nullPos - pos - 8;
-    const unsigned int valLegnth = data.toUInt(pos, false);
+    const unsigned int valLength = data.toUInt(pos, false);
 
-    if(valLegnth >= data.size() || pos > data.size() - valLegnth) {
+    if(valLength >= data.size() || pos > data.size() - valLength) {
       debug("APE::Tag::parse() - Invalid val length. Stopped parsing.");
       return;
     }
@@ -439,6 +439,6 @@ void APE::Tag::parse(const ByteVector &data)
       debug("APE::Tag::parse() - Skipped an item due to an invalid key.");
     }
 
-    pos += keyLength + valLegnth + 9;
+    pos += keyLength + valLength + 9;
   }
 }
--- a/taglib/ape/apetag.cpp
+++ b/taglib/ape/apetag.cpp
@@ -421,6 +421,11 @@ void APE::Tag::parse(const ByteVector &data)
     const unsigned int keyLength = nullPos - pos - 8;
     const unsigned int valLegnth = data.toUInt(pos, false);
 
+    if(valLegnth >= data.size() || pos > data.size() - valLegnth) {
+      debug("APE::Tag::parse() - Invalid val length. Stopped parsing.");
+      return;
+    }
+
     if(keyLength >= MinKeyLength
       && keyLength <= MaxKeyLength
       && isKeyValid(data.mid(pos + 8, keyLength)))
--- a/taglib/mpeg/id3v2/frames/ownershipframe.cpp
+++ b/taglib/mpeg/id3v2/frames/ownershipframe.cpp
@@ -118,7 +118,7 @@ void OwnershipFrame::parseFields(const ByteVector &data)
   int pos = 0;
 
   // Need at least 1 byte for the encoding
-  if(data.size() - pos < 1) {
+  if(data.isEmpty()) {
     return;
   }
 
--- a/taglib/mpeg/id3v2/frames/ownershipframe.cpp
+++ b/taglib/mpeg/id3v2/frames/ownershipframe.cpp
@@ -117,6 +117,11 @@ void OwnershipFrame::parseFields(const ByteVector &data)
 {
   int pos = 0;
 
+  // Need at least 1 byte for the encoding
+  if(data.size() - pos < 1) {
+    return;
+  }
+
   // Get the text encoding
   d->textEncoding = String::Type(data[0]);
   pos += 1;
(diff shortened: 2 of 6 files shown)


taglib/mpeg/id3v2/frames/ownershipframe.cpp
```diff
@@ -117,6 +117,11 @@ void OwnershipFrame::parseFields(const ByteVector &data)
 {
   int pos = 0;
 
+  // Need at least 1 byte for the encoding
+  if(data.size() - pos < 1) {
```
ufleisch, line 121: `pos` is always `0` here, so I would prefer `if(data.isEmpty()) {`

taglib/ape/apetag.cpp
```diff
@@ -421,6 +421,11 @@ void APE::Tag::parse(const ByteVector &data)
     const unsigned int keyLength = nullPos - pos - 8;
     const unsigned int valLegnth = data.toUInt(pos, false);
 
+    if(valLegnth >= data.size() || pos > data.size() - valLegnth) {
```
ufleisch: The first term is equivalent to `1 > data.size() - valLegnth` and always true if the second term is true because `pos` starts at `0`.

taglib/mpeg/id3v2/frames/ownershipframe.cpp
bobsayshilol: Ah, yeah. I was following the convention of other uses of `pos` below, like the `pos += 1;` that could be `pos = 1;` right after the encoding read.

taglib/ape/apetag.cpp
bobsayshilol: > The first term is equivalent to `1 > data.size() - valLegnth`

I don't think this is correct, as mentioned briefly in 3d97905. Consider the case where `data.size() = 100` and `valLegnth = 101`. It's obvious that the original condition `valLegnth >= data.size()` evaluates to true, however `data.size() - valLegnth = 100 - 101 = UINT_MAX` due to both variables being unsigned types which means that `1 > data.size() - valLegnth` evaluates to false. This is why `valLegnth` must first be validated against `data.size()` before it can be subtracted from it.

PR #1012, opened 2021-05-08
Fix "error: duplicate ‘volatile’" on systems without HAVE_GCC_ATOMIC in 1.12

olegantonyan:
If you don't have `HAVE_GCC_ATOMIC` defined (for some reason) you'll get an error:
```
../../../libs/taglib/taglib-1.12/taglib/toolkit/trefcounter.cpp:55:21: error: duplicate ‘volatile’
   55 | # define ATOMIC_INT volatile int
      |                     ^~~~~~~~
../../../libs/taglib/taglib-1.12/taglib/toolkit/trefcounter.cpp:69:14: note: in expansion of macro ‘ATOMIC_INT’
   69 |     volatile ATOMIC_INT refCount;
   ```

### DIFF
--- a/taglib/toolkit/trefcounter.cpp
+++ b/taglib/toolkit/trefcounter.cpp
@@ -66,7 +66,7 @@ namespace TagLib
     RefCounterPrivate() :
       refCount(1) {}
 
-    volatile ATOMIC_INT refCount;
+    ATOMIC_INT refCount;
   };
 
   RefCounter::RefCounter() :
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: taglib/toolkit/trefcounter.cpp [hunk 1]
```diff
@@ -66,7 +66,7 @@ namespace TagLib
     RefCounterPrivate() :
       refCount(1) {}
 
-    volatile ATOMIC_INT refCount;
+    ATOMIC_INT refCount;
```
ufleisch, line 69: But this would remove `volatile` for all the other cases (`HAVE_GCC_ATOMIC, HAVE_WIN_ATOMIC, HAVE_MAC_ATOMIC, HAVE_IA64_ATOMIC`) too, which is not intended. Wouldn't it be better to remove `volatile` only in the `else` part, i.e. in [line 55 of trefcounter.cpp](https://github.com/taglib/taglib/blob/4971f8fb032f0cde8c2437cea44b6217b0f37269/taglib/toolkit/trefcounter.cpp#L55)?

olegantonyan: Yep. Fixed

PR #1018, opened 2021-06-17
clang-tidy fixes

neheb:
All are compatible with C++98.


ufleisch:
Thanks for your contributions.

PR #1019, opened 2021-06-17
use for_each instead of a raw loop

neheb:
Shorter.

### DIFF
--- a/bindings/c/tag_c.cpp
+++ b/bindings/c/tag_c.cpp
@@ -258,8 +258,7 @@ void taglib_tag_free_strings()
   if(!stringManagementEnabled)
     return;
 
-  for(List<char *>::ConstIterator it = strings.begin(); it != strings.end(); ++it)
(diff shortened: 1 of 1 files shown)


bindings/c/tag_c.cpp
```diff
@@ -258,8 +258,7 @@ void taglib_tag_free_strings()
   if(!stringManagementEnabled)
     return;
 
-  for(List<char *>::ConstIterator it = strings.begin(); it != strings.end(); ++it)
-    free(*it);
+  std::for_each(strings.begin(), strings.end(), ::free);
```
ufleisch, line 261: I am not sure if this is really better, because it uses the non-const `begin()`, which will `detach()` the `String` whereas the  loop using an iterator explicitly calls the const `begin()` thereby avoiding detaching the container.

neheb: OK.

PR #1020, opened 2021-06-17
use std::pair instead of 2D C array

neheb:
Mostly the same. Except there's now an association between both values.

Constructor had to be used since there's no uniform initialization.

This also results in a smaller so file. I have absolutely no idea why.


neheb:
replaced with make_pair. Shorter.

ufleisch:
This does not compile, you missed some parentheses after `make_pair`, e.g.

```
taglib/taglib/mpeg/id3v2/id3v2framefactory.cpp:398:19: error: expected ‘}’ before string constant
  398 |     std::make_pair"BUF", "RBUF"),
      |                   ^~~~~
```

neheb:
> This does not compile, you missed some parentheses after `make_pair`, e.g.
> 
> ```
> taglib/taglib/mpeg/id3v2/id3v2framefactory.cpp:398:19: error: expected ‘}’ before string constant
>   398 |     std::make_pair"BUF", "RBUF"),
>       |                   ^~~~~
> ```

Fixed.

scotchi:
Merging this apparently broke the build for clang-13 (guessing at least, since the error messages there include `make_pair`), see: https://github.com/taglib/taglib/commit/17213546276f91b6bbb3c23df1ddc17d1a20ab58

neheb:
What a goose chase...

This PR has nothing to do with it. The make_pair in your error logs have to do with std::map, not the ones here. Even tag 1.12 fails).

The actual issue can be demonstrated with the following patch:

```
--- a/taglib/toolkit/tmap.tcc
+++ b/taglib/toolkit/tmap.tcc
@@ -36,6 +36,7 @@ template <class Key, class T>
 template <class KeyP, class TP>
 class Map<Key, T>::MapPrivate : public RefCounterOld
 {
+static_assert(std::is_constructible<KeyP, TP&>::value, "TP must be copy constructable");
 public:
   MapPrivate() : RefCounterOld() {}
 #ifdef WANT_CLASS_INSTANTIATION_OF_MAP
```

libc++ demands a copyable type for std::map.

PR #1021, opened 2021-07-04
clang-tidy: replace static_cast with dynamic_cast

neheb:
Found with cppcoreguidelines-pro-type-static-cast-downcast

### DIFF
--- a/taglib/mpeg/id3v2/id3v2framefactory.cpp
+++ b/taglib/mpeg/id3v2/id3v2framefactory.cpp
@@ -340,19 +340,20 @@ void FrameFactory::rebuildAggregateFrames(ID3v2::Tag *tag) const
   {
     TextIdentificationFrame *tdrc =
       dynamic_cast<TextIdentificationFrame *>(tag->frameList("TDRC").front());
-    UnknownFrame *tdat = static_cast<UnknownFrame *>(tag->frameList("TDAT").front());
+    UnknownFrame *tdat = dynamic_cast<UnknownFrame *>(tag->frameList("TDAT").front());
 
     if(tdrc &&
        tdrc->fieldList().size() == 1 &&
        tdrc->fieldList().front().size() == 4 &&
+       tdat &&
        tdat->data().size() >= 5)
     {
       String date(tdat->data().mid(1), String::Type(tdat->data()[0]));
       if(date.length() == 4) {
         tdrc->setText(tdrc->toString() + '-' + date.substr(2, 2) + '-' + date.substr(0, 2));
         if(tag->frameList("TIME").size() == 1) {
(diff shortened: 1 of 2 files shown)


taglib/mpeg/id3v2/id3v2framefactory.cpp
```diff
@@ -340,7 +340,7 @@ void FrameFactory::rebuildAggregateFrames(ID3v2::Tag *tag) const
   {
     TextIdentificationFrame *tdrc =
       dynamic_cast<TextIdentificationFrame *>(tag->frameList("TDRC").front());
-    UnknownFrame *tdat = static_cast<UnknownFrame *>(tag->frameList("TDAT").front());
+    UnknownFrame *tdat = dynamic_cast<UnknownFrame *>(tag->frameList("TDAT").front());
```
ufleisch, line 343: With the `dynamic_cast` `tdat` might become `nullptr`, therefore it should be checked. I would change the next line to `if(tdrc && tdat &&`.

taglib/mpeg/id3v2/id3v2framefactory.cpp
```diff
@@ -351,7 +351,7 @@ void FrameFactory::rebuildAggregateFrames(ID3v2::Tag *tag) const
       if(date.length() == 4) {
         tdrc->setText(tdrc->toString() + '-' + date.substr(2, 2) + '-' + date.substr(0, 2));
         if(tag->frameList("TIME").size() == 1) {
-          UnknownFrame *timeframe = static_cast<UnknownFrame *>(tag->frameList("TIME").front());
+          UnknownFrame *timeframe = dynamic_cast<UnknownFrame *>(tag->frameList("TIME").front());
```
ufleisch, line 354: `timeframe` could be `nullptr` and should be checked in the next line with `if(timeframe && timeframe->data().size() >= 5) {`.

taglib/mpeg/id3v2/id3v2tag.cpp
```diff
@@ -197,7 +197,7 @@ String ID3v2::Tag::genre() const
   // appended to the genre string.  Multiple fields will be appended as the
   // string is built.
 
-  TextIdentificationFrame *f = static_cast<TextIdentificationFrame *>(
+  TextIdentificationFrame *f = dynamic_cast<TextIdentificationFrame *>(
```
ufleisch, line 200: If we touch this, it could be optimized a bit. `d->frameListMap["TCON"]` is called three times and the same `dynamic_cast` is made above.

neheb: I think this is what you meant.

ufleisch:
If have added a commit with what I meant. Is this OK?

neheb:
Sure

PR #1033, opened 2021-12-08
Add compile_commands.json to gitignore

ratijas:
Other entries were copied from KDE/marble repo as well.

See also: https://invent.kde.org/sdk/kdesrc-build/-/issues/86


ratijas:
Thanks!

PR #1035, opened 2021-12-27
enable 64-bit file and stream offsets on 32-bit architectures

Smartsmurf:
I use taglib on a 32-bit ARM powered devices. I had issues parsing MP4 files > 2 GB since the "long" datatype is 32-bits on that platform.

Attempting to compile taglib with directive -D_FILE_OFFSET_BITS=64 showed that it does not switch to 64-bit offsets.

The pull requests changes the datatypes long into off_t which will be changed by the compiler into off64_t when above directive is presented. Therefore this pull requests si free from regression on 64-bit platforms.


ufleisch:
Thanks for your contribution. However, your changes will not preserve binary compatibility and can therefore not be included in TagLib 1.x. In the taglib2 branch, changes were already made to support 64 bit file offsets. I once stumbled across the same problem with the TagLib version linked statically to the 64-bit build of [Kid3](https://kid3.kde.org/). In order to fix the problem there, I patch the TagLib sources when building for this architecture with commit [980ddd57f06a3954bf5f41f4ca4e7891b5d78e6c](https://github.com/ufleisch/taglib/commit/980ddd57f06a3954bf5f41f4ca4e7891b5d78e6c), basically a backport of the changes in the taglib2 branch. However, that commit changes a lot more files than your PR (54 instead of 8), so I assume that your PR does not handle all cases, maybe only for MP4? By the way, this problem does not exist on AMD64 Linux, since `long` has 64 bits there.

PR #1036, opened 2021-12-29
Remove MP4 meta atom when empty tag is saved

ufleisch:
Currently, MP4 tags can only grow. If items are removed, they are just replaced by padding in the form of "free" atoms. This change will remove the whole "meta" atom when an MP4 tag without items is saved. This will make it possible, to bring the file back to its pristine state without metadata.

### DIFF
--- a/taglib/mp4/mp4file.cpp
+++ b/taglib/mp4/mp4file.cpp
@@ -176,6 +176,26 @@ MP4::File::save()
 }
 
 bool
+MP4::File::strip(int tags)
+{
+  if(readOnly()) {
+    debug("MP4::File::strip() - Cannot strip tags from a read only file.");
+    return false;
+  }
+
+  if(!isValid()) {
+    debug("MP4::File::strip() -- Cannot strip tags from an invalid file.");
+    return false;
+  }
+
+  if(tags & MP4) {
+    return d->tag->strip();
+  }
+
+  return true;
+}
+
+bool
 MP4::File::hasMP4Tag() const
 {
   return (d->atoms->find("moov", "udta", "meta", "ilst") != 0);
--- a/taglib/mp4/mp4file.h
+++ b/taglib/mp4/mp4file.h
@@ -49,6 +49,19 @@ namespace TagLib {
     {
     public:
       /*!
+       * This set of flags is used for strip() and is suitable for
+       * being OR-ed together.
+       */
+      enum TagTypes {
+        //! Empty set.  Matches no tag types.
+        NoTags  = 0x0000,
+        //! Matches MP4 tags.
+        MP4     = 0x0001,
+        //! Matches all tag types.
+        AllTags = 0xffff
+      };
+
+      /*!
        * Constructs an MP4 file from \a file.  If \a readProperties is true the
        * file's audio properties will also be read.
        *
@@ -113,6 +126,15 @@ namespace TagLib {
        * This returns true if the save was successful.
        */
       bool save();
+
+      /*!
+       * This will strip the tags that match the OR-ed together TagTypes from the
+       * file.  By default it strips all tags.  It returns true if the tags are
+       * successfully stripped.
+       *
+       * \note This will update the file immediately.
+       */
+      bool strip(int tags = AllTags);
 
       /*!
        * Returns whether or not the file on disk actually has an MP4 tag, or the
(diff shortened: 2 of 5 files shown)


scotchi:
I don't have a strong opinion on this, and don't know the mp4 format at all, but it's worth perhaps considering that this would require rewriting the entire file, which in batch operations would be very slow (or, in video files, which may be multiple gigabytes).  For a multigig file, by default avoiding rewriting the whole thing by leaving a couple kb of padding is probably preferable.

ufleisch:
I just checked this situation in practice by removing the tags from some MP4 videos stored on a slow USB stick. In some files, the "udta" atom containing the meta atom is located at the end of the file, so only the end of the file has to be modified. The removal of the "meta" atom in such files took less than a second. But there are also videos which have a considerable amount of data in an "mdat" atom at the end. Most of the videos I checked come only with a "©too" atom containing the encoder, e.g. "Lavf58.76.100" inside "meta" and without padding in the form of a "free" atom. Therefore you will have the same situation when you add a tag for the first time. Indeed I had to wait for 20 seconds both for adding the first tag and for removal of all tags in files with a lot of data at the end. As expected, adding additional tags not exceeding the padding is fast afterwards.

Since both slow situations - adding a tag for the first time and removing all tags - are not the standard situation, I could live with that. Another option would be to use a `strip()` method as it is done for other tags. But since there is only one kind of tag in MP4 files, the `tags` flag would only have one option and the realization would be a bit different: The `strip()` method would remove all items and additionally set a boolean marker flag. The actual removal of the "meta" atom would still be performed in `save()`, but only when this "strip marker" is set.

scotchi:
Serious question:  are there any user benefits of that 20 second write, or is it just going kind of OCD-ish on things?  I often have the same tick, but at least in the case of MP3s, I made the opposite decision, and MP3s are a couple orders of magnitude smaller.  I do feel like it might make more sense when the explicit intent isn't to strip the tags to leave the tag there with no atoms and some padding rather than remove everything and potentially have to rewrite a bunch of large files.  I'm thinking of the case that someone does this not just to one file, but to a collection, and, say, several minutes of high I/O and a possibly blocked UI result for no actual user benefit.

ufleisch:
Not that I have such a diagnosis (maybe I would get it if I were a child in these times), but I strip the padding from all the tags after saving them. In my case these are MP3 files. I could also switch off the corresponding warning of MP3 Diags, if it was just to get no warnings. I also remove embedded images and store album covers as a separate JPEG file inside the album folders since I do not want to waste space with the same image in all tracks of an album. I know that the space is neglectable compared with the overall size, but I think that people who use tagging software love to keep things tidy. If noone would bother about the "wasted space" there would not be such a warning in MP3 Diags and in the case of MP4 files, AtomicParsley would not show an atom overhead percentage, nor would oggz-info show an Ogg overhead percentage.

I personally am a bit worried when I use a software which operates on my data but is not able to bring the data back to the previous state. Therefore I consider it as a requirement for every format supported by TagLib to not only be able to add meta data but also be able to clean it up and bring the files back to the untagged state. That is also why I have added tests which check if the file has the original contents after adding and then removing meta data.

I have received bug reports for Kid3, my main project which is using TagLib, from people complaining that meta data cannot be removed cleanly from MP4 files ("Kid3 removes picture and tags, but the size of the file doesn't reduce", "If I retry to remove tags of the AAC file by Kid3 that I had already removed tags by TagScanner, Kid3 changes the digest of the file", not publicly visible, received by E-mail). In such cases I tell them to use the other plugin supporting MP4 files, which uses mp4v2, but since that library is no longer maintained and does not support so many of these fancy atoms which are introduced by Apple from time to time, this is a workaround at best, and I would rather see them using TagLib for all formats. A user has also confirmed that "Taglib has substantially better performance over Mp4v2 plugin". So what I would like to introduce is at least the possibility to remove MP4 metadata with TagLib without bothering users who do not appreciate such a feature. Maybe the solution with a `strip()` method would be better, so that the current behavior would stay the same but it would be possible for software like Kid3 to remove the meta data.

scotchi:
Yeah, I didn't mean that it shouldn't be possible -- just that potentially that the semantics should be explicit -- that `strip()` should remove the tag completely, whereas just saving things would leave the padding.

I don't have a really strong opinion on this, but the common case of using TagLib isn't using a tagger like kid3 (where I imagine users are more picky and aware of the actual contents of the tags), but people just updating meta data.  I could even imagine a simple heuristic where e.g. if the file size didn't change by more than 0.1%, that we leave the padding in place.

ufleisch:
I added one more commit to implement a `strip()` method which behaves like its `MPEG::File::strip()` counterpart. The behavior of `save()` can stay as it is. If it is OK, you could squash the two commits.

Issue #1043, opened 2022-03-06
APE tag incorrect values


smithjd15:
This looks to be the same issue and is related to the id3 tag parsing: https://github.com/robertmuth/apetag/issues/22

Id3 magic can be found as part of an ape header, giving a false positive.

PR #1044, opened 2022-03-07
Utils: Fix potential ID3v1 false positive in the presence of an APE tag.

smithjd15:
This could cause incorrect tag parsing. Closes https://github.com/taglib/taglib/issues/1043.

### DIFF
--- a/taglib/tagutils.cpp
+++ b/taglib/tagutils.cpp
@@ -38,11 +38,22 @@ long Utils::findID3v1(File *file)
   if(!file->isValid())
     return -1;
 
-  file->seek(-128, File::End);
-  const long p = file->tell();
+  // Differentiate between a match of APEv2 magic and a match of ID3v1 magic.
 
-  if(file->readBlock(3) == ID3v1::Tag::fileIdentifier())
-    return p;
+  if (file->length() >= 131) {
+    file->seek(-131, File::End);
+    const long p = file->tell() + 3;
+    const TagLib::ByteVector data = file->readBlock(8);
+
+    if(data.containsAt(ID3v1::Tag::fileIdentifier(), 3) && (data != APE::Tag::fileIdentifier()))
+      return p;
+  } else {
+    file->seek(-128, File::End);
+    const long p = file->tell();
+
+    if(file->readBlock(3) == ID3v1::Tag::fileIdentifier())
+      return p;
+  }
 
   return -1;
 }
--- a/tests/test_apetag.cpp
+++ b/tests/test_apetag.cpp
@@ -29,6 +29,7 @@
 #include <tstringlist.h>
 #include <tbytevectorlist.h>
 #include <tpropertymap.h>
+#include <apefile.h>
 #include <apetag.h>
 #include <tdebug.h>
 #include <cppunit/extensions/HelperMacros.h>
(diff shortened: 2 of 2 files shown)


ufleisch:
What you are fixing here is the case where the APE tag has exactly such a length that its leading "APETAGEX" identifier has just the position so that its "TAG" part is at the position where the "TAG" identifier for an ID3v1 tag is supposed to be. I agree that this case should be fixed, however, the code is used not only for files containing an APE tag but also for all other formats which can have an ID3v1 tag. Therefore I would like the code to stay as efficient as it is today, but the new code uses two seeks and two reads in the normal case whereas the old code only uses one seek and one read. It would be possible to do it also with one seek and one read since the contents of the second read are already contained in the first read. But then, the code would no longer work if a file has less then 131 bytes, for example a file containing only an ID3v1 tag. If we still want to be able to handle such files, we could just read 5 bytes after position -128 and then check if they are "TAGEX". In such a rare case (it is really an APE tag or an ID3v1 with a title starting with "EX"), we could check if the bytes from position -131..-128 are "APE". This would be a bit more complicated, but it still use only one seek and one read in the normal case while still supporting files having less than 131 bytes.

Besides this, I would like to see a test case for such a file.

smithjd15:
Less seeks, id3v1-as-a-file fixed, and a unit test.

smithjd15:
Correct p.

taglib/tagutils.cpp
```diff
@@ -38,11 +38,22 @@ long Utils::findID3v1(File *file)
...
+
+    if(data.containsAt(ID3v1::Tag::fileIdentifier(), 3) && (data != APE::Tag::fileIdentifier()))
+      return p;
+  } else {
+    file->seek(-128, File::End);
+    const long p = file->tell();
+
+    if(file->readBlock(3) == ID3v1::Tag::fileIdentifier())
+      return p;
+  }
```
ufleisch, line 56: The inefficiency here is that `file->length()` uses two seeks. I would suggest the following code, which only uses additional system calls in the rare case of an ID3v1/APEv2 collision:

```
long Utils::findID3v1(File *file)
{
  if(!file->isValid())
    return -1;

  file->seek(-128, File::End);
  const long p = file->tell();

  const ByteVector data = file->readBlock(5);
  if(data.startsWith(ID3v1::Tag::fileIdentifier())) {
    if(data.endsWith("EX")) {
      // Differentiate between a match of APEv2 magic and a match of ID3v1 magic.
      file->seek(-131, File::End);
      if(file->readBlock(8) == APE::Tag::fileIdentifier())
        return -1;
    }
    return p;
  }

  return -1;
}
```

smithjd15: Seek and tell are nearly cost-free. In which case the current patch is preferable because it saves a read over the worst-case, and this would happen a lot with APEv2 tags. The current patch should also perform equivalently to the current code.

Quick test clearing caches between runs, `time bash -c 'for f in *; do tagreader "$f" &>/dev/null; done'` on a HDD directory of 156 files containing APEv2 tags. At least two runs, last run result:

Current code: 
`real    0m9.587s
user    0m1.460s
sys     0m0.876s`

Current patch:
`real    0m8.393s
user    0m1.368s
sys     0m0.775s`

Suggested:
`real    0m8.566s
user    0m1.250s
sys     0m0.730s`

No appreciable difference between either patch. Not sure what caused a full second penalty for the current code. All 3 runs were built with the same compiler flags.

ufleisch: OK, you are right, one should always measure rather than assume what is faster. I will merge the current state.

PR #1046, opened 2022-03-13
Improve FrameFactory subclassing merge

Max-F-Helm:
This PR changes the ID3v2-FrameFactory so it is possible to subclass it for implementing own frames.
Additionally it splits up the createFrame() method to make its usage in subclasses a bit more easy.

As this is a binary-incompatible change I do not expect you to take this PR in consideration until next major release.

### DIFF
--- a/taglib/mpeg/id3v2/id3v2framefactory.cpp
+++ b/taglib/mpeg/id3v2/id3v2framefactory.cpp
@@ -116,9 +116,9 @@ FrameFactory *FrameFactory::instance()
   return &factory;
 }
 
-Frame *FrameFactory::createFrame(const ByteVector &origData, const Header *tagHeader) const
+std::pair<Frame::Header *, bool> FrameFactory::prepareFrameHeader(
+  ByteVector &data, const Header *tagHeader) const
 {
-  ByteVector data = origData;
   unsigned int version = tagHeader->majorVersion();
   auto header = new Frame::Header(data, version);
   ByteVector frameID = header->frameID();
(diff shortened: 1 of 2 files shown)


Max-F-Helm:
Do you mean with 'coding style' how the code is formatted (indentions, position of brackets) or how the functionality is written?
If it is the first, the I will fix it (I mus admit that I should have looked closer to the existing code before committing).
If it is the latter, the I would ask you to give me some suggestions how I could improve my PR.

scotchi:
There are a lot of lines changed in your diff where you *just* changed the coding style -- you didn't actually change any of the program.  If you look at the *files changed* tab above, you'll see that most of the places are just where you're inserting spaces before parenthesis, converting spaces to tabs, etc.

Aside from the fact that submissions to TagLib need to follow TagLib's coding style, there absolutely can't be lines changed that *only* change the coding style *away* from TagLib's standard style.

taglib/mpeg/id3v2/id3v2framefactory.cpp
```diff
@@ -129,207 +129,24 @@ Frame *FrameFactory::createFrame(const ByteVector &data, unsigned int version) c
 
 Frame *FrameFactory::createFrame(const ByteVector &origData, Header *tagHeader) const
 {
-    return createFrame(origData, const_cast<const Header *>(tagHeader));
```
Max-F-Helm, line 132: this has nothing to do with my changes but it had a differing indention than the other lines, so i kept the change

Max-F-Helm:
ok, fixed it

ufleisch:
I have cleaned up your PR so that all the unnecessary reformatting and moving is eliminated and the changes are clearly visible. This reduces the changed line count from +229-195 to +50-10 and makes it reviewable again. I also tried to make the code more readable by adding comments and using C++17 structured bindings instead of reference-to-pointer. Please have a look at it.

Max-F-Helm:
Looks good.
Thanks for all the effort.

PR #1050, opened 2022-06-12
Streamline CppUnit and ZLIB search algorithms

zlaski:
This refactors module search, and uses it for CppUnit and ZLIB searches.  Neither search functioned properly before now.

This is undoubtedly an improvement over the present state of affairs, but it has not been thoroughly tested.


scotchi:
CMake now comes with a module to find ZLib, so it'd probably make more sense to use the built-in function.

PR #1071, opened 2022-11-14
cmake: generate pc files with the use of prefix

dabrain34:


ufleisch:
Thanks for your contribution. However, I feel a bit reluctant to change the build system once more. We made changes to the build system between versions 1.12 and 1.13. See the discussion for PR #996 for details, the mentioned document [pkg-config and .pc files](https://docs.conan.io/en/1.14/integrations/pkg_config_pc_files.html) from the Conan docs lists various approaches to deal with the problem of absolute paths in _.pc_ files. Probably, currently [Approach 1: Import and patch the prefix in the .pc files](https://docs.conan.io/en/1.14/integrations/pkg_config_pc_files.html#approach-1-import-and-patch-the-prefix-in-the-pc-files) is used, therefore _taglib-config.cmake_ contains the following lines

```
prefix=
exec_prefix=
```

Your patch seems to use [Approach 3: Use --define-prefix](https://docs.conan.io/en/1.14/integrations/pkg_config_pc_files.html#approach-3-use-define-prefix) or [Approach 4: Use PKG_CONFIG_$PACKAGE_$VARIABLE](https://docs.conan.io/en/1.14/integrations/pkg_config_pc_files.html#approach-4-use-pkg-config-package-variable).

In order not to annoy many packagers by forcing them to adapt their scripts with every release, I would suggest to leave it as it is.

dabrain34:
On cerbero, the packaging system for GStreamer, we have been adopting the approach 1 for taglib to patch its result see [patch](https://gitlab.freedesktop.org/gstreamer/cerbero/-/merge_requests/1017) but most of our packages are using the approach 3 , it would be better to use the `{prefix}` approach which is the common approach for most of the packages which are usually respecting lib and include folder tree. Then if a packager needs to define specific lib or include he should  use the approach 4 in the CMake command line.

neheb (approved):

ufleisch:
I assume that this change will only affect the generation of the *taglib_c.pc* and *taglib.pc* files. Would it be OK to name the CMake variables `PC_PREFIX, PC_EXEC_PREFIX, PC_LIBDIR, PC_INCLUDEDIR` instead of `prefix, exec_prefix, libdir, includedir` to make this more clear?

neheb:
Note that this PR fixes wrong paths when cross compiling: https://github.com/openwrt/packages/blob/master/libs/taglib/Makefile#L53

taglib.pc.cmake
```diff
@@ -1,7 +1,7 @@
-prefix=@CMAKE_INSTALL_PREFIX@
-exec_prefix=@CMAKE_INSTALL_PREFIX@
-libdir=@CMAKE_INSTALL_FULL_LIBDIR@
-includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
+prefix=@prefix@
+exec_prefix=@exec_prefix@
+libdir=@libdir@
+includedir=@includedir@
```
ufleisch, line 4:
```suggestion
prefix=@PC_PREFIX@
exec_prefix=@PC_EXEC_PREFIX@
libdir=@PC_LIBDIR@
includedir=@PC_INCLUDEDIR@
```

ufleisch (changes requested):

ufleisch:
I have simplified the changes and fixed the case for taglib_c. Could you please have a look at it?

dabrain34:
LGTM

trofi:
`nixpkgs` noiotice the regression in absolute `CMAKE_INSTALL_LIBDIR` path support: https://github.com/taglib/taglib/issues/1098

PR #1073, opened 2022-11-27
clang-tidy cleanups

neheb:


taglib/toolkit/tbytevectorstream.cpp
```diff
@@ -81,7 +81,7 @@ ByteVector ByteVectorStream::readBlock(unsigned long length)
 void ByteVectorStream::writeBlock(const ByteVector &data)
 {
   unsigned int size = data.size();
-  if(long(d->position + size) > length()) {
+  if((d->position + size) > length()) {
```
ufleisch, line 84: Redundant parentheses

neheb: Fixed.

PR #1074, opened 2022-12-05
clang: remove const return

neheb:
Found with readability-const-return-type

### DIFF
--- a/taglib/toolkit/tmap.h
+++ b/taglib/toolkit/tmap.h
@@ -159,7 +159,7 @@ namespace TagLib {
      * If the map does not contain \a key, it returns defaultValue.
      * If no defaultValue is specified, it returns a default-constructed value.
      */
-    const T value(const Key &key, const T &defaultValue = T()) const;
+    T value(const Key &key, const T &defaultValue = T()) const;
 
     /*!
      * Returns a reference to the value associated with \a key.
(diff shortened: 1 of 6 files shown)


neheb:
Some of these feel like they should be references.

But that's an ABI break.

scotchi:
Is changing return types in master really a good idea?

ufleisch:
According to some opinions on StackOverflow (e.g. [Purpose of returning by const value?](https://stackoverflow.com/questions/8716330/purpose-of-returning-by-const-value)) it does not really make sense to return a value (not a reference or pointer) as const. I also checked the resulting binaries with `abi-compliance-checker`, which reported only a problem of low severity "Replacement of return type may indicate a change in its semantic meaning" (binary) and "Recompilation of a client program may be broken" (source). I cannot imagine that this will break compilation.
If you think that this is a bad idea, I will revert the commit, but it would be even more helpful to hear your opinion before i merge if you find the time to do so. I usually leave MRs open for at least a week, but maybe I was too fast this time.

scotchi:
Ah, sorry, was responding from my phone:  I thought I'd remembered that some of the things changed from a reference type to a non-reference type.  Return types don't get mangled into the ABI, and constness shouldn't affect run-time behavior, so this seems ok.

neheb:
Using const here might break RVO. Not sure. Anyway, it’s standard practice to return non const, except for pointers and references.

scotchi:
Sure, there's no argument that they should return const, but it's also important to not break the ABI (which I don't think this change will).

PR #1075, opened 2022-12-10
clang-tidy: member function const

neheb:
Found with readability-make-member-function-const

### DIFF
--- a/taglib/toolkit/trefcounter.h
+++ b/taglib/toolkit/trefcounter.h
@@ -83,25 +83,25 @@ namespace TagLib
 #ifdef TAGLIB_ATOMIC_MAC
     void ref() { OSAtomicIncrement32Barrier(const_cast<int32_t*>(&refCount)); }
     bool deref() { return ! OSAtomicDecrement32Barrier(const_cast<int32_t*>(&refCount)); }
-    int32_t count() { return refCount; }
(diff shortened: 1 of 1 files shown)


scotchi:
The first part of this should definitely not be applied:  `setProperties()` is logically non-const, and is only not virtual because of ABI constraints (based on a non-const base class method).

scotchi:
As a general comment, it makes sense to actually read the patches before submitting pull requests, rather than just blindly running automated tools and not sanity checking them to see if there's a reason that the convention isn't followed.

neheb:
Removed.

ufleisch:
I would not advice to make this change. Common policies about binary compatibility (e.g. [KDE Policies/Binary Compatibility Issues With C++](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B)) state that CV qualifiers of methods shall not be changed. I see that in this case, the binary compatibility would not be affected as the class is only used as a base class for internal non exported classes, but it would set a bad example. And the fact that the class is one which will be removed anyway when we prepare the 2.0 release in the not too distant future, I do not think that it would be worth it.

ufleisch:
`RefCounterOld` is now removed, so this MR is no longer applicable.

PR #1093, opened 2023-07-22
fix deprecated implicit copy operator=

neheb:
Just inherit the functions. Nothing special is being done here.


neheb:
https://github.com/taglib/taglib/pull/1095 fixes it differently.

PR #1095, opened 2023-07-25
Use unique_ptr for d-pointers

neheb:


neheb:
fun bug I found.

ufleisch (commented):
> fun bug I found.

Which bug?

neheb:
> > fun bug I found.
> 
> Which bug?

Pointers pointing to non existent stuff. See last commit. Found with unique_ptr.

taglib/fileref.h
```diff
@@ -95,7 +95,7 @@ namespace TagLib {
       /*!
        * Destroys this FileTypeResolver instance.
        */
-      virtual ~FileTypeResolver() = 0;
+      virtual ~FileTypeResolver() = default;
```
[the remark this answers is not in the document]
neheb: Fixed.

taglib/fileref.h
```diff
@@ -115,11 +115,6 @@ namespace TagLib {
     class TAGLIB_EXPORT StreamTypeResolver : public FileTypeResolver
     {
     public:
-      /*!
-       * Destroys this StreamTypeResolver instance.
-       */
-      ~StreamTypeResolver() override = 0;
```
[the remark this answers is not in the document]
neheb: Fixed.

taglib/mpeg/id3v2/frames/textidentificationframe.cpp
```diff
@@ -335,6 +339,8 @@ PropertyMap TextIdentificationFrame::makeTMCLProperties() const
 // UserTextIdentificationFrame public members
 ////////////////////////////////////////////////////////////////////////////////
 
+UserTextIdentificationFrame::~UserTextIdentificationFrame() = default;
```
ufleisch, line 342: Why not
```
UserTextIdentificationFrame::~UserTextIdentificationFrame()
{
  delete d;
}
```

But then `d(nullptr)` should be added to `UserTextIdentificationFrame::UserTextIdentificationFrame(const ByteVector &data)`.

scotchi: I feel like these emails always catch me at a moment where I don't have time to do a deep dive, but...

Has anyone looked at how Qt handles d-pointers now?  Manual deletion or even setting a smartpointer to nullptr feels clunky in 2023.

As a style note, with C++11 and higher, I do prefer initializing variables in the header rather than in the constructor.  Thoughts?  e.g.

    class Foo
    {
      class Private;
      Private *d = nullptr;
    };

sbooth: One pattern using `unique_ptr` is the following:

*header:*
```c++
class Foo
{
public:
  Foo();
  ~Foo();
private:
  class Private;
  std::unique_ptr<Private> _private;
};
```

*implementation:*
```c++
class Foo::Private
{
  /// ...
};

Foo::Foo()
  : _private{std::make_unique<FooPrivate>()}
{}

// Destructor cannot be defaulted in header because Private is an incomplete type
Foo::~Foo() = default;
```

neheb: I have a commit that does just that. This PR is a stepping stone to that.

neheb:
I gave up. Manually adding delete d to destructors caused issues. Decided to just use unique_ptr.

neheb:
Alright. No more d(nullptr) or funny shenanigans.

neheb:
converted newly added d pointers.

taglib/mpeg/id3v2/frames/textidentificationframe.cpp
neheb: @scotchi from ChatGPT:

QScopedPointer is part of the Qt library and follows the rules of the pre-C++11 era, where move semantics were not standardized. Consequently, you cannot transfer ownership of a QScopedPointer from one instance to another using move operations like std::move().

For Qt-based code and modern C++ practices, consider using std::unique_ptr for move semantics and transferring ownership between objects. If you need to use Qt-specific memory management classes with move semantics, you can use QSharedPointer, which provides similar functionality to std::shared_ptr in standard C++. Keep in mind that QSharedPointer has the overhead of reference counting, which may or may not be desirable depending on your specific use case.

sbooth:
I converted this PR to draft since it's still changing frequently. Once everything is settled I'll be happy to review and merge it.

neheb:
Hmm? I force pushed a non conflicting version. It's fine now. My other PRs don't conflict anymore.

sbooth:
If all the proposed changes for this PR are included I'll review it.

neheb:
Yeah. Everything’s here.

sbooth:
@neheb I can't adequately review this PR while it keeps changing. Once it's stable I will continue working through it.

neheb:
fair enough. Last change was adding = delete to class with added destructor. There isn't anything left.

sbooth:
@neheb could you resolve the conflicts with this branch? I think this PR is otherwise ready for merge.

neheb:
Done.

PR #878, opened 2018-09-08
Add DSF and DSDIFF file types management

jonaski:
Resuming PR from @damien78
I have merged with master and fixed the conflicts
I also added DSF and DSDIFF to detectByExtension

PR #1096, opened 2023-07-25
Add DSD Interchange File Format (DSDIFF) support

sbooth:
DSDIFF support from #878 by @damien78 and @scotchi

Format specification at https://www.sonicstudio.com/pdf/dsd/DSDIFF_1.5_Spec.pdf


ufleisch:
@sbooth Could you have a look at my changes? I tried to bring the DSDIFF implementation a bit further by fixing the adding and removing of DIIN chunks and modernizing the code. Now the tests are green and your test file can be brought back to its original state by stripping the tags. As I do not have equipment using DSDIFF files or any real DSDIFF files, it would be good if you could make some more tests, especially concerning the following points:

- Currently, files without an ID3 chunk are considered invalid. Is this true for files "in the wild" or should adding such a chunk be implemented?
- The test file has a root level ID3 chunk, so I could not test with ID3 child chunks within a PROP chunk and neither the removal of such a duplicate ID3 chunk.
-  The test file does not come with a DIIN chunk, so I could only test with a self-generated DIIN chunk, and only with such chunks at the end of the file. It would also be good to verify if the implementation behaves correctly in the presence of existing DIIN chunks containing MARK and MANF chunks.

sbooth:
@ufleisch Thank you for working on this and fixing the tests. I am not the original author of this code and I actually thought about rewriting it based on a DFF parser I wrote for playback (https://github.com/sbooth/SFBAudioEngine/blob/main/Decoders/SFBDSDIFFDecoder.mm) because I'm not necessarily a fan of the code's structure although I'm not sure it would be worth it now that you've got things working.

I don't really use DFF files other than for testing so I don't have many files "from the wild."  I do have a few samples that I downloaded years ago that don't seem to be available any longer.  I'm happy to share them with you directly but they are quite large (130 MiB to 1.1 GiB) so committing them to the repo isn't a good option.

It's my impression that DSF is much more widespread and most DSD downloads these days seem to be DSF.

My best guesses for answers to your questions, based on primarily the spec and inspecting the files I have:

1) ID3 is definitely not mentioned in the spec and none of the files I have contain ID3 tags. I assume this is an unofficial extension so I don't think that files without an ID3v2 chunk should be considered invalid. I think supporting a local ID3 chunk in the Form DSD Chunk (`FRM8`) is reasonable, but ID3 within other local chunks defined by the spec (such as `PROP`) isn't a great idea in my opinion. Perhaps for reading only if at all?

The spec does say

```
2.4 HANDLING OF UNRECOGNIZED CHUNKS
When an application encounters a local chunk that is not recognised it must discard (skip) it while reading the Form DSD chunk. If an application copies the Form DSD Chunk without edits, it is nicer, but not necessary, to copy unrecognised chunks too. But if an application modifies the data in any way, then it must discard all unrecognised chunks: the writing application cannot guarantee that the unrecognised data is still consistent with the modified data.
```

So at least an ID3 chunk within the `FRM8` chunk should be skipped by software that doesn't support it.

2) I have no such files either.

3) None of my sample files have DIIIN chunks either, sorry.

ufleisch:
My assumption that DFF files without ID3 chunk are considered invalid was wrong. They are accepted and an ID3 chunk is correctly added and removed. I have now added a test with an ID3 chunk within a PROP chunk and some other tests so that we now have a pretty good test coverage. One thing which is not covered is the reading of the "DST " chunk. The test file does not have such a chunk, so if you have some files with such a chunk, you could run a test on them. Besides this, I think that the code would be ready to be merged.

sbooth:
None of my files have DST chunks either. I will continue to search for more files to test with.

I did find one sample file with a DIIN chunk at the end of the file. The comments were read correctly but there were no MARK or MANF present to further test the functionality.

I also think this is ready to merge and perhaps open up for wider testing.

PR #1097, opened 2023-07-28
Use C++11 atomics for RefCounter

sbooth:
Since it appears the next version of TagLib will use C++17 features it's possible to simplify `RefCounter` using C++11 atomics.


neheb:
Why not just use shared_ptr?

PR #1099, opened 2023-07-29
various cleanups

neheb:


sbooth:
This PR would be simpler to review if it was split into multiple PRs each limited to a single topic.

neheb:
Well, except for the last one these are small changes.

neheb:
Actually that namespace change to static inline was wrong. Removed.

PR #1100, opened 2023-07-30
clang-tidy: replace RefCounter with shared_ptr

neheb:
The latter is C++11.

Found with clang-analyzer-webkit.NoUncountedMemberChecker


sbooth:
I think this change makes sense, thanks.

neheb:
Rebased.

taglib/mpeg/mpegheader.h
```diff
@@ -29,6 +29,8 @@
 #include "taglib.h"
 #include "taglib_export.h"
 
+#include <memory>
+
```
sbooth, line 33: I noticed that `memory` is included here but not in most of the other headers. For consistency it should probably be included in all of them that use `shared_ptr`.

neheb: Right. I did this as it seems most cop files use includes inherited from elsewhere.

Maybe include-what-you-use should be ran against the codebase.

sbooth: That sounds like a great candidate for a PR once I get the current few merged.

neheb: You still want me to add memory everywhere in this PR?

sbooth: I think it's a good technique but not necessary for this PR. I'm not aware of a taglib style for includes.

sbooth:
The only thing left for this PR that I know if is removing the `trefcounter` sources and adjusting the CMakeFiles appropriately.

neheb:
Done.

PR #1105, opened 2023-08-04
remove pointless loop

neheb:


sbooth:
Looks good, thanks.

taglib/mpeg/id3v2/frames/textidentificationframe.cpp
```diff
@@ -364,10 +364,7 @@ String UserTextIdentificationFrame::toString() const
 {
   // first entry is the description itself, drop from values list
   StringList l = fieldList();
-  for(auto it = l.begin(); it != l.end(); ++it) {
-    l.erase(it);
-    break;
-  }
+  l.erase(l.begin());
```
hyperquantum, line 367: This change removes the check that verifies that the collection is not empty.

The equivalent of the original code without the loop would be:
```cpp
if (l.begin() != l.end())
    l.erase(l.begin());
```

Is it safe to assume that the condition is always true?

sbooth: I think you're correct, there should be a check because the single iterator version of `erase` requires that the iterator be dereferenceable (unlike the two iterator version).

neheb:
AFAIK, empty() should be used instead of comparing begin and end. But it's unimplemented. So I went with size().

sbooth:
> AFAIK, empty() should be used instead of comparing begin and end. But it's unimplemented. So I went with size().

It would be simple enough to add `empty` to tlist also.

neheb:
nvm there's isEmpty().

PR #1107, opened 2023-08-07
clang-tidy: convert to algorithms

neheb:
Found with: readability-use-anyofallof

### DIFF
--- a/taglib/ape/apetag.cpp
+++ b/taglib/ape/apetag.cpp
@@ -65,13 +65,7 @@ namespace
         return false;
     }
 
-    const String upperKey = String(key).upper();
-    for(auto k : invalidKeys) {
-      if(upperKey == k)
-        return false;
-    }
-
-    return true;
+    return std::none_of(invalidKeys.begin(), invalidKeys.end(), [upperKey = String(key).upper()](auto k) { return upperKey == k; });
   }
 }  // namespace
 
--- a/taglib/mp4/mp4atom.cpp
+++ b/taglib/mp4/mp4atom.cpp
@@ -97,19 +97,12 @@ MP4::Atom::Atom(File *file)
     if(name == c) {
       if(name == "meta") {
         offset_t posAfterMeta = file->tell();
-        ByteVector nextSize = file->readBlock(8).mid(4, 4);
         static constexpr std::array metaChildrenNames {
           "hdlr", "ilst", "mhdr", "ctry", "lang"
         };
-        bool metaIsFullAtom = true;
-        for(auto child : metaChildrenNames) {
-          if(nextSize == child) {
-            // meta is not a full atom (i.e. not followed by version, flags). It
-            // is followed by the size and type of the first child atom.
-            metaIsFullAtom = false;
-            break;
-          }
-        }
+        // meta is not a full atom (i.e. not followed by version, flags). It
+        // is followed by the size and type of the first child atom.
+        auto metaIsFullAtom = std::none_of(metaChildrenNames.begin(), metaChildrenNames.end(), [nextSize = file->readBlock(8).mid(4, 4)](const auto &child) { return nextSize == child; });
         // Only skip next four bytes, which contain version and flags, if meta
         // is a full atom.
         file->seek(posAfterMeta + (metaIsFullAtom ? 4 : 0));
--- a/taglib/mpeg/id3v2/id3v2framefactory.cpp
+++ b/taglib/mpeg/id3v2/id3v2framefactory.cpp
@@ -147,11 +147,9 @@ std::pair<Frame::Header *, bool> FrameFactory::prepareFrameHeader(
   }
 #endif
 
-  for(auto it = frameID.cbegin(); it != frameID.cend(); it++) {
-    if( (*it < 'A' || *it > 'Z') && (*it < '0' || *it > '9') ) {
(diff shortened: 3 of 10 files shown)


taglib/ape/apetag.cpp
```diff
@@ -65,13 +65,7 @@ namespace
...
     }
 
-    const String upperKey = String(key).upper();
-    for(auto k : invalidKeys) {
-      if(upperKey == k)
-        return false;
-    }
-
-    return true;
+    return std::none_of(invalidKeys.begin(), invalidKeys.end(), [upperKey = String(key).upper()](auto k) { return upperKey == k; });
```
sbooth, line 68: Is the predicate `upperKey = String(key).upper()` evaluated for each iteration?

neheb: no, only once. That's why it's in the capture list.

neheb: btw to clarify, I need to use a lambda here because of ==. String needs to be on the left hand side.

sbooth, line 68: Would `const auto &k` make more sense in the parameter list? Most of the other suggested lambdas use const references for objects so I wonder what is different about this one.

taglib/mp4/mp4atom.cpp
```diff
@@ -97,19 +97,12 @@ MP4::Atom::Atom(File *file)
...
-          if(nextSize == child) {
-            // meta is not a full atom (i.e. not followed by version, flags). It
-            // is followed by the size and type of the first child atom.
-            metaIsFullAtom = false;
-            break;
-          }
-        }
+        // meta is not a full atom (i.e. not followed by version, flags). It
+        // is followed by the size and type of the first child atom.
+        bool metaIsFullAtom = std::none_of(metaChildrenNames.begin(), metaChildrenNames.end(), [nextSize = file->readBlock(8).mid(4, 4)](const auto &child) { return nextSize == child; });
```
sbooth, line 105: Should we prefer `auto metaIsFullAtom` here? It is inconsistent with the suggestion using `find_if` in textidentificationframe.

taglib/mpeg/id3v2/id3v2framefactory.cpp
```diff
@@ -147,11 +147,9 @@ std::pair<Frame::Header *, bool> FrameFactory::prepareFrameHeader(
   }
 #endif
 
-  for(auto it = frameID.cbegin(); it != frameID.cend(); it++) {
-    if( (*it < 'A' || *it > 'Z') && (*it < '0' || *it > '9') ) {
-      delete header;
-      return {nullptr, false};
-    }
+  if(std::any_of(frameID.begin(), frameID.end(), [](auto c) { return (c < 'A' || c > 'Z') && (c < '0' || c > '9'); })) {
```
sbooth, line 150: The original loop uses `cbegin` and `cend`. Is there an advantage to using the const versions?

taglib/riff/riffutils.h
```diff
@@ -44,13 +44,7 @@ namespace TagLib
...
           return false;
 
-        for(auto it = name.begin(); it != name.end(); ++it) {
-          const int c = static_cast<unsigned char>(*it);
-          if(c < 32 || 127 < c)
-            return false;
-        }
-
-        return true;
+        return std::none_of(name.begin(), name.end(), [](char c) { return c < 32 || 127 < c; });
```
sbooth, line 47: Is `char` the correct type for `c`?

taglib/toolkit/tstring.cpp
```diff
@@ -510,20 +510,12 @@ String String::stripWhiteSpace() const
 
 bool String::isLatin1() const
 {
-  for(ConstIterator it = begin(); it != end(); ++it) {
-    if(*it >= 256)
-      return false;
-  }
-  return true;
+  return std::none_of(this->begin(), this->end(), [](auto c) { return c >= 256; });
```
sbooth, line 513: I have the same question as above about using const vs non-const iterators.

taglib/ape/apetag.cpp
neheb: auto vs const auto& boils down to whether or not the parameter is trivially copyable. No real difference except the former being shorter.

taglib/mp4/mp4atom.cpp
neheb: find_if and none_of have different return values. auto offers no advantage here.

taglib/mpeg/id3v2/id3v2framefactory.cpp
neheb: Not really. This piece of code was actually transformed into a for range loop first which is incompatible with cbegin and cend.

taglib/mp4/mp4atom.cpp
sbooth: I understand that, the question I'm raising is whether stylistically we should prefer the use of `auto`. It seems to be the recommendation in https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines

taglib/riff/riffutils.h
neheb: Original code casts it to unsigned. Maybe it'd be good to do that here too.

neheb:
Updated

PR #1110, opened 2023-08-07
clang-tidy: init members by default

neheb:
Found with modernize-use-default-member-init


sbooth:
This is a good change but will take a bit of time to review.

PR #1111, opened 2023-08-08
clang: fix documentation

neheb:
Found with Wdocumentation

### DIFF
--- a/taglib/toolkit/tfile.h
+++ b/taglib/toolkit/tfile.h
@@ -66,8 +66,8 @@ namespace TagLib {
      * Specify which tags to strip either explicitly, or on save.
      */
     enum StripTags {
-      StripNone,  //<! Don't strip any tags
-      StripOthers //<! Strip all tags not explicitly referenced in method call
+      StripNone,  //!< Don't strip any tags
+      StripOthers //!< Strip all tags not explicitly referenced in method call
     };
 
     /*!
@@ -75,8 +75,8 @@ namespace TagLib {
      * types should be synchronized.
      */
     enum DuplicateTags {
-      Duplicate,     //<! Synchronize values between different tag types
-      DoNotDuplicate //<! Do not synchronize values between different tag types
+      Duplicate,     //!< Synchronize values between different tag types
+      DoNotDuplicate //!< Do not synchronize values between different tag types
     };
 
     /*!
--- a/taglib/audioproperties.h
+++ b/taglib/audioproperties.h
@@ -26,6 +26,7 @@
 #ifndef TAGLIB_AUDIOPROPERTIES_H
 #define TAGLIB_AUDIOPROPERTIES_H
 
+#include "taglib.h"
 #include "taglib_export.h"
 
 #include <memory>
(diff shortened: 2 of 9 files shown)


neheb:
added extra somewhat related commit.

Even though this is C++17 now, should C++11 compatibility for the headers be kept?

sbooth:
> Even though this is C++17 now, should C++11 compatibility for the headers be kept?

My personal opinion is that C++17 everywhere is fine, but I'm not sure what @scotchi and @ufleisch think.

neheb:
For exiv2 for example, headers are C++11 compatible. Although [[nodiscard]] is used...

Maybe [[deprecated]] should be used instead. C++ compilers are required to ignore unknown attributes per the standard.

sbooth:
I agree that `[[deprecated]]` makes more sense than the GNU version.

neheb:
you want TAGLIB_DEPRECATED removal here?

ufleisch:
Please do not re-add `TAGLIB_DEPRECATED`. I want TagLib 2.0 to build without any warnings. These functions were deprecated, but must remain there in order to keep TagLib 2.0 source-compatible with TagLib 1.x. The doxygen deprecation comments can still be there but must not have a corresponding deprecation attribute.

I also agree that C++17 everywhere is fine and that `[[deprecated]]` should be used instead of the GNU version.

scotchi:
Just thinking outloud:  I wonder if it'd make sense to set a define via CMake that is something like TAGLIB_INTERN or something, and then to have the deprecation warning only be shown when these methods are used by external projects?

ufleisch:
Concerning `TAGLIB_INTERN`: `MAKE_TAGLIB_LIB` is already defined and could be used.

neheb:
@ufleisch the reason I added it is because compilation fails with -Werror otherwise.

neheb:
Removed anyways.

taglib/toolkit/tfile.h
```diff
@@ -66,17 +66,17 @@ namespace TagLib {
      * Specify which tags to strip either explicitly, or on save.
      */
     enum StripTags {
-      StripNone,  //<! Don't strip any tags
-      StripOthers //<! Strip all tags not explicitly referenced in method call
+      StripNone,  ///< Don't strip any tags
+      StripOthers ///< Strip all tags not explicitly referenced in method call
```
ufleisch, line 70: `//!<` would be more consistent with the doxygen comment style used in the rest of the code.

taglib/toolkit/tfile.h
```diff
@@ -66,17 +66,17 @@ namespace TagLib {
...
 
     /*!
      * Used to specify if when saving files, if values between different tag
      * types should be synchronized.
      */
     enum DuplicateTags {
-      Duplicate,     //<! Synchronize values between different tag types
-      DoNotDuplicate //<! Do not synchronize values between different tag types
+      Duplicate,     ///< Synchronize values between different tag types
+      DoNotDuplicate ///< Do not synchronize values between different tag types
```
ufleisch, line 79: `//!<` would be more consistent with the doxygen comment style used in the rest of the code.

neheb:
//<! is not valid Doxygen.

ufleisch:
I am talking about `//!<`, not `//<!`.

PR #1116, opened 2023-08-15
clang-tidy: return braced init list

neheb:


sbooth:
Could you provide the rationale for this PR?

neheb:
https://clang.llvm.org/extra/clang-tidy/checks/modernize/return-braced-init-list.html

sbooth:
I guess I can buy that justification but these changes still seem like mostly a stylistic choice. Even the justification given by the clang docs sounds a bit like a value judgment (e.g. using needlessly instead of not required). I don't have an issue with this coding style but I'm not fully convinced that it's better. Not having to type the return type is convenient when writing a function but I don't know if it makes the code any more readable or easier to understand.

ufleisch:
I also think that explicit types are more readable. They could also help to catch some errors. Returning braced initializer lists is helpful in generic code.

PR #1120, opened 2023-08-16
clang-tidy: use using instead of typedef

neheb:


neheb:
Updated.

scotchi:
TagLib in general doesn't do indented assigment, e.g.:
````
auto foo        = 1;
auto barbazquux = 2;
````

It feels odd introducing it here.

**Comment by neheb** [comment]
You told me in a previous PR it does. It’s part of the clang-format file I have for taglib

**Comment by scotchi** [comment]
I'm pretty skeptical that I made that comment.  It's never been part of my personal coding style, and TagLib's style is based on how I did things by default 20 years ago.

`grep -r '=  ' *` also doesn't turn up more than a few lines.

**Comment by neheb** [comment]
Found it: https://github.com/taglib/taglib/pull/1101#discussion_r1285121671

**Comment by neheb** [comment]
btw:

```
git grep -r '  = ' taglib | wc -l
324
```

sbooth:
That was me, not @scotchi, and the point of that comment was about limiting the changes in that PR to non-stylistic ones. I've noticed in the past few weeks while reviewing PRs that TagLib does have a general coding style but it isn't uniformly used.

scotchi:
A couple of comments:

- I'm happy that @sbooth has been handling the PRs on these since there's no way I'd have the time or interest to handle this many stylistic changes, but it's probably worth @neheb knowing that @sbooth didn't write any big chunks of TagLib (unless my memory and grep are serving me wrong) there will probably be some things he's less familiar with.
- As I've noted before, a lot of these things really need to be reviewed in a more holistic sense.  Sometimes formatting can make sense in one place where it doesn't in another.  Sometimes code is context dependent.  I'm generally not a huge fan of code tools being semi-blindly applied to a whole repo without reviewing each change to see what the impact on style, git-blame usability, compatibility requirements, etc. are.

PR #1123, opened 2023-08-23
make swap function noexcept

neheb:


ufleisch:
TagLib is not using exceptions at all (the only place where exceptions are used seems to be the 3rd party utf8 library, but its exceptions will not go outside), which I think is a good thing. I would not be happy if there would be `noexcept` specifiers for the whole library. Is there a strong reason why this has to be added?

neheb:
https://clang.llvm.org/extra/clang-tidy/checks/performance/noexcept-move-constructor.html

swap is basically move

neheb:
https://github.com/taglib/taglib/pull/1135 gets rid of swap in most cases.

PR #1124, opened 2023-08-24
tests: switch to GTest

neheb:
GTest seems to be better maintained and has more features.

Most of this commit was done with sed and clang-format.

Replacement for: https://github.com/taglib/taglib/pull/825


neheb:
Interesting that I cannot reproduce this locally. Even with containers.

neheb:
https://github.com/google/googletest/issues/4079

agh.

ufleisch:
According to https://github.com/google/googletest/issues/3552#issuecomment-918520846, this will probably never be fixed for GCC 11.

neheb:
Guess I'll abandon this. Oh well.

neheb:
Maybe the overall solution is to mandate a more recent version of gtest. Maybe implement as a git subpproject.

ufleisch:
The initial intention was to provide something which "is better maintained and has more features" than CppUnit. However, if I look at the current situation with the incompatibility with GCC 11 and that this will probably not be fixed, I fear that switching from CppUnit to GTest will cause more pain than joy to the users of TagLib. After all, they have successfully set up CppUnit, and it is working quite well. The additional features of GTest, e.g. GMock, may be nice, but I do not see that we need them for TagLib. So I would suggest that we stick with CppUnit for now.

neheb:
Catch2 could be an alternative. Although version 3 is completely incompatible with 2. Weird choice by upstream.

ufleisch:
The important thing for version 2.0.0 is to come up with a stable API, so that later versions can remain binary compatible. I there is really a need to replace the unit testing framework, this could be done later (e.g. when nobody is using a combination of compiler and unit testing framework which is not working). So I would suggest that we put our energy in the things which have to be ready with the release of version 2.0.0. What could be helpful is if you could publish your scripts which can migrate the test cases using `sed` and `clang-format`.

neheb:
I think I posted the clang-format script once. There was no interest. The sed stuff is very unpolished and still manual.

PR #1125, opened 2023-08-25
use if constexpr for template

neheb:


scotchi:
My gut feeling on this is that it's adding code complexity for something that's a relatively minor case:  constants in the code.  There just aren't that many of them.  This feels like an unnecessary micro-optimization that's being done just because it can be.  I'd prefer to see this actually showing up on profiling as a hot-point before making the code more complex just for the hell of it.

scotchi:
Actually, now looking at it, this seems to be a new type of basically limited template specialization that I'm unfamiliar with, but again, I feel like my point stands:  things that add code and make it more complex without increasing performance in some measurable way, I tend to be skeptical of.  (In fact, a lot of TagLib intentionally makes suboptimal choices where it makes the code simpler, and doesn't bother with them unless they show up in profiling.)

neheb:
hmmm? it's reducing the template from 2 to 1.

sbooth:
Isn't there an issue with an undefined return value if `nextUTF16` is called with a type that isn't `char` or `wchar_t`?

neheb:
Yes. In the case of this template, it's local to the file and only used 2 times

sbooth:
That's true but rewriting it this way seems to substitute compile-time checks for runtime checks. Even though the callers of this function are well-known with the current code there is no possibility for an undefined return value.

neheb:
eh no, if constexpr is specifically at compiletime.

sbooth:
My words were imprecise but here is what I meant:

If `nextUTF16` is called with a `float` argument (for which no template exists), the `if constexpr` method generates a compiler warning by default:

```
warning: non-void function does not return a value [-Wreturn-type]
in instantiation of function template specialization 'nextUTF16<float>' requested here
```

but the code compiles and unless I'm mistaken (which is definitely a possibility) at runtime will result in UB because of the lack of return value.

The current way fails to link when called with a `float` argument:

```
Undefined symbols for architecture x86_64:
  "unsigned short nextUTF16<float>(float const**)", referenced from:
```

so it is not possible to get UB at runtime.

PR #1126, opened 2023-08-30
manual range loop conversions

neheb:


ufleisch:
The changes which convert an iteration using a constant iterator (either with `cbegin()`/`cend()` or implicitly on a `const` container) negatively impact performance. For example

```
  for(auto it = objects.cbegin(); it != objects.cend(); ++it) {
    data.append((*it)->render(file));
  }
```

is changed in *asffile.cpp* to

```
  for(const auto &object : objects) {
    data.append(object->render(file));
  }
```

which is equivalent to non-const iteration

```
  for(auto it = objects.begin(); it != objects.end(); ++it) {
    data.append((*it)->render(file));
  }
```

The problem is that `begin()` will call `detach()` (thereby making a copy of the container) on the container which `begin() const` and `cbegin()` would avoid.

This could be solved by using `std::as_const()` where `cbegin()`/`cend()` where used.

```
#include <utility>
(..)
  for(const auto &object : std::as_const(objects)) {
    data.append(object->render(file));
  }
```

An alternative would be to make a const copy, but `std::as_const()` is probably more efficient. See [Goodbye, Q_FOREACH](https://www.kdab.com/goodbye-q_foreach/) for a blog post on that subject related to Qt.

neheb:
That's just strange. Sounds like every const auto & for range loop should use std::as_const then.

ufleisch:
Not every for loop range has to be changed, just the ones where I used `cbegin()`/`cend()`. I only used these where the container is not `const`. Whether a const iterator is returned by `begin()` depends only on the container.

neheb:
Here's a fun issue I noticed:

```
--- a/taglib/mpeg/id3v2/id3v2tag.cpp
+++ b/taglib/mpeg/id3v2/id3v2tag.cpp
@@ -26,6 +26,7 @@
 #include "id3v2tag.h"
 
 #include <algorithm>
+#include <utility>
 
 #include "tfile.h"
 #include "tbytevector.h"
@@ -372,8 +373,7 @@ void ID3v2::Tag::removeFrame(Frame *frame, bool del)
 
 void ID3v2::Tag::removeFrames(const ByteVector &id)
 {
-  auto frames = d->frameListMap[id];
-  for(const auto &frame : frames)
+  for(const auto &frame : std::as_const(d->frameListMap[id]))
     removeFrame(frame, true);
 }
 
```

This change fails the tests. Any idea why?

ufleisch:
I would really advice to use `std::as_const()` only where `cbegin()`/`cend()` were used. The original code was

```
  const FrameList l = d->frameListMap[id];
  for(auto it = l.begin(); it != l.end(); ++it)
    removeFrame(*it, true);
```

so I would change it to

```
  const FrameList l = d->frameListMap[id];
  for(const auto &frame : l)
    removeFrame(frame, true);
```

or, to be a bit closer on your current changes

```
  const auto frames = d->frameListMap[id];
  for(const auto &frame : frames)
    removeFrame(frame, true);
```

This adds `const` to your changes. The const copy makes sure that a const iterator is used. Making a copy is advisable in the case of using an iterator because it is always better to call `begin()` and `end()` on the same instance.

In your case, the segmentation fault probably happens because `removeFrame()` is calling `erase()` on `d->frameListMap[id]`, which will invalidate the iterator you are using. So making a copy is really necessary here.

ufleisch:
@sbooth I have created a PR to restore the const iterators replaced in #1107 and also reduced the lengths of some lines which were a bit too long, I think we should not have lines longer than 100 characters. See #1128.

neheb:
> Not every for loop range has to be changed, just the ones where I used `cbegin()`/`cend()`. I only used these where the container is not `const`. Whether a const iterator is returned by `begin()` depends only on the container.

they don't have to but it seems like a good idea anyway.

neheb:
added as_const.

ufleisch:
I agree that preferring `cbegin()`/`cend()` over `begin()`/`end()` is a good idea. But it does not make sense to call `std::as_const()` on a container which is already `const`, so I suggest that you remove it in these cases.

neheb:
Fixed.

ufleisch:
@neheb I will replace const container copies in my last commit with `std::as_const()` where possible. I am currently testing it and will push it when finished.

PR #1134, opened 2023-09-06
add do/while to mandate ;

neheb:


ufleisch:
Could you come up with a better commit message?

neheb:
done.

ufleisch:
Sorry, I know that I made a comment about lines being longer than 100 characters. Actually, I am more in the 80 characters camp, but I have found that I have a hard time getting this through when I am working with other people. But most lines in TagLib are not longer than 80 characters, and I think of 100 characters as a hard limit when the restriction to 80 characters would make the code less readable.
In this PR, the `\` characters are definitely too far on the right side, I even had to collapse the file tree in GitHub's "Files changed" tab to prevent the lines from being wrapped. So please reformat the code to make the lines smaller.

neheb:
Fixed.

PR #1135, opened 2023-09-06
default some more

neheb:
swap is not needed here as shared_ptr is copyable.

### DIFF
--- a/taglib/mp4/mp4coverart.cpp
+++ b/taglib/mp4/mp4coverart.cpp
@@ -47,14 +47,8 @@ MP4::CoverArt::CoverArt(Format format, const ByteVector &data) :
   d->data = data;
 }
 
-MP4::CoverArt::CoverArt(const CoverArt &item) = default;
-
-MP4::CoverArt &
-MP4::CoverArt::operator=(const CoverArt &item)
-{
-  CoverArt(item).swap(*this);
-  return *this;
-}
+MP4::CoverArt::CoverArt(const CoverArt &) = default;
+MP4::CoverArt &MP4::CoverArt::operator=(const CoverArt &) = default;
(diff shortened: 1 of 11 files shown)


taglib/mp4/mp4coverart.cpp
```diff
@@ -47,14 +47,8 @@ MP4::CoverArt::CoverArt(Format format, const ByteVector &data) :
...
 
-MP4::CoverArt::CoverArt(const CoverArt &item) = default;
-
-MP4::CoverArt &
-MP4::CoverArt::operator=(const CoverArt &item)
-{
-  CoverArt(item).swap(*this);
-  return *this;
-}
+MP4::CoverArt::CoverArt(const CoverArt &)                 = default;
```
ufleisch, line 50:
```suggestion
MP4::CoverArt::CoverArt(const CoverArt &) = default;
```

taglib/mp4/mp4item.cpp
```diff
@@ -53,14 +53,8 @@ MP4::Item::Item() :
...
 
-MP4::Item::Item(const Item &item) = default;
-
-MP4::Item &
-MP4::Item::operator=(const Item &item)
-{
-  Item(item).swap(*this);
-  return *this;
-}
+MP4::Item::Item(const Item &)                 = default;
```
ufleisch, line 56:
```suggestion
MP4::Item::Item(const Item &) = default;
```

neheb: Fixed.

taglib/mp4/mp4coverart.cpp
neheb: Fixed.

PR #1139, opened 2023-09-06
use using declarations

neheb:


ufleisch:
Is there a reason why these `using` declarations are not below `public:`? Both seems to work, is there a difference?

neheb:
I don't think using declarations care about public/private since they inherit that from the base class.

ufleisch:
You are right, that standard (ISO/IEC 14882:2020(E), 9.9 The using declaration, 19) says

> A synonym created by a using-declaration has the usual accessibility for a member-declaration. A using-declarator that names a constructor does not create a synonym; instead, the additional constructors are accessible if they would be accessible when used to construct an object of the corresponding base class, and the accessibility of the using-declaration is ignored.

But I would consider it more readable if these `using`-declarations would be under the `public:` accessibility to make it clear.

neheb:
Fixed.

PR #1140, opened 2023-09-09
replace vector with array

neheb:
No need for vector. Allows simplifying some code.

### DIFF
--- a/taglib/tagunion.cpp
+++ b/taglib/tagunion.cpp
@@ -33,6 +33,8 @@
 #include "xiphcomment.h"
 #include "infotag.h"
 
+#include <array>
+
 using namespace TagLib;
 
 #define stringUnion(method)                                          \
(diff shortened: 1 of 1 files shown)


taglib/tagunion.cpp
```diff
@@ -199,12 +195,5 @@ void TagUnion::setTrack(unsigned int i)
...
 {
-  if(d->tags[0] && !d->tags[0]->isEmpty())
-    return false;
-  if(d->tags[1] && !d->tags[1]->isEmpty())
-    return false;
-  if(d->tags[2] && !d->tags[2]->isEmpty())
-    return false;
-
-  return true;
+  return std::none_of(d->tags.begin(), d->tags.end(), [](auto tag) { return tag && !tag->isEmpty(); });
```
ufleisch, line 198:
```suggestion
  return std::none_of(d->tags.begin(), d->tags.end(),
    [](auto tag) { return tag && !tag->isEmpty(); });
```
The line is longer than 100 characters.

neheb: Fixed.

PR #1142, opened 2023-09-09
add utf8cpp git submodule

neheb:


ufleisch:
Could you add instructions in INSTALL.md how to build TagLib with utf8cpp as a git submodule? CMake complains about not finding utf8cppConfig.cmake.

CMakeLists.txt
```diff
@@ -147,6 +147,8 @@ endif()
 
 configure_file(taglib/taglib_config.h.cmake "${CMAKE_CURRENT_BINARY_DIR}/taglib_config.h")
 
+find_package(utf8cpp REQUIRED)
```
ufleisch, line 150: This no longer builds with the submodule, I would change it like this to work with both a library (preferred if existing) or a git submodule.
```suggestion
find_package(utf8cpp QUIET)
if(NOT utf8cpp_FOUND)
  if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/utfcpp/CMakeLists.txt)
    add_subdirectory("3rdparty/utfcpp")
    message(STATUS "Using utfcpp from ${utf8cpp_SOURCE_DIR}")
  else()
    message(FATAL_ERROR
      "utfcpp not found. Either install package (probably utfcpp, utf8cpp, or libutfcpp-dev) "
      "or fetch the git submodule using\n"
      "git submodule update --init")
  endif()
else()
  message(STATUS "Using utfcpp ${utf8cpp_VERSION} from ${utf8cpp_CONFIG}")
endif()
```

neheb: Second commit is a test commit to make sure external utfcpp works.

jacobvosmaer:
@ufleisch out of curiosity, what exactly did we gain by this change?

This dependency is a header-only library so none of the benefits of shared libraries (space saving, security updates) apply. As a downstream user of TagLib I only see the negative of having to install a new dependency for TagLib.

ufleisch:
The downside is for the developer using TagLib who now needs to provide another dependency, either as a development package of a distribution or as a git sub module. However, if we would keep on bundling an old version of utf8cpp and some bug or even vulnerability of this header-only library is discovered, we would have to provide an update of TagLib. With the current solution, distributions just have to update utf8cpp and rebuild TagLib. So the security benefit still applies, at least at build time, and space is also saved, at least in the form that we do not bundle duplicated code, which is convenience and not best practice. With larger header-only libraries such as boost, you would probably not copy them either.

jacobvosmaer:
Thanks @ufleisch. Yes I also guess that most consumers will install from a package manager which handles the dependency for them.

Still, in the case of a security issue, you would have to update the submodule too, wouldn't you?

ufleisch:
Yes, you would have to update the submodule or the package.
By the way, we released TagLib 2.0.1, so you could now proceed with the integration into taglib-ruby.

jacobvosmaer:
@ufleisch thanks! Yes I saw your announcement about 2.0.1 on the mailing list. I'm working on integrating it here: https://github.com/robinst/taglib-ruby/pull/143

PR #1145, opened 2023-09-16
clang-tidy: use nodiscard

neheb:


ufleisch:
I have here the same feelings as with `noexcept` (#1123): This defaces the code without a big benefit.

PR #1152, opened 2023-09-29
cmake: support both relative and absolute paths for pc files

### LINKED ISSUE #1098: 0.13.1 regression: absolute `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR` are not supported anymore
https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#module:GNUInstallDirs says that `CMAKE_INSTALL_*` can accept both absolute and relative paths:

```
CMAKE_INSTALL_<dir>

    Destination for files of a given type. This value may be passed to the DESTINATION options of [install()](https://cmake.org/cmake/help/latest/command/install.html#command:install) commands for the corresponding file type. It should typically be a path relative to the installation prefix so that it can be converted to an absolute path in a relocatable way (see CMAKE_INSTALL_FULL_<dir>). However, an absolute path is also allowed.
```

`nixpkgs` package repository usually uses `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR` outside the `CMAKE_INSTALL_PREFIX`.

`taglib-0.13.0` used to work as is.

But with https://github.com/taglib/taglib/commit/967c0eefedcf7dea71832400791bed7246a41299 `tagilb.pc` files now append extra unexpected prefix:

Building:

```
$ mkdir b
$ cd b
$ cmake .. -DCMAKE_INSTALL_LIBDIR=/absolute-lib -DCMAKE_INSTALL_INCLUDEDIR=/absolute-include -DCMAKE_INSTALL_PREFIX=/usr
$ make
$ make install DESTDIR=${PWD}/_i_
```

Inspecting the paths:

```
$ cat _i_/absolute-lib/pkgconfig/taglib.pc
prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}//absolute-lib
includedir=${prefix}//absolute-include
```

Note the unexpected `${prefix}/`.

Noticed the deficiency when tried to package `0.13.1` as https://github.com/NixOS/nixpkgs/pull/245926.

So far we are working it around by using relative paths, but we would like to switch back to absolute paths to split out development outputs.

Thanks!

jpalus:
cmake supports both absolute and relative to prefix install dirs. #1071 effectively mandated relative paths only. check if install dir is relative or absolute and either include ${prefix} in pc file or not

Fixes #1098


ufleisch:
Thanks, I just removed two spaces to make the formatting consistent.

PR #1153, opened 2023-10-01
Map: Support Braced List Initialization

complexlogic:
Adds C++11 braced list initialization to `Map`, same as for `std::map`. For example:

```C++
TagLib::Map<TagLib::String, int> m1 {
  {"ONE", 1},
  {"TWO", 2},
  {"THREE", 3}
};
```

This can be useful for implementing a static lookup table.

Tests have been added for both construction and assignment.

### DIFF
--- a/taglib/toolkit/tmap.h
+++ b/taglib/toolkit/tmap.h
@@ -28,6 +28,8 @@
 
 #include <map>
 #include <memory>
+#include <initializer_list>
+#include <utility>
 
 #include "taglib.h"
 
@@ -73,6 +75,11 @@ namespace TagLib {
      * pass-by-value usage.
      */
     Map(const Map<Key, T> &m);
+
+    /*!
+     * Constructs a Map with the contents of the braced initializer list.
+     */
+    Map(std::initializer_list<std::pair<const Key, T>> init);
 
     /*!
      * Destroys this instance of the Map.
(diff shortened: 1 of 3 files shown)


ufleisch:
Thanks, looks good. I actually missed this recently. I will merge it if no one else objects.
This would also be useful for `TagLib::List`, if you find the time.

complexlogic:
I'll make another PR for `TagLib::List`.

complexlogic:
Hey @ufleisch, after some thought I believe my implementation for assignment is not correct here. Since `Map` is *implicitly shared*, we don't want to assign the members of the private class object directly, since they would be potentially shared with other `Map` instances. The correct thing to do would be to reassign the entire shared pointer, which would decrement the reference count of the pointed-to private class object (and cause it to be deleted it if it's zero).

So instead of:
```C++
template <class Key, class T>
Map<Key, T> &Map<Key, T>::operator=(std::initializer_list<std::pair<const Key, T>> init)
{
  d->map = init;
  return *this;
}
```

...it should actually be:
```C++
template <class Key, class T>
Map<Key, T> &Map<Key, T>::operator=(std::initializer_list<std::pair<const Key, T>> init)
{
  d = std::make_shared<MapPrivate<Key, T>>(init);
  return *this;
}
```

Thoughts?

ufleisch:
You are probably right. Or you could use the following code, which is a bit more "idiomatic" and should make exactly the same thing.

```
template <class Key, class T>
Map<Key, T> &Map<Key, T>::operator=(std::initializer_list<std::pair<const Key, T>> init)
{
  Map(init).swap(*this);
  return *this;
}
```

PR #1154, opened 2023-10-01
List: Support Braced List Initialization

complexlogic:
Adds C++11 braced list initialization to `List`, same as for `std::list`. For example:

```C++
TagLib::List<int> m1 {
  1,
  2,
  3
};
```

- Tests have been added for construction and assignment, for both the regular and pointer specialization.
- For assignment, verified that the members are deleted when the list is of pointer type and auto delete is set to true

### DIFF
--- a/tests/test_list.cpp
+++ b/tests/test_list.cpp
@@ -34,6 +34,7 @@ class TestList : public CppUnit::TestFixture
   CPPUNIT_TEST_SUITE(TestList);
   CPPUNIT_TEST(testAppend);
   CPPUNIT_TEST(testDetach);
+  CPPUNIT_TEST(bracedInit);
   CPPUNIT_TEST_SUITE_END();
 
 public:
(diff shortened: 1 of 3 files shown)


tests/test_list.cpp
```diff
@@ -71,6 +72,56 @@ class TestList : public CppUnit::TestFixture
...
+      2,
+      3
+    };
+    CPPUNIT_ASSERT_EQUAL(3, static_cast<int>(l3.size()));
+    CPPUNIT_ASSERT_EQUAL(1, l3[0]);
+    CPPUNIT_ASSERT_EQUAL(2, l3[1]);
+    CPPUNIT_ASSERT_EQUAL(3, l3[2]);
+
+    List<int*> l4 = {
+      new int(1),
```
neheb, line 109: how nasty, I can't tell if this is a memory leak or not.

complexlogic: The `ListPrivate` shared pointer is assigned a new `ListPrivate` object constructed from the braced list. The [assignment operator of the shared pointer](https://en.cppreference.com/w/cpp/memory/shared_ptr/operator%3D) releases ownership of the old object. When its reference count reaches 0 the object is deleted automatically.

So there shouldn't be any memory leak as long as autodelete is set on the `List`.

neheb: I trust your judgement here.

In a different project, GCC's -fanalyzer complained when using shared_ptr(make_shared I think) and initializer_list saying there was a memory leak. I fixed by avoiding the latter and using std::move. No idea how much of a false positive it was.

PR #1155, opened 2023-10-01
Proofreading

nmariusp:


bindings/c/tag_c.h
```diff
@@ -58,7 +58,7 @@ extern "C" {
 
 /*
  * These are used for type provide some type safety to the C API (as opposed to
- * using void *, but pointers to them are simply cast to the corresponding C++
+ * using void * ), but pointers to them are simply cast to the corresponding C++
```
ufleisch, line 61: What about changing "These are used for type provide some type safety" to "These are used to provide some type safety"? Why a space after "void *"?

ufleisch (commented):
Thanks! I would not modify _id3v2.4.0-structure.txt_ as this is a copy of an "original standard document" which should stay as it is with all its misspellings and accidental tab characters.

PR #956, opened 2020-04-16
Attached picture c binding

kdeng00:
Addresses the original Issue #953 to provide c bindings for attaching cover art to mpeg files

Issue #953, opened 2020-04-10
id3v2 writer and reader for taglib_c


torikulhabib:
I am having trouble attaching images using taglib_c to my application.
https://github.com/torikulhabib/niki
help me please.
https://github.com/taglib/taglib/blob/master/bindings/c/tag_c.h

kdeng00:
What issues are you experiencing? Could you provide some code and your build configuration?

torikulhabib:
I am constrained in the code attaching the picture to taglib_c
https://github.com/taglib/taglib/tree/master/bindings/c
only available for the album title ... and there is no attachment for the picture. this is the sample code that I used in valac https://github.com/torikulhabib/niki/blob/789356cc045231ebfbfc14e11509d269eed9bacb/src/MediaEditor.vala#L378

kdeng00:
Hey @torikulhabib, check out this branch to see if it meets your needs:
https://github.com/kdeng00/taglib/tree/attached_picture_c_binding

I added some wrappers for interacting with covers. Note: It's only applicable for certain use cases. Can be further improved.

kdeng00:
Hey @torikulhabib,

Check out my branch again. I added some wrappers similar to the ones to mentioned in the email.

To attach the an image follow this steps:
1. Create TagLib::MPEG::File wrapper
2. Create TagLib::ID3v2::Tag wrapper
3. Check if the picture frame list is empty
4. (Optional): Remove any **APIC** (Frame where cover art is stored) if necessary
5. Create ID3v2::AttachedPicture wrapper
6. Set the picture type
7. Set the picture
8. Add the ID3v2::AttachedPicture wrapper to ID3v2::Tag wrapper
9. Save the TagLib::MPEG::File song file
10. Free any resources

Check the [tag_c.h](https://github.com/kdeng00/taglib/blob/attached_picture_c_binding/bindings/c/tag_c.h#L140) file. Starts at line 140 and ends at 238

torikulhabib:
thanks this working.
and need add set mimetype;

kdeng00:
I can add that. I'll clean up some things and open a PR to get it in the master branch of the master repo

torikulhabib:
add delete all tag please

kdeng00:
Delete all tags?

torikulhabib:
i mean clear all tag.
there are option to clear all.

kdeng00:
I'll let you know when I get to it

kdeng00:
[...]

PR #1156, opened 2023-10-08
Unified interface for complex properties like pictures (#94)

ufleisch:
Provides a dynamic interface for properties which cannot be represented with simple strings, e.g. pictures. The keys of such properties can be queried using `complexPropertyKeys()`, which could return for example ["PICTURE"]. The property can then be read using `complexProperties("PICTURE")`, which will return a list of variant maps containing the picture data and attributes. Adding a picture is as easy as

    t->setComplexProperties("PICTURE", {
      {
        {"data", data},
        {"pictureType", "Front Cover"},
        {"mimeType", "image/jpeg"}
      }
    });

Complex properties are flexible enough for most frames. At the moment they are implemented for pictures in APE, ASF, MP4, ID3v2, Vorbis and FLAC. For ID3v2, GEOB frames are supported too. Embedded pictures can be tested using the _tagreader_ and _tagwriter_ examples.

Related issues and pull requests are #94, #397, #666, #773, #951, #952, #953. When this PR is merged, it should be easy to implement a replacement for #956.

### DIFF
--- a/taglib/asf/asftag.cpp
+++ b/taglib/asf/asftag.cpp
@@ -29,6 +29,8 @@
 #include <utility>
 
 #include "tpropertymap.h"
+#include "asfattribute.h"
+#include "asfpicture.h"
 
 using namespace TagLib;
 
@@ -372,4 +374,55 @@ PropertyMap ASF::Tag::setProperties(const PropertyMap &props)
   }
 
   return ignoredProps;
+}
+
+StringList ASF::Tag::complexPropertyKeys() const
+{
+  StringList keys;
+  if(d->attributeListMap.contains("WM/Picture")) {
+    keys.append("PICTURE");
+  }
+  return keys;
+}
+
+List<VariantMap> ASF::Tag::complexProperties(const String &key) const
+{
+  List<VariantMap> properties;
+  const String uppercaseKey = key.upper();
+  if(uppercaseKey == "PICTURE") {
+    const AttributeList pictures = d->attributeListMap.value("WM/Picture");
+    for(const Attribute &attribute : pictures) {
+      ASF::Picture picture = attribute.toPicture();
+      VariantMap property;
+      property.insert("data", picture.picture());
+      property.insert("mimeType", picture.mimeType());
+      property.insert("description", picture.description());
+      property.insert("pictureType",
+        ASF::Picture::typeToString(picture.type()));
+      properties.append(property);
+    }
+  }
+  return properties;
+}
+
+bool ASF::Tag::setComplexProperties(const String &key, const List<VariantMap> &value)
+{
+  const String uppercaseKey = key.upper();
+  if(uppercaseKey == "PICTURE") {
+    removeItem("WM/Picture");;
+
+    for(auto property : value) {
+      ASF::Picture picture;
+      picture.setPicture(property.value("data").value<ByteVector>());
+      picture.setMimeType(property.value("mimeType").value<String>());
+      picture.setDescription(property.value("description").value<String>());
+      picture.setType(ASF::Picture::typeFromString(
+        property.value("pictureType").value<String>()));
+      addAttribute("WM/Picture", Attribute(picture));
+    }
+  }
+  else {
+    return false;
+  }
+  return true;
 }
--- a/taglib/toolkit/tvariant.cpp
+++ b/taglib/toolkit/tvariant.cpp
@@ -1,0 +1,391 @@
+/***************************************************************************
+    copyright            : (C) 2023 by Urs Fleisch
+    email                : ufleisch@users.sourceforge.net
+ ***************************************************************************/
+
+/***************************************************************************
+ *   This library is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU Lesser General Public License version   *
+ *   2.1 as published by the Free Software Foundation.                     *
+ *                                                                         *
+ *   This library is distributed in the hope that it will be useful, but   *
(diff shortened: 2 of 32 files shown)


taglib/toolkit/tvariant.cpp
```diff
@@ -0,0 +1,391 @@
...
+  std::ostream &s;
+};
+
+} // namespace
+
+class Variant::VariantPrivate
+{
+public:
+  VariantPrivate() = default;
+  VariantPrivate(const StdVariantType &v) : data(v) {}
```
neheb, line 199: wonder if pass by value makes sense here.

sbooth:
@ufleisch this looks great!

complexlogic:
This won't work for FLAC files. In the case of FLAC, pictures are stored in `METADATA_BLOCK_PICTURE`, which is not part of the Vorbis Comment (in contrast to Ogg). In TagLib's implementation, you access the pictures of a FLAC file using `FLAC::File::pictureList`, instead of `Ogg::XiphComment::pictureList`.

So you'll need to add override functions for `FLAC::File` that handle the case where `key` is equal to PICTURE, otherwise returning the `Tag` implementation.

complexlogic:
Do you plan to add an API to remove complex properties? For example, suppose a user wants to delete a picture (without replacing it with anything).
 
The parallel here would be the `PropertyMap` API. You can delete properties by setting the value to an empty `StringList`. But a likewise operation is not currently possible with complex properties.

sbooth:
FLAC is tricky because it might contain album art in three places: ID3v2 `APIC` frames, Base-64 encoded `METADATA_BLOCK_PICTURE` blocks in the Xiph comment (as with Vorbis), and in the native `METADATA_BLOCK_PICTURE` metadata blocks. I don't know if I've seen a file with album art in all three locations but two isn't terribly uncommon and can be quite confusing for end users.

When it comes to complex properties what would a library client expect- to see only the artwork from `METADATA_BLOCK_PICTURE` metadata blocks or all available art?

I've sometimes wondered if `File` would benefit from something like a `standardize()` member function. In this context I imagine it would copy the ID3v2 tag to the Xiph comment, and move any album art in the Xiph comment to native metadata blocks.

taglib/toolkit/tvariant.cpp
ufleisch: GCC generates the same code, but with Clang, this save two assembly instructions, so I have adapted it.

ufleisch:
Good you mentioned the FLAC case, I forgot this. I have now overriden the functions in `FLAC::File` and adapted _tagwriter_ to access complex properties via `File`. For pictures, the `FLAC::File` is used (and thus METADATA_BLOCK_PICTURE), other complex properties are looked up via the default `TagLib::File` implementations, which behaves the same as for the normal properties, i.e. it goes via `tag()`, which is typically a `TagUnion` and sets the properties in all tags which support them. So the currently sole case is a PICTURE which is stored in METADATA_BLOCK_PICTURE and a GENERALOBJECT which is stored in the ID3v2 tag of a FLAC file as a GEOB frame. Typical use cases should go through `FileRef::file()` and will only use pictures, which should work as expected.

Removing complex properties already worked - you just have to pass an empty list. I have now documented this explicitly, added a test case and also supported this operation in _tagwriter_.

taglib/asf/asftag.cpp
```diff
@@ -373,3 +375,52 @@ PropertyMap ASF::Tag::setProperties(const PropertyMap &props)
...
+      ASF::Picture picture;
+      picture.setPicture(property.value("data").value<ByteVector>());
+      picture.setMimeType(property.value("mimeType").value<String>());
+      picture.setDescription(property.value("description").value<String>());
+      picture.setType(ASF::Picture::typeFromString(
+        property.value("pictureType").value<String>()));
+      addAttribute("WM/Picture", Attribute(picture));
+    }
+  }
+  else {
```
neheb, line 422: Could avoid the else with early return

ufleisch: Yes, but I intentionally coded it this way to allow adding another property without having to restructure the code. Also the `complePropertyKeys()` methods could be only one statement, but this way, it is simple to add other keys.

taglib/toolkit/tvariant.cpp
```diff
@@ -0,0 +1,391 @@
...
+  std::ostream &s;
+};
+
+} // namespace
+
+class Variant::VariantPrivate
+{
+public:
+  VariantPrivate() = default;
+  VariantPrivate(StdVariantType v) : data(v) {}
```
neheb, line 199: why is this not moved?

ufleisch: I actually do not like move semantics in C++. Reconsidering it again, I changed it back to a good old const reference. Since all types contained in the variant are cheap to copy, this is not really a rewarding use of move semantics. I tried it with `std::move()` before and GCC was generating larger code. And I do not want to open the door for more `std::move()` "improvements" in TagLib. After all, TagLib uses implicitly shared strings and containers, which are in my opinion a less error-prone and more user friendly way to avoid costly copies and reallocations.

complexlogic:
@ufleisch Now that this has been merged, I recommend closing the issues/PRs that you referenced in the original comment.

whyman:
Is this going to make it into a 1.x release? Or is it a 2.0 thing?

ufleisch:
It is a 2.0 thing, but this does not matter that much since TagLib 2.0 is now developed on the master branch and will be the next (binary incompatible but mostly source compatible) release.

PR #1160, opened 2023-10-18
List: Add Sort Functions

complexlogic:
I would like to be able to sort the objects contained in a `List`. We cannot use `std::sort()` on `List` because the iterators do not support random access. The underlying STL class `std::list` does have a sort function, however. This PR adds sort functions to `List` based on `std::list::sort()`. See the [Cppreference page](https://en.cppreference.com/w/cpp/container/list/sort) for more information.

The basic function without arguments will sort the `List` in ascending order. The class contained in `List` needs to support the `<` operator.

Additionally, there is an overload which takes a function object `Comp`. You can use this to provide custom sorting logic to the function. For example, a lambda based on some attribute of a class that can be accessed through a getter function:
```C++
list.sort([](const auto a, const auto b){ return a.getter() < b.getter(); }); 
``` 

Tests have been added for the basic sorting function and for the overload.

### DIFF
--- a/taglib/toolkit/tlist.h
+++ b/taglib/toolkit/tlist.h
@@ -277,7 +277,7 @@ namespace TagLib {
     bool operator!=(const List<T> &l) const;
 
     /*!
-     * Sorts this list in ascending order using operator< of T
+     * Sorts this list in ascending order using operator< of T.
      */
     void sort();
 
--- a/taglib/toolkit/tlist.h
+++ b/taglib/toolkit/tlist.h
@@ -276,6 +276,19 @@ namespace TagLib {
      */
     bool operator!=(const List<T> &l) const;
 
+    /*!
+     * Sorts this list in ascending order using operator< of T
+     */
+    void sort();
+
+    /*!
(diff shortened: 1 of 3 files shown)


taglib/toolkit/tlist.h
```diff
@@ -276,6 +276,19 @@ namespace TagLib {
      */
     bool operator!=(const List<T> &l) const;
 
+    /*!
+     * Sorts this list in ascending order using operator< of T
```
ufleisch, line 280:
```suggestion
     * Sorts this list in ascending order using operator< of T.
```

taglib/toolkit/tlist.h
```diff
@@ -276,6 +276,19 @@ namespace TagLib {
...
+     */
+    void sort();
+
+    /*!
+     * Sorts this list in ascending order using using the comparison
```
ufleisch, line 285:
```suggestion
     * Sorts this list in ascending order using the comparison
```

ufleisch (commented):
Just two minor things concerning the comments.

Issue #987, opened 2020-12-27
taglib_file_new from memory


dajuric:
Create a new function where files can be read from memory and written to memory. That would be useful when files are obtained from HTTP source using curl.

ufleisch:
You can already initialize a `ByteVectorStream` from a `ByteVector` and then create a `FileRef` (or a subclass of `File`) from it.

scotchi:
I think this person was asking for such in the C bindings.  I have no particularly strong feelings on such.

ufleisch:
Implemented with #1162.

PR #1162, opened 2023-10-22
Improve FileRef, C bindings to read file from memory (#987)

ufleisch:
Use `FileRef` for `TagLib_File` (C bindings), which allows to drop its deprecated `create()` method and delete duplicated code. It also allows to easily provide a feature to access files from memory with the C bindings (#987).
`FileRef` now has all the property methods, so that it is possible to do without its `file()` method, the use of which is discouraged in the documentation.
Noteworthy changes: `FileRef` is no longer virtual (the destructor was the only virtual method), and printing a `String` to `cout` will now use UTF-8 instead of ISO-8859-1 encoding.

### DIFF
--- a/bindings/c/tag_c.cpp
+++ b/bindings/c/tag_c.cpp
@@ -92,67 +92,79 @@ void taglib_free(void* pointer)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-// TagLib::File wrapper
+// TagLib::FileRef wrapper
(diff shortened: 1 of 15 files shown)


ufleisch:
Resolved https://github.com/taglib/taglib/commit/9b0ab877e97397fbb0362cba57be846cc50bfb77#commitcomment-130650351.

Issue #1163, opened 2023-11-01
SEGV on taglib library when writing tags


CarlosAndresRamirez:
**Description**
Taglib through version TagLib 1.13.1 is vulnerable to special crafted headers in .wav files. This degrades the availability and integrity of software systems, libraries and devices that rely on taglib library.

**Technical Details**
After media file has been already saved, the malformed headers trigger removal of chunks at function TagLib::RIFF::File::removeChunk(), in file rifffile.cpp that eventually produces a crash when calling RIFF::File::updateGlobalSize()

**ASAN Output**
**./tagwriter -t algo tagcrash-1.wav**

AddressSanitizer:DEADLYSIGNAL
=================================================================
==3824310==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000014 (pc 0x5555557305db bp 0x7fffffffd520 sp 0x7fffffffd460 T0)
==3824310==The signal is caused by a READ memory access.
==3824310==Hint: address points to the zero page.
    #0 0x5555557305db in std::__detail::_MakeUniq<TagLib::ByteVector::ByteVectorPrivate>::__single_object std::make_unique<TagLib::ByteVector::ByteVectorPrivate, TagLib::ByteVector::ByteVectorPrivate&, int, unsigned int&>(TagLib::ByteVector::ByteVectorPrivate&, int&&, unsigned int&) /usr/lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/bits/unique_ptr.h:1070
    #1 0x5555557305db in TagLib::ByteVector::ByteVector(TagLib::ByteVector const&) /home/andres/misc/framework/repos/taglib/taglib/toolkit/tbytevector.cpp:345:5
    #2 0x55555595a6e6 in Chunk::Chunk(Chunk const&) /home/andres/misc/framework/repos/taglib/taglib/riff/rifffile.cpp:36:8
    #3 0x55555595a6e6 in TagLib::RIFF::File::updateGlobalSize() /home/andres/misc/framework/repos/taglib/taglib/riff/rifffile.cpp:364:23
    #4 0x55555595d053 in TagLib::RIFF::File::removeChunk(TagLib::ByteVector const&) /home/andres/misc/framework/repos/taglib/taglib/riff/rifffile.cpp:274:7
[...]

PR #1164, opened 2023-11-03
Fix crash with invalid WAV files (#1163)

ufleisch:
With specially crafted WAV files having the "id3 " chunk as the only valid chunk, when trying to write the tags, the existing "id3 " chunk is removed, and then vector::front() is called on the now empty chunks vector.
Now it is checked if the vector is empty to avoid the crash.


CarlosAndresRamirez:
Thank you @ufleisch for your promptly response to this issue. Much appreciated.

PR #1172, opened 2023-11-06
Use ID3v2::FrameFactory also for WAV and DSDIFF

ufleisch:


taglib/riff/wav/wavfile.h
```diff
@@ -86,9 +90,13 @@ namespace TagLib {
          * responsible for deleting it after the File object.
          *
          * \note In the current implementation, \a propertiesStyle is ignored.
+         *
+         * If this file contains and ID3v2 tag the frames will be created using
```
hyperquantum, line 94:
```suggestion
         * If this file contains an ID3v2 tag the frames will be created using
```

Issue #1185, opened 2023-12-06
warning C4251 due to missing dll-interface in TagLib::FileName


daschuer:
With MSVC 14.29.30133 (Visual Studio 2019) I get the following warning. 

```
Warning: C:\buildenv\mixxx-deps-rel-2.4-x64-windows-3e909e2\installed\x64-windows\include\taglib\tiostream.h(53): warning C4251: 'TagLib::FileName::m_name': class 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>' needs to have dll-interface to be used by clients of class 'TagLib::FileName'
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\include\xstring(4871): note: see declaration of 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>'
Warning: C:\buildenv\mixxx-deps-rel-2.4-x64-windows-3e909e2\installed\x64-windows\include\taglib\tiostream.h(54): warning C4251: 'TagLib::FileName::m_wname': class 'std::basic_string<wchar_t,std::char_traits<wchar_t>,std::allocator<wchar_t>>' needs to have dll-interface to be used by clients of class 'TagLib::FileName'
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\include\xstring(4872): note: see declaration of 'std::basic_string<wchar_t,std::char_traits<wchar_t>,std::allocator<wchar_t>>'
Warning: C:\buildenv\mixxx-deps-rel-2.4-x64-windows-3e909e2\installed\x64-windows\include\taglib\mp4atom.h(88): warning C4251: 'TagLib::MP4::Atom::children': class 'TagLib::List<TagLib::MP4::Atom *>' needs to have dll-interface to be used by clients of class 'TagLib::MP4::Atom'
C:\buildenv\mixxx-deps-rel-2.4-x64-windows-3e909e2\installed\x64-windows\include\taglib\mp4atom.h(41): note: see declaration of 'TagLib::List<TagLib::MP4::Atom *>'
[...]

PR #1188, opened 2023-12-09
Fix warnings issued by MSVC (#1185)

ufleisch:
I do not like the many `TAGLIB_EXPORT` line introduced here either, but it is probably still better than forcing MSVC users to disable warnings. An alternative would be not to use `std::unique_ptr`.

Unfortunately, MSVC exports everything (not only public members) when `__declspec(dllexport)` is set at the class level via `TAGLIB_EXPORT`, which leads to many "needs to have dll-interface to be used by clients" 4251 warnings issued by MSVC, because the `std::unique_ptr` pimpls are exported too. As it is not possible to "unexport" private members, all public and protected class member now have to be exported. The MP4 atom classes have to be refactored to use private implementation classes and a few other warnings had to be fixed. Now MSVC compiles the code without warnings, and there is only a single suppression in _tpropertymap.cpp_ because the case for exported template instantiations does not seem to be solvable.

Also remove deprecated functions:
- `APE::Item::value()`: use `binaryData()`
- `APE::Item::toStringList()`: use `values()`
- `MPEG::ID3v2::Frame::Header::unsycronisation()`: use `unsynchronisation()`


sbooth:
I'm not a user of MSVC so I can't really test this idea, but would creating a `.DEF` file be an alternative to these changes? I don't think the public API changes often enough that a `.DEF` file would be unmanageable to keep updated once created.

ufleisch:
I only use this compiler when I have to. I am therefore reluctant to deface all headers or manage .DEF files because of it. I have now cleaned up the changes into separate commits and reduced the changes to keep exporting the whole class and explicitly suppress the MSVC warning for STL private members, this should be OK since they are not accessible for clients. I have only switched to exporting individual members where it is needed, i.e. for classes deriving from templates (`StringList, ByteVectorList, PropertyMap`).

sbooth:
Too bad not supporting MSVC isn't an option 😆

sbooth:
In all seriousness though I think these changes are an excellent improvement.

neheb (approved):

neheb:
wonder if clang on Windows behaves differently.

ufleisch:
> wonder if clang on Windows behaves differently.

I have tested it with MSYS Clang, it builds without warnings.

neheb:
> > wonder if clang on Windows behaves differently.
> 
> I have tested it with MSYS Clang, it builds without warnings.

I meant clang-cl or Clang on Windows, not MSYS2.

PR #1189, opened 2023-12-17
Fix implicit conversions

sbooth:


neheb:
In regards to warnings:

```
../taglib/asf/asfattribute.cpp: In member function ‘int TagLib::ASF::Attribute::dataSize() const’:
../taglib/asf/asfattribute.cpp:251:5: warning: this statement may fall through [-Wimplicit-fallthrough=]
  251 |     if(d->pictureValue.isValid())
      |     ^~
../taglib/asf/asfattribute.cpp:253:3: note: here
  253 |   case GuidType:
      |   ^~~~
../taglib/asf/asfattribute.cpp: In member function ‘TagLib::ByteVector TagLib::ASF::Attribute::render(const TagLib::String&, int) const’:
../taglib/asf/asfattribute.cpp:290:5: warning: this statement may fall through [-Wimplicit-fallthrough=]
  290 |     if(d->pictureValue.isValid()) {
      |     ^~
../taglib/asf/asfattribute.cpp:294:3: note: here
  294 |   case GuidType:
      |   ^~~~
../taglib/dsdiff/dsdifffile.cpp:69:34: warning: comparison is always false due to limited range of data type [-Wtype-limits]
   69 |       if(name[i] < 32 || name[i] > 127)
../taglib/mpeg/id3v2/id3v2framefactory.cpp(125): warning C4389: '!=': signed/unsigned mismatch
2023-12-17T22:48:26.3616104Z ../taglib/mp4/mp4atom.cpp:305:2: warning: extra ';' [-Wpedantic]
2023-12-17T22:48:26.3616858Z  };
```

taglib/dsdiff/dsdifffile.cpp
```diff
@@ -66,7 +66,7 @@ namespace
       return false;
 
     for(int i = 0; i < 4; i++) {
-      if(name[i] < 32 || name[i] > 127)
+      if(name[i] < 32 /*|| name[i] > 127*/)
```
neheb, line 69: Should this not be 126?

taglib/asf/asfattribute.cpp
```diff
@@ -251,7 +251,7 @@ int ASF::Attribute::dataSize() const
     if(d->pictureValue.isValid()) {
       return d->pictureValue.dataSize();
     }
-    break;
+    return d->byteVectorValue.size();
```
neheb, line 254: why not just [[fallthrough]] ?

taglib/dsdiff/dsdifffile.cpp
sbooth: I had to reference EA IFF 85 to make sure but you're correct.

taglib/asf/asfattribute.cpp
sbooth: I thought it was clearer to duplicate the return than fall through since there isn't a logical relationship between GuidType and BytesType other than the internal storage. But I don't have strong feelings about it.

neheb: It's moot AFAIK. The compiler should be able to optimize it away.

PR #1190, opened 2023-12-17
find_if to find by using operator==

neheb:
maybe?


sbooth:
I'm not sure that this is any clearer than the existing code.

ufleisch:
I think that the existing code is easier to read.

PR #1197, opened 2023-12-20
Remove UTF16LE means swap compatibility hack

ufleisch:


taglib/toolkit/tutils.h
```diff
@@ -203,11 +203,12 @@ namespace TagLib
       /*!
        * Returns the byte order of the system.
        */
-      inline ByteOrder systemByteOrder()
+      inline constexpr ByteOrder systemByteOrder()
```
neheb, line 206: I wonder if inline is needed here...

neheb: oh no it's not. constexpr implies inline.

taglib/toolkit/tutils.h
```diff
@@ -203,11 +203,12 @@ namespace TagLib
...
       {
-        union {
+        union IntCharUnion {
           int  i;
           char c;
+          constexpr IntCharUnion() : i{} {}
```
neheb, line 211: is this even needed? i can be default initialized AFAIK.

taglib/toolkit/tstring.cpp
```diff
@@ -38,7 +38,7 @@ namespace
   using namespace TagLib;
 
   // Returns the native format of std::wstring.
-  String::Type wcharByteOrder()
+  constexpr String::Type wcharByteOrder()
   {
     if(Utils::systemByteOrder() == Utils::LittleEndian)
```
neheb, line 43: can't use if constexpr?

taglib/toolkit/tutils.h
ufleisch: Removed it.

taglib/toolkit/tutils.h
ufleisch: If I use

```
union {
  int  i;
  char c;
} u = { 0 };
```

GCC accepts it, but Clang complains

```
taglib/toolkit/tutils.h:214:12: note: read of member 'c' of union with active member 'i' is not allowed in a constant expression
        if(u.c == 1)
```

taglib/toolkit/tstring.cpp
ufleisch: I will use the ternary operator.

taglib/toolkit/tutils.h
neheb: I meant int I{};

ufleisch:
```
union {
  int  i {};
  char c;
} u;
```

Clang accepts it, but GCC complains

```
error: uninitialized variable ‘u’ in ‘constexpr’ function
```

neheb: What a disaster. OK

neheb (approved):

neheb:
I came up with

```
      constexpr ByteOrder systemByteOrder()
      {
        union IntCharUnion {
          int  i;
          char c;
          constexpr IntCharUnion(int j) : i(j) {}
        } u{1};

        if(u.c == 1)
          return LittleEndian;
        return BigEndian;
      }
```

not sure if better.

taglib/toolkit/tstring.cpp
neheb: I tried this. Looks like it can't:

```
error: accessing ‘TagLib::Utils::{anonymous}::systemByteOrder()::IntCharUnion::c’ member instead of initialized ‘TagLib::Utils::{anonymous}::systemByteOrder()::IntCharUnion::i’ member in constant expression
```

too bad std::endian is not available.

Issue #1196, opened 2023-12-20
Remove deprecated functions


sbooth:
Currently the following functions are marked with doxygen flag `\deprecated`:

- [TagLib::MPEG::File(FileName file, bool readProperties, Properties::ReadStyle propertiesStyle)](https://github.com/taglib/taglib/blob/1ee7493abcb8c96834084f873d5639831fe33b0d/taglib/mpeg/mpegfile.h#L83)
- [TagLib::FLAC::File(FileName file, bool readProperties, Properties::ReadStyle propertiesStyle)](https://github.com/taglib/taglib/blob/1ee7493abcb8c96834084f873d5639831fe33b0d/taglib/flac/flacfile.h#L94)

Should they be removed for 2.0? I'm happy to open a PR with the changes.

ufleisch:
No, unfortunately this "merging with the above constructor" was not possible while keeping an acceptable degree of source compatibility. On the constructors below, I wrote "kept for source compatibility", maybe this should be written also in the comment of these constructors. Code using simple `File(fileName)` would probably work, but as soon as additional arguments are used in existing code, wrong arguments would be passed because C++ would implicitly convert between integrals, booleans and pointers, i.e. it would not even complain, which results in bugs which are hard to find.

sbooth:
That makes sense. Should they be decorated with `TAGLIB_DEPRECATED`?

ufleisch:
I would suggest that we add an `ID3v2::FrameFactory *frameFactory = nullptr` as a last parameter to those constructors in MPEG, FLAC, and TrueAudio which do not yet have a `FrameFactory` parameter, then they are the same as the constructors for DSF, DFF, AIFF, WAV, the types which did not have a `FrameFactory` in TagLib 1. This because `frameFactory` is probably the parameter which is most often omitted. We could then make the other constructors (having `FrameFactory` as the second parameter) deprecated. I will try this suggestion and check if it is really a good idea.

ufleisch:
I have added a PR #1198 with the idea mentioned in my previous comment.

PR #1198, opened 2023-12-20
Unify File constructors with ID3v2::FrameFactory parameter (#1196)

ufleisch:
Make constructors consistent so that the FrameFactory is at the end and optional. Mark the alternative constructors as deprecated.


sbooth:
I think these changes are good, especially for API consistency across file types.

PR #1199, opened 2023-12-22
clang-tidy cleanups mostly

neheb:


taglib/toolkit/tstring.cpp
```diff
@@ -520,12 +520,12 @@ bool String::isAscii() const
 
 String String::number(int n) // static
 {
-  return Utils::formatString("%d", n);
+  return std::to_string(n);
```
sbooth, line 523: Is this locale-independent? For some reason I found the description on [cppreference](https://en.cppreference.com/w/cpp/string/basic_string/to_string) confusing.

ufleisch: I would understand "The results of overloads for integer types do not rely on the current C locale" like this. But it is "not guaranteed by the standard".

PR #1207, opened 2024-01-18
cppcheck + clang-tidy

neheb:


taglib/mp4/mp4itemfactory.h
```diff
@@ -158,7 +158,7 @@ namespace TagLib {
       /*!
        * Destroys the frame factory.
        */
-      virtual ~ItemFactory();
+      ~ItemFactory();
```
ufleisch, line 161: `ItemFactory` is meant to be subclassed and has virtual methods, so the destructor should be virtual too.

taglib/mpeg/id3v2/id3v2framefactory.h
```diff
@@ -137,7 +137,7 @@ namespace TagLib {
       /*!
        * Destroys the frame factory.
        */
-      virtual ~FrameFactory();
+      ~FrameFactory();
```
ufleisch, line 140: `FrameFactory` is meant to be subclassed and has virtual methods, so the destructor should be virtual too.

taglib/riff/rifffile.cpp
```diff
@@ -49,7 +49,7 @@ class RIFF::File::FilePrivate
   {
   }
 
-  const Endianness endianness;
+  Endianness endianness;
```
ufleisch, line 52: Why? It is only initialized in the constructor, so `const` is correct.

neheb:
On Thu, Jan 18, 2024 at 9:18 PM Urs Fleisch ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In taglib/mp4/mp4itemfactory.h
> <https://github.com/taglib/taglib/pull/1207#discussion_r1458351354>:
>
> > @@ -158,7 +158,7 @@ namespace TagLib {
>        /*!
>         * Destroys the frame factory.
>         */
> -      virtual ~ItemFactory();
> +      ~ItemFactory();
>
> ItemFactory is meant to be subclassed and has virtual methods, so the
> destructor should be virtual too.
>
should be public then, no?

> ------------------------------
>
> In taglib/mpeg/id3v2/id3v2framefactory.h
> <https://github.com/taglib/taglib/pull/1207#discussion_r1458352159>:
>
> > @@ -137,7 +137,7 @@ namespace TagLib {
>        /*!
>         * Destroys the frame factory.
>         */
> -      virtual ~FrameFactory();
> +      ~FrameFactory();
>
> FrameFactory is meant to be subclassed and has virtual methods, so the
> destructor should be virtual too.
> ------------------------------
>
> In taglib/riff/rifffile.cpp
> <https://github.com/taglib/taglib/pull/1207#discussion_r1458353168>:
>
> > @@ -49,7 +49,7 @@ class RIFF::File::FilePrivate
>    {
>    }
>
> -  const Endianness endianness;
> +  Endianness endianness;
>
> Why? It is only initialized in the constructor, so const is correct.
>
The warning mostly has to do with const preventing moves. This is a trivial
type but still a good idea to do everywhere.

ufleisch:
Sorry, I was a bit hasty in commenting, I deleted the comments after reading the C++ guidelines. I think that everything is OK. However, I have a few remarks:
- I would rather keep `fileName` as the parameter name for file names and `file` for `File` objects, see my added commit "fixup match function argument names". If you agree, you could use it to fix up your "match function argument names" comment.
- The lines added by the "cppcheck: assign d to a value" will not result in different code because the `d` pimpls in these two classes seem to be a reservation to have the possibility to add some members, but they are actually empty. Adding this code could be deferred to the (probably never happening) moment when these pimpls are filled with members.

neheb:
Right. One of the reasons why I added the assignments is because these functions look like they can be defaulted even though they cannot. The assignments make that clear.

the fix up commit looks fine to me. I did it that way initially to avoid having to modify headers.

PR #1208, opened 2024-01-19
cppcheck: add const to functions

neheb:
Found with functionConst

not sure about this...


ufleisch:
I would not change this. These methods either mutate the state via the `d` pointer or return a non const pointer which could be used to mutate the state, so although they are const from a compiler perspective they do not leave the object unmodified and should therefore not be const.

PR #1209, opened 2024-01-20
More code inspections

ufleisch:

PR #1210, opened 2024-01-20
clang-tidy fixes

neheb:


neheb:
I'll rebase after https://github.com/taglib/taglib/pull/1209 is merged.

neheb:
rebased.

taglib/ape/apeproperties.cpp
```diff
@@ -137,7 +137,7 @@ void APE::Properties::read(File *file, offset_t streamLength)
     analyzeOld(file);
 
   if(d->sampleFrames > 0 && d->sampleRate > 0) {
-    const double length = d->sampleFrames * 1000.0 / d->sampleRate;
+    const auto length = static_cast<double>(d->sampleFrames) * 1000.0 / d->sampleRate;
```
sbooth, line 140: What does the `static_cast` add here? Since `1000.0` is a double `d->sampleFrames` is promoted to a double before multiplication.

neheb: It’s an implicit conversion that -Wconversion warns on. -Wconversion is added by utf8cpp because of…reasons

sbooth: Fair enough. I suppose I don't consider a promotion to double in the same way as an implicit narrowing conversion but it doesn't hurt to make it explicit.

PR #1216, opened 2024-01-27
FileStream: Fix opening long paths on Windows

jonaski:
To make sure paths longer than MAX_PATH (260) can be opened, prefix local paths with `\\?\`, and UNC paths with `\\?\UNC\`.

I've tested on Windows 10 22H2 (Build 19045.3930), even when setting LongPathsEnabled to 1 in the registry, it still won't open files with long paths without prefixing them.

For more information see.: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

### DIFF
--- a/taglib/toolkit/tfilestream.cpp
+++ b/taglib/toolkit/tfilestream.cpp
@@ -49,6 +49,10 @@ namespace
 
   const FileHandle InvalidFileHandle = INVALID_HANDLE_VALUE;
 
+  const std::wstring LongLocalPathPrefix = L"\\\\?\\";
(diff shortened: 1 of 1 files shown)


ufleisch:
Thanks for your contribution. I have formatted and slightly optimized your commit. Is this OK?

neheb:
Is it possible to change the define to use CreateFile2?

ufleisch:
Although I would not advise anyone to use Windows 7, there are still a lot of PCs running Windows 7. I do not see the benefit of `CreateFile2()` which would justify locking out the users of Windows < 8 and the risk of breaking something.

PR #1221, opened 2024-03-24
Fix building with -DBUILD_TESTING=ON -DBUILD_BINDINGS=OFF

ufleisch:


neheb:
so std::filesystem uses std::uintmax_t.

https://en.cppreference.com/w/cpp/filesystem/file_size

Probably simpler to do the same here.

ufleisch:
The problem on Illumos is that the name `offset_t` is already defined. We need a signed type here.

PR #1222, opened 2024-03-28
tag_c set properties use utf-8 encoding, 

shenngw1127:
Hi,

Thanks for your great works! Now i can edit the tag in .dsf .mp3 .flac ... files via Rust call C lib. This lib use unique key for properties. It is ease for usage.

i just modified the tag_c.h and tag_c.cpp. 

1. fixed tag_c set properties use utf-8 encoding.

2. add 2 function taglib_property_keys_strings, taglib_property_get_strings which logic was same as the function without "_strings", but it return value with len (how many strings). It was facilitated decode from char** to multiple strings by Rust language.

also i just modified the Rust interface (taglib-rust) project and i would pull request to. (now just in my repo address: https://github.com/shenngw1127/taglib-rust/tree/dev)

### DIFF
--- a/bindings/c/tag_c.cpp
+++ b/bindings/c/tag_c.cpp
@@ -410,14 +410,14 @@ void _taglib_property_set(TagLib_File *file, const char* prop, const char* value
   if(value) {
     auto property = map.find(prop);
     if(property == map.end()) {
-      map.insert(prop, StringList(value));
+      map.insert(prop, StringList(String(value, String::UTF8)));
     }
     else {
       if(append) {
-        property->second.append(value);
+        property->second.append(String(value, String::UTF8));
       }
       else {
-        property->second = StringList(value);
+        property->second = StringList(String(value, String::UTF8));
       }
     }
   }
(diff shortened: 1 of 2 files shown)


ufleisch:
Thanks for your contribution, you found a bug! The property values in the C bindings are always interpreted as Latin-1 encoded. Your PR always uses UTF-8 encoding, but the encoding should respect the value set via `taglib_set_strings_unicode()`, which is UTF-8 by default. I have fixed this in another PR #1224.

Concerning your other changes: The C bindings consistently use a NULL-terminated array of `char *` C-strings when an array of strings is used. You seem to prefer an array plus a length. The advantage would be that you do not have to iterate over the strings to find out the length, but it is a bit more complicated, and probably, you are interested in all strings anyway and will iterate over them. You also returned an empty string list instead of `NULL` from `taglib_property_get_strings()` in the case of invalid parameters. I think that it is better to be able to differentiate between errors and normal return values. In the Rust bindings, you should probably return an `Error` in a `Result` in this case. I also hope that it is possible to interface with Rust with the `NULL` terminated string lists instead of your `TagLib_Strings` structure, maybe it needs a little conversion in the Rust bindings, but I would prefer not to include this indirection inside the TagLib C-bindings and use PR #1224 instead.

shenngw1127:
Thanks for your reply.
i just tested the PR #1224, that is ok.

Second, i have found the way to decode NULL terminated array in Rust. so i have used the master branch in git repo.
Then i will drop the fork soon.

ufleisch:
Good to hear that. The UTF-8 fix is now included in the 2.0.1 release, so you can use this for the Rust bindings.

Issue #1230, opened 2024-05-14
"Invalid atom type" error with m4a files created by Android


garrybrady:
I've built TagLib 2.0.1 from source on macOS and have been using the examples tagreader / tagwriter.

All works well apart from m4a files created on Android devices. 

I've used the app I develop, a selection of voice recorders and also several different Android devices but all combinations display "Invalid atom type".

I've attached a sample recording.

[android.m4a.zip](https://github.com/taglib/taglib/files/15310532/android.m4a.zip)

garrybrady:
Have tested on both Intel & M1 chips.

sbooth:
The issue is with the `ilst` box which contains a box of type 0x00000001 and size 26. TagLib checks that box types contain printable characters:

https://github.com/taglib/taglib/blob/ebf4c5bbb1ce85a71764122cbc3cc3f556e66586/taglib/mp4/mp4atom.cpp#L100

I believe that the box type *normally* contains printable characters but I haven't yet found a reference *requiring* printable characters. I also believe that unknown box types should be ignored and skipped. 

MP4 is based on the QuickTime file format and [Apple says](https://developer.apple.com/documentation/quicktime-file-format/atoms):

> Type
> A 32-bit integer that contains the type of the atom. This can often be usefully treated as a four-character field with a mnemonic value, such as 'moov' (0x6D6F6F76) for a movie atom, or 'trak' (0x7472616B) for a track atom, but non-ASCII values (such as 0x00000001) are also used. Knowing an atom’s type allows you to interpret its data. An atom’s data can be arranged as any arbitrary collection of fields, tables, or other atoms. The data structure is specific to the atom type. An atom of a given type has a defined data structure. If your application encounters an atom of an unknown type, it should not attempt to interpret the atom’s data. Use the atom’s size field to skip this atom and all of its contents.

I think it probably makes sense for TagLib to skip unknown boxes.

garrybrady:
That's great, thanks. Have tried removing the check for printable chars and all now seems fine on all files.

ufleisch:
Fixed by #1231.

Issue #1077, opened 2023-01-21
Can not detect audio property and tag for a few M4A file format


nathnaeld:
I was working on parsing M4A files. TagLib works well on almost all but it fails on a few M4A files. It does not consider them valid files. I have attached a sample audio.

[audio.zip](https://github.com/taglib/taglib/files/10472790/audio.zip)

ufleisch:
I cannot see a problem with the file *[12] In Real Life.m4a* in the attached archive.

40th:
Looks like some sort of atom problem.  Pic should be a start (.length == 0).

![atom heart](https://40th.com/gfx/gh/one_too_many.png)

ufleisch:
Thanks for the additional information. I now think that I can figure out why you see a problem and I do not. If I look at the same place (_mp4file.cpp_, l. 39pp)  with your example file in the debugger, I see `length=3342441612` whereas you see `length=0`. The reason is that I am on Linux (amd64 architecture) where `long` (the type of `length`) has 64 bits whereas on Windows (I suppose you are on Windows) it has only 32 bits. The length of 3342441612 is too large for a signed 32 value (`INT_MAX` is 2147483647).
Unfortunately, changing the type of `length` to a 64-bit value would break binary compatibility, so we cannot do this for TagLib 1.x, we can do this when we start working on TagLib 2.x where we can break binary compatibility. There are a lot of other places which are affected by `long` having only 32 bits on Windows. For my project [Kid3](https://kid3.kde.org), I have patched TagLib for the Windows version, see [taglib-1.13-win00-large_file.patch](https://invent.kde.org/multimedia/kid3/-/blob/master/packaging/patches/taglib-1.13-win00-large_file.patch), so if you are building TagLib yourself, you could apply this patch too.

40th:
So we now agree on the value of the length, but why does a 2:57 AAC track have a length of 3,342,441,612 for anything?

![atom heart](https://40th.com/gfx/gh/one_too_many_2.png)

ufleisch:
I agree that more should be checked. The last "header" `c7 39 9c 8c 92 78 b1 f9` is obviously garbage, so that file should be detected as invalid. Probably these header bytes alone should be enough to `setValid(false)`, I will try to find a specification which contains information about valid length and atom name values.

40th:
> I would suggest that we mark such files as invalid.

[...]

PR #1231, opened 2024-05-16
Skip unknown MP4 boxes

sbooth:
I believe that the box type *normally* contains printable characters but I haven't yet found a reference *requiring* printable characters.

MP4 is based on the QuickTime file format and [Apple says](https://developer.apple.com/documentation/quicktime-file-format/atoms):

> Type
> A 32-bit integer that contains the type of the atom. This can often be usefully treated as a four-character field with a mnemonic value, such as 'moov' (0x6D6F6F76) for a movie atom, or 'trak' (0x7472616B) for a track atom, but non-ASCII values (such as 0x00000001) are also used. Knowing an atom’s type allows you to interpret its data. An atom’s data can be arranged as any arbitrary collection of fields, tables, or other atoms. The data structure is specific to the atom type. An atom of a given type has a defined data structure. If your application encounters an atom of an unknown type, it should not attempt to interpret the atom’s data. Use the atom’s size field to skip this atom and all of its contents.

I think it probably makes sense for TagLib to skip unknown boxes.

The relevant standard is [ISO/IEC 14496-12](https://www.iso.org/standard/83102.html) which unfortunately I don't have access to.


sbooth:
I'm not sure if the debug message about skipped atoms is needed.

sbooth:
Semi-related to #1077

sbooth:
Closes #1230

taglib/mp4/mp4atom.cpp
```diff
@@ -131,6 +124,8 @@ MP4::Atom::Atom(File *file)
     }
   }
 
+  debug("MP4: Skipping atom '" + d->name + "'");
+
```
ufleisch, line 128: This will print a lot of atom names while parsing and possibly disturb the output because `d->name` can be anything. Therefore I would remove this.

sbooth: You're right, even in my testing it was a bit annoying

Issue #1239, opened 2024-07-26
Setting arbitrary tags for MP4/M4A?


sentriz:
hi there,

I'm wonder if setting any arbitrary tag is supported for mp4. the usecase is setting non-standard tag names like `albumartists_credit` or `albumartist_credit`. which works for flac, but seems like for mp4 the tag is [ignored](https://github.com/taglib/taglib/blame/c4ed5900/taglib/mp4/mp4tag.cpp#L495-L501) if it's not present in the [`namePropertyMap`](https://github.com/taglib/taglib/blob/master/taglib/mp4/mp4itemfactory.cpp#L307)

other tagging software that I know of will set tag to be something like `----:com.apple.iTunes:<my tag key>` instead of ignoring it. (picard/puddletag/mutagen) 

do you know if this is possible with taglib? 

thanks!

ufleisch:
Thanks for the report. I have fixed it with PR #1240, please test it.

sentriz:
amazing thanks!

PR #1240, opened 2024-07-27
Support free form tags with MP4 properties (#1239)

ufleisch:


sentriz:
can confirm this works great :+1:

PR #1241, opened 2024-07-30
Remove 'using namespace std' to avoid potential conflicts in tagreader.cpp example

nekiwo:
Although there are no particular problems with global scope in this example, it might cause issues in the future or when users try to copy-paste code from it directly.


ufleisch:
For reasons of consistency, _tagwriter.cpp_ and _framelist.cpp_ from the _examples_ folder should then undergo the same treatment.

Issue #1249, opened 2024-11-02
Pictures larger with 16 MB or more corrupt FLAC files


ufleisch:
The size of FLAC metadata blocks is stored in only 24 bits. If a block exceeds a size of $2^{24}$ bytes = 16 MBytes (e.g. because it contains a large picture), an incorrect size is stored and the file is corrupted. This affects both FLAC and Ogg FLAC files.

PR #1250, opened 2024-11-02
Do not store too large FLAC metadata blocks (#1249)

ufleisch:
The size of FLAC metadata blocks is stored in only 24 bits. Remove blocks exceeding this limit when saving FLAC and Ogg FLAC files.

### DIFF
--- a/taglib/flac/flacfile.cpp
+++ b/taglib/flac/flacfile.cpp
@@ -265,12 +265,19 @@ bool FLAC::File::save()
   // Render data for the metadata blocks
 
   ByteVector data;
-  for(const auto &block : std::as_const(d->blocks)) {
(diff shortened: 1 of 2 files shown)


scotchi:
I wonder if providing a constant or a "blockOverMaxSize(...)" or something might be useful?

ufleisch:
> I wonder if providing a constant or a "blockOverMaxSize(...)" or something might be useful?

As this issue is caused at a low level (reusing the most significant byte of a 32-bit size so that actually only 24 bits are usable), I think that the fix is more readable if it also operates on this low level.

scotchi:
I don't have a strong opinion on this, but I actually meant for users of the API to check images against so that they wouldn't be silently dropped.

PR #1254, opened 2024-11-21
Fixed warning about shadowing variable

MonkeybreadSoftware:
You can't name parameter and structure field the same as complier frequently complain about this.


MonkeybreadSoftware:
e.g.

```
mp4atom.h:63:29: warning: constructor parameter 'type' shadows the field 'type' of 'AtomData' [-Wshadow-field-in-constructor]
   63 |       AtomData(AtomDataType type, const ByteVector &data) :
      |                             ^
mp4atom.h:65:20: note: previous declaration is here
   65 |       AtomDataType type;

```

PR #1255, opened 2024-11-27
gcc: fix various Wnrvo warnings

neheb:
GCC cannot elide copies with multiple returns.


scotchi:
I feel like this needs significantly more explanation.  This makes the code less readable for optimizations that don't obviously have any benefit?

PR #1257, opened 2024-12-08
Add Shorten (SHN) support

### LINKED ISSUE #1256: Shorten (SHN) support
A format I'd like to see TagLib support is [Shorten](https://en.wikipedia.org/wiki/Shorten_(codec)).

It's an older format and doesn't really have any native tags/metadata so it would be a bit of an odd addition, however it would be great to have audio properties support at the least.

There is precedent for formats without native tags already- module (mod) files.

I'm happy to generate a PR for this if it's likely to be accepted.

sbooth:
Closes #1256

### DIFF
--- a/bindings/c/CMakeLists.txt
+++ b/bindings/c/CMakeLists.txt
@@ -25,6 +25,7 @@ include_directories(
   ${CMAKE_CURRENT_SOURCE_DIR}/../../taglib/ogg/opus
   ${CMAKE_CURRENT_SOURCE_DIR}/../../taglib/dsf
   ${CMAKE_CURRENT_SOURCE_DIR}/../../taglib/dsdiff
+  ${CMAKE_CURRENT_SOURCE_DIR}/../../taglib/shn
 )
 
 set(tag_c_HDRS tag_c.h)
--- a/bindings/c/tag_c.cpp
+++ b/bindings/c/tag_c.cpp
@@ -55,6 +55,7 @@
 #include "opusfile.h"
 #include "dsffile.h"
 #include "dsdifffile.h"
+#include "shnfile.h"
 #include "tag.h"
 #include "id3v2framefactory.h"
 
@@ -189,6 +190,9 @@ TagLib_File *taglib_file_new_type(const char *filename, TagLib_File_Type type)
     break;
   case TagLib_File_DSDIFF:
     file = new DSDIFF::File(filename);
+    break;
+  case TagLib_File_SHN:
+    file = new SHN::File(filename);
     break;
   default:
     break;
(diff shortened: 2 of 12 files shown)


scotchi:
- Probably preferable to use `Shorten` for the namespace name rather than `SHN`.
- The coding style / naming for shnfile is very different from TagLib's style.  Was that code pasted from somewhere else?  There's no copyright statement indicating such.
- String literals obviously can't be in single quotes.

In shnfile there are quite a few things that could be cleaned up a bit (e.g. using TagLib's `ByteVector` instead of arrays that are created with `new`.  Quite a few things could be `constexpr` instead of `const`, etc., but I thought I'd wait to hear where that code's actually from before going into too much detail.

sbooth:
> * Probably preferable to use `Shorten` for the namespace name rather than `SHN`.

I went back and forth on this. I couldn't find a clear standard in TagLib for the naming. File extensions seemed to be common for namespace names- Monkey's Audio uses `APE`, Musepack uses `MPC`, but True Audio uses `TrueAudio`.

> * The coding style / naming for shnfile is very different from TagLib's style.  Was that code pasted from somewhere else?  There's no copyright statement indicating such.

The code is modified from a [Shorten decoder](https://github.com/sbooth/SFBAudioEngine/blob/main/Sources/CSFBAudioEngine/Decoders/SFBShortenDecoder.mm) I wrote in Objective-C++ a few years ago. The naming conventions are a bit of a mishmash as I've ported portions back and forth to different languages several times including this pared-down port to C++. I was generally inspired by the original shorten written by [Tony Robinson](https://tonyrobinson.com/publications) as the reference for the Shorten format.

> * String literals obviously can't be in single quotes.

Those are four-character codes (integers masquerading), not string literals, but they can easily be changed to integers. 

> In shnfile there are quite a few things that could be cleaned up a bit (e.g. using TagLib's `ByteVector` instead of arrays that are created with `new`. Quite a few things could be `constexpr` instead of `const`, etc., but I thought I'd wait to hear where that code's actually from before going into too much detail.

ufleisch:
It looks good to me. I just added a few commit with stylistic changes, you can still remove them if you do not agree.

sbooth:
Thanks for the review- I am fine with those changes. I pushed another commit with some additional explicit types in an attempt to further standardize and group constants.

What is your opinion on `SHN` vs. `Shorten` for the namespace name?

ufleisch:
> What is your opinion on `SHN` vs. `Shorten` for the namespace name?

Probably, `Shorten` would be more consistent with most (but not all) of the other namespace names.

sbooth:
If there are no further review comments or objections in the next few days I will squash and merge this into `master`.

PR #1258, opened 2024-12-08
Use fixed width types

sbooth:
Most 64-bit Linux systems and macOS use LP64 with 64-bit `long` and pointers. Windows is slightly different and uses LLP64. 

This means that on LP64 systems `long` and `long long` are the same size.

As usual cppreference.com has a good page on [Fundamental Types](https://en.cppreference.com/w/cpp/language/types).

The current definition of the `byteSwap` functions use `unsigned short`, `unsigned int` and `unsigned long long` for 16-, 32-, and 64-bit uints.

The definitions I've seen for the various builtin `bswap` functions are defined using fixed with types.  For example [gcc](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html) has the folllwing:

> Built-in Function: uint16_t __builtin_bswap16 (uint16_t x)
> Returns x with the order of the bytes reversed; for example, 0xaabb becomes 0xbbaa. Byte here always means exactly 8 bits.
> 
> Built-in Function: uint32_t __builtin_bswap32 (uint32_t x)
> Similar to __builtin_bswap16, except the argument and return types are 32-bit.
> 
> Built-in Function: uint64_t __builtin_bswap64 (uint64_t x)
> Similar to __builtin_bswap32, except the argument and return types are 64-bit.

And macOS uses something similar for the `OSSwap` functions.

I think it would be great if TagLib followed suit.

### DIFF
--- a/taglib/toolkit/tbytevector.cpp
+++ b/taglib/toolkit/tbytevector.cpp
@@ -140,14 +140,13 @@ T toNumber(const ByteVector &v, size_t offset, bool mostSignificantByteFirst)
   ::memcpy(&tmp, v.data() + offset, sizeof(T));
 
   if(swap) {
+    static_assert(sizeof(T) == 2 || sizeof(T) == 4 || sizeof(T) == 8, "Byte swap requested for type with invalid size");
     if constexpr (sizeof(T) == 2)
       return Utils::byteSwap(static_cast<uint16_t>(tmp));
     else if constexpr (sizeof(T) == 4)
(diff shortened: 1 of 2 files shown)


sbooth:
Apparently I need to install Ubuntu for testing because these changes compile fine for me on macOS with clang. This might be a more wide-ranging PR than I initially suspected.

ufleisch:
Just rebased, looks good to me.

Issue #1262, opened 2024-12-23
Allow format support to be configured at compile time


OxygenCobalt:
My app's use case only reasonably expects a few file formats supported by the playback infrastructure, but since I more or less have to use FileRef for format inference, a lot of unused formats are (AFAIK) linked into my app, which winds up bloating it's size. Some way to dynamically enable or disable formats as a cmake configuration would be nice as a result. Is this already present? Is it reasonable? Thanks.

ufleisch:
Please have a look at #1263.

PR #1263, opened 2024-12-28
Compile time configuration of supported formats (#1262)

ufleisch:
I have added CMake options `WITH_APE`, `WITH_ASF`, `WITH_DSF`, `WITH_MOD`, `WITH_MP4`, `WITH_RIFF`, `WITH_TRUEAUDIO`, `WITH_VORBIS`, by default, they are all `ON`.

When all these options are set to `OFF` the size of a release build shared library on Debian GNU/Linux is reduced from 1.6M to 720K and the library only supports MPEG files with ID3 tags. I first thought about making ID3 also configurable, but `ifdef`ing all the ID3 frame factories out of the various constructors and private implementation classes would make the code even more ugly than it gets with the currently used `#ifdef`s, and the library with all options `OFF` would be useless.

As mentioned above, the code does not get prettier. It could be cleaned up a bit by merging the same conditions in the `CMakeList.txt` files and the `#include` sections, but I left it like this in order to generate exactly the same shared library in release mode as before. I could pretty it up a bit if such a change is acceptable.

Besides comparing the binaries before and after the changes, I used the following tests, which check the "one-hot" configurations rather than every possible combination.

```
#!/bin/sh
set -e
ALL_CFGS="APE ASF DSF MOD MP4 TRUEAUDIO VORBIS RIFF"
for cfg in $ALL_CFGS; do
  cmake -DWITH_$cfg=ON .
done
echo '### All ON'
ninja check
for cfg in $ALL_CFGS; do
  cmake -DWITH_$cfg=OFF .
  echo "### $cfg OFF"
  ninja check
  cmake -DWITH_$cfg=ON .
done

for cfg in $ALL_CFGS; do
  cmake -DWITH_$cfg=OFF .
done
echo '### All OFF'
ninja check
for cfg in $ALL_CFGS; do
  cmake -DWITH_$cfg=ON .
  echo "### $cfg ON"
  ninja check
  cmake -DWITH_$cfg=OFF .
done

for cfg in $ALL_CFGS; do
  cmake -DWITH_$cfg=ON .
  echo "### Keeping $cfg ON"
  ninja check
done
```

Any opinions?


OxygenCobalt:
This does exactly what I need @ufleisch, thank you so much. Looks good to me overall (albeit I'm not fully knowledgeable about taglib's codebase).

Tolriq:
Might require a new option for SHN.

ufleisch:
> Might require a new option for SHN.

Is already done, see for example https://github.com/taglib/taglib/blob/97e72e179ce8ab392a39e2defc2a6e06571a47c9/CMakeLists.txt#L106

OxygenCobalt:
Updates on when this would be merged/released @ufleisch ?

ufleisch:
> Updates on when this would be merged/released @ufleisch ?

I was away on vacation the whole month, but now I suppose it is finished, and I will merge it this weekend if nobody objects. The release will be later, maybe in spring.

OxygenCobalt:
Cool, I'll likely checkout with this patch on my branch once merged @ufleisch, so take your time w/the release.

Issue #1268, opened 2025-04-02
Adaptation Proposal of [TagLib] for  OpenHarmony Operating System


dearysp:
Background
OpenHarmony is an open - source project incubated and operated by the OpenAtom Foundation. Its goal is to build an operating system framework and platform for intelligent terminal devices in the era of full - scenario, full - connectivity, and full - intelligence.

TagLib, as a powerful open-source audio metadata library, has had a profound impact on the multimedia software ecosystem.

We will submit the adaptation code through a pull request (PR) as soon as possible. We created this issue to have in - depth discussions with the community about merging the adaptation code. We sincerely welcome your valuable suggestions.

Adaptation Plan
We have completed the adaptation of TagLib on OpenHarmony. The specific content is as follows:

Merge into the main community.
Daily maintenance by the community.
Test Results
Large - scale compatibility tests have been completed, covering different versions of the HarmonyOS system and various types of Harmony devices. The test results show that the modified version is highly compatible with the HarmonyOS system, and there are no obvious stability issues.

Benefits for the Upstream Community
It can reach the large user group and developer community of OpenHarmony.
By supporting the trend of cutting - edge operating system development, it ensures the long - term development potential of the project.
It enhances the project's processing ability in diverse operating environments.

Kyant0:
The library is cross platform, so it's not an issue.

dearysp:
[srt_oh_pkg.patch](https://github.com/user-attachments/files/19667094/srt_oh_pkg.patch)

Sorry for being wordy.

We've modified the tests/utils.h to add support for OpenHarmony OS. Please review the patch.

If there are no issues, I will submit the PR

dearysp:
[taglib_oh_test.patch](https://github.com/user-attachments/files/19667114/taglib_oh_test.patch)

Sorry for being wordy.

We've modified the tests/utils.h to add support for OpenHarmony OS. Please review the patch.

If there are no issues, I will submit the PR

sbooth:
Am I correct that the issue for your integration is the lack of `P_tmpdir` on OpenHarmony OS?

dearysp:
​In the C standard library, P_tmpdir is  set to /tmp
 However, in OpenHarmony OS, temporary files are stored in the /data/local/tmp directory, which differs from the Linux convention.

Kyant0:
Why not use ifndef? Won't the hard coded value break Linux?

ufleisch:
[...]

PR #1270, opened 2025-04-14
Support custom temp and tests directories (#1268)

ufleisch:
The following user-settable values for CMake are supported:

- TESTS_DIR: Tests directory, is path to unit test data when 'data' is appended. Can be used to run the unit tests on a target.
- TESTS_TMPDIR: Directory for temporary files created during unit tests, system tmpdir is used if undefined. Has to be defined on systems without global temporary directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,10 @@ set(TAGLIB_INSTALL_SUFFIX "" CACHE STRING
   "Suffix added to installed files (include directory, libraries, .pc)")
 
 add_definitions(-DHAVE_CONFIG_H)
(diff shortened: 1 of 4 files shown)


dabing518:
What is the progress of the OpenHarmony Os adaptation? When will it be officially released? Thank you!

ufleisch:
> What is the progress of the OpenHarmony Os adaptation? When will it be officially released? Thank you!

Version 2.1 will be released on May 31.
https://mail.kde.org/pipermail/taglib-devel/2025-May/003096.html

dabing518:
您好，邮件我已收到，谢谢！

Issue #1245, opened 2024-10-15
[Feature Request][MP4] Support for arbitrary atom access


acolombier:
It would be great if taglib could allow accessing MP4 atoms directly, including fetching, adding, updating and deleting.

I currently reused some of the private implementation of the library to do so as [part of personal project](https://github.com/acolombier/stemgen/blob/main/stembox/stembox_api.cpp), but if you would be interested, I'd like to make that part of the TagLib public library. This would help the Mixxx project to [edit NI Stem track metadata](https://github.com/mixxxdj/mixxx/issues/13710)

I appreciate this would expose quite a bit of the private library as I have currently implemented, so I'd be happy to rework that to some extend to make it manageable over time in the API.

ufleisch:
This sounds interesting. TagLib already has some extension points, the `TagLib::MP4::ItemFactory` if you simply want to support additional atoms inside `moov.udta.meta.ilst`. Your use case goes beyond this, so there would be the possibility to extend `TagLib::File::MP4` (and the dependencies like the corresponding `Tag` as needed) and register it with `TagLib::FileRef::addFileTypeResolver()`. This enhanced MP4 resolver would then be used in your application instead of the default implementation. If this is not possible or involves copying or duplicating a lot of code from TagLib because it is not accessible, a pull request fixing that would be welcome. Note, however, that binary compatibility (see e.g. [Policies/Binary Compatibility Issues With C++](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B) must be preserved.

acolombier:
FYI, I am working on [an alternative to add first class support to STEM](https://github.com/acolombier/taglib/pull/1) directly within TagLib.

The reason for this change in direction is the desire to use this in the Rust wrapper, that only support the C TagLib.

Appreciate this may not be a direction you would like to take tho so I have keep the PR on my repo so far.

ufleisch:
Sorry for not answering earlier. I do not know anything about STEM, so I am probably not a great help here. Concerning `TagLib::MP4::Tag::save()`, making it virtual would probably break binary compatibility on Windows. But it is apparently only called from `MP4::File::save()`, which is virtual, so you could add a new non-virtual method to `MP4::Tag` and reimplement `MP4::File` overriding `save()` and using the reimplemented `MP4::Tag` class. Exposing `d->atoms->path()` should be possible.

I just had a short look into your fork. Exposing new functionality in the C bindings with complex properties is the correct thing to do.

[...]

PR #1299, opened 2026-01-20
MP4: Add support for NI STEM

acolombier:
Following up on #1245 and attempting to address it.

I have been using this patch for a little while on the project I referenced in the feature request issue and it has been working well, alongside other MP4 tag editing, including complex property such as cover (COVR).

I would be very happy to submit that to taglib, as it would allow Mixxx (one of the many user of taglib) to provide first class support to stem, via TagLib, which is already the foundation library for tag editing.

Let me know if there would be a need to prepare or provide any further work to make this feature in a state it could be considered for adoption.

### DIFF
--- a/taglib/mp4/mp4stem.h
+++ b/taglib/mp4/mp4stem.h
@@ -26,57 +26,53 @@
 #ifndef TAGLIB_MP4STEM_H
 #define TAGLIB_MP4STEM_H
 
-#include "tlist.h"
 #include "tbytevector.h"
 #include "taglib_export.h"
-#include "mp4atom.h"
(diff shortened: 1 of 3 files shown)


acolombier:
Not sure I understand this comment, but just to make this clear, I wrote this code myself entirely. I have pushed the squashed version, but happy to push the many commit with back and forth changes, to prove the iteration before I got something working.

Of course, this doesn't say we might have a very close implementation result, but just wanted to make it clear this is not stolen IP! I like the idea of asking them to contribute back, tho not sure they would immediately comply.

PR #1301, opened 2026-01-24
Verify the ID3v2 version and revision are not 0xFF

sbooth:


sbooth:
That seems like a much better approach. I will revert the file length checks from this PR and keep the changes focused on the ID3v2 version and revision.

sbooth:
@ufleisch This is 9 commits more than necessary but I think it's ready for review.

I don't know the reason why the version and revision weren't checked before; maybe some tags that would otherwise be valid would fail? I think that since the spec is clear about this that it's a good check to add.

PR #1302, opened 2026-01-29
Add check for ID3v2 frame data length

sbooth:


sbooth:
@ufleisch is the original author of this change

PR #1303, opened 2026-02-03
Add DSD information to WavPack properties

damien78:
WavPack files can contain DSD or PCM stream.
Exposing this information (isDSD) is useful to recompute the actual DSD rate (e.g. DSD64 when the reported sample rate is 352800 and bit depth 8bit)


sbooth:
Thank you for this PR.

Should the method be named `isDSD` since DSD is an acronym?

PR #1305, opened 2026-02-09
Improve detection of files with incorrect extensions

lepre-zhang:
TagLib incorrectly identifies files when the extension doesn't match the content.
For example, an AAC-in-MP4 file renamed to `.mp3` gets parsed with the MPEG parser,
returning invalid/empty metadata rather than being correctly identified as MP4 (with `detectByContent()`).

This happens as many parsers don't call `setValid(false)` when parsing incompatible formats.
They return a "valid" File object with garbage data.
Content-based detection never runs because `isValid()` returns true.

After constructing a File based on extension, we should verify it with `isSupported()`.
If the content doesn't match the extension, we attempt content-based detection.
And fall back to the original parser if content detection also fails
(handles partially broken files that might still be parsable).

I have tested this locally on a folder containing 230 files (~1.5GiB)
and there was no measurable performance drop.

### DIFF
--- a/taglib/fileref.cpp
+++ b/taglib/fileref.cpp
@@ -95,6 +95,10 @@ namespace
 {
   List<const FileRef::FileTypeResolver *> fileTypeResolvers;
 
+  // Forward declaration so detectByExtension() can call detectByContent().
+  File *detectByContent(IOStream *stream, bool readAudioProperties,
+                        AudioProperties::ReadStyle audioPropertiesStyle);
+
   // Detect the file type by user-defined resolvers.
 
   File *detectByResolvers(FileName fileName, bool readAudioProperties,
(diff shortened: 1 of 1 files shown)


taglib/fileref.cpp
```diff
@@ -95,6 +95,10 @@ namespace
 {
   List<const FileRef::FileTypeResolver *> fileTypeResolvers;
 
+  // Forward declaration so detectByExtension() can call detectByContent().
+  File *detectByContent(IOStream *stream, bool readAudioProperties,
+                        AudioProperties::ReadStyle audioPropertiesStyle);
```
MinnieTheMoocher: yikes! why not #include the proper header declaring it??

PR #1334, opened 2026-04-06
Fix unbounded recursion in EBML/Matroska MasterElement and MP4 atoms (#1326)

ufleisch:
Credits for reporting go to https://github.com/ericliu-12.


ufleisch:
Thanks for the fast response! Yes, it would be great if you could test it before the merge.

PR #1336, opened 2026-04-10
Matroska: Use seek head for faster element lookup (#1321)

ufleisch:
Reading Matroksa files on slow network drives took quite a while because the whole file was scanned for elements. Now, if a seek-head is found, scanning is stopped and only the elements referenced in the seek-head are read.

The time for reading a 781 MByte Matroska-Video (DVD grabbed using HandBrake) over a slow SMB connection (Raspberry Pi over WLAN) is now reduced from 16s to 0.3s.

### DIFF
--- a/examples/tagreader.cpp
+++ b/examples/tagreader.cpp
@@ -117,4 +117,3 @@ int main(int argc, char *argv[])
   }
   return 0;
 }
-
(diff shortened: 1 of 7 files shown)


HomerJau:
I tested this today in a Windows NET based tagging app using an interop to x64 tag.dll and read performance on a large 11GB MKV has gone from a several seconds to instant. Thanks to all involved!

PR #1341, opened 2026-04-18
Fix DSD/DSF signed integer issues (#1332)

ufleisch:


ufleisch:
@sbooth Could you please have a look at this PR? It tries to solve the issues mentioned in https://github.com/taglib/taglib/issues/1332#issuecomment-4267043732. I had a look at the specifications and the chunk offset and size fields seem to be all unsigned 64-bit integers, so I changed the code so that consistently `unsigned long long` is used for the related values. The tests pass, but as I do not use these audio formats, it would be good if you could make some "real world" tests.

sbooth:
@ufleisch thank you for taking care of this- I think the changes look good. I don't have a large amount of DSF/DFF files but I ran some checks on the ones I do have (which are normal, valid files) and everything worked as expected. I don't have any corrupt/fuzzed files to test with.

PR #1325, opened 2026-03-20
Feature/mp4 chapterlist

ryanfrancesconi:
Hi, 
Trying this PR again... Full disclaimer: I created this mostly with Claude Code as this was really a headache. It took a lot to get it right but I think it's good.

Two main things here:

1. Add Nero-style chapter marker support for MP4
    - Implement read/write/remove of Nero-style chapter markers (chpl atom) in MP4 files.  
    - The chpl atom lives at moov/udta/chpl, storing up to 255 chapter entries with 100-nanosecond timestamps and UTF-8 titles.
    - Includes CppUnit tests covering round-trip read/write, remove, tag preservation, and reading from files with no chapters.

2. Add QuickTime-style chapter track support for MP4
   - The existing MP4ChapterList handles Nero-style chapters (chpl atom),
which store chapter markers in moov/udta/chpl. While functional, the
Nero format has limited support outside of a few players.

    - QuickTime-style chapter tracks are the native chapter format for
Apple's ecosystem. They use a disabled text track (hdlr type "text")
referenced by a chap track-reference in the audio track's tref box. It is also the format that
AVFoundation reads natively via AVAssetChapterMetadataGroup.

    - The implementation produces output that matches ffmpeg's chapter track
structure byte-for-byte: per-sample stts entries (required by
AVFoundation), encd atoms for UTF-8 text encoding, edts/elst edit
lists, gmhd with gmin+text media information, and disabled tkhd
flags (track_in_movie only). These details were verified against
ffmpeg reference output to ensure maximum compatibility.

Key behaviors:
- write() inserts tref + chapter trak as a single contiguous block,
  then appends text samples in an mdat atom at EOF
- Handles non-zero first chapter times by prepending a dummy chapter
  at time 0 (stripped on read)
- Overwrite support: removes existing chapter track before writing
- Preserves existing metadata tags and audio data integrity
- Uses timescale=1000 (milliseconds) for chapter track timing

New files:
  taglib/mp4/mp4qtchapterlist.h
  taglib/mp4/mp4qtchapterlist.cpp

Tests: 7 new tests covering write/read round-trip, remove, overwrite,
tag preservation, empty file read, timestamp precision, and non-zero
first chapter handling. All 39 MP4 tests pass.

PR #1343, opened 2026-04-22
Fix/qt chapter orphaned mdat

ryanfrancesconi:
I got a bit out of sync. It was simpler to make a new branch than to resolve the differences in the mp4 feature branch after the upstream changes. 

I believe this is now clean based on the @ufleisch comment here: https://github.com/taglib/taglib/pull/1325

Ryan

⸻

Fix — taglib​/mp4​/mp4qtchapterlist​.cpp

New helper removeQTChapterTrack() (anonymous namespace):

1. Reads stco​[0] from the chapter trak before deleting it — this gives the chapter data offset, so stco​[0] - 8 = start of the mdat atom header.
2. Records the tref length from the audio trak.
3. Removes the chapter trak from moov (in-memory + on-disk), updates parent sizes and chunk offsets.
4. Removes the audio trak's tref.
5. Computes the adjusted mdat offset: both removals are inside moov (which precedes the EOF mdat), so adjusted​Offset = stored​Offset - chapter​Len - tref​Len. Verifies the "mdat" fourcc is present before removing.

Both write() (replace path) and remove() previously had 16 lines of inline removal logic duplicated between them. Both are now a single call to remove​QTChapter​Track().

⸻

Test — tests​/test​_mp4​.cpp

New testQTChapterListNoOrphanedMdat: runs 3 add/remove cycles on no​-tags​.m4a, counts top-level mdat atoms via MP4::​Atoms before and after, and asserts the count is unchanged.


ufleisch:
Thanks for the quick fix, the multiple mdat are gone, however, not only the chapters!
I have an audio book of 484 MB after running it through the `MP4::QtChapterList::write()` function with unchanged chapter data, it is shrinked down to 5.4 MB and the whole audio track is gone. Please tell me if you need more information, but I suppose that it is easy to reproduce. I tested it with another file, and there is also

```
if(Atom *existingChapter = findChapterTrak(file, &atoms, audio.trak)) {
    removeQTChapterTrack(file, &atoms, moov, existingChapter, audio.trak);
```

removing the audio track and afterwards, the metadata and chapters are still there (correct, as nothing has been changed), but the audio track is gone and that file shrank from 712 MB to 8 MB.

ryanfrancesconi:
Ohhhh. Checking.

ryanfrancesconi:
@ufleisch  I've added more tests to fill some gaps found, all pass locally here. I'll hold off pushing anything else into this branch, but it appears to be ok to me. After you confirm on your end with the previous test file, I'll push a few more things up.

Also have tested outside of taglib in my application setup where more application level tests were added as well.

Ryan

ufleisch:
I tested earlier this evening and it looked good, so just go on with your commits.

ufleisch:
I made a few more tests and encountered no problems. I have added a commit to avoid writing the chapters if they are set to the existing value. I could merge this into master if you agree.

ryanfrancesconi:
as in, setting exact values twice, ignore second call? sure.

Issue #1344, opened 2026-04-23
TagLib freezes when analyzing a corrupt file


Managor:
Link to the file here: https://drive.google.com/file/d/1ttPTzrJbEhoCSSJZpg86nhDQHigv9TSl/view?usp=drive_link

And link to a KDE bugreport here: https://bugs.kde.org/show_bug.cgi?id=478809

ufleisch:
The file has a huge number of atoms and is probably malicious. I suggest that we limit the number of atom siblings at the top level to prevent such attacks. I have supplied a PR #1345 for this, please have a look at it.

Managor:
Likely not malicious. I downloaded it from a live stream on youtube long ago. Probably something went wrong in the process

ufleisch:
Fixed in version v2.3.

PR #1345, opened 2026-04-24
Limit MP4 atom sibling count at top level (#1344)

ufleisch:


Managor:
Compiled and tested this. Dolphin no longer freezes.

PR #1347, opened 2026-04-26
[Matroska] Read Performance

HomerJau:
Reading Large Matroska files over a network (SMB & NFS) is extremely very slow.  Minutes for a 50GB mkv file.

* TagLib performs many small reads (often just a few bytes) interspersed with seeks.  Over network filesystems (SMB, NFS) each tiny read becomes a round-trip.

These changes improve performance considerably:
- Increase Fast Mode read buffer to 10MB (from 512Kb in a previous PR)
- Performance Optimisation: Skip Cues + attachment data in Fast mode

### DIFF
--- a/taglib/matroska/ebml/ebmlmkattachments.cpp
+++ b/taglib/matroska/ebml/ebmlmkattachments.cpp
@@ -22,24 +22,55 @@
 #include "ebmlstringelement.h"
 #include "ebmluintelement.h"
 #include "ebmlbinaryelement.h"
+#include "ebmlutils.h"
+#include "tfile.h"
 #include "matroskaattachments.h"
 #include "matroskaattachedfile.h"
 
 using namespace TagLib;
 
-EBML::MkAttachments::MkAttachments(int sizeLength, offset_t dataSize, offset_t offset):
+EBML::MkAttachments::MkAttachments(int sizeLength, offset_t dataSize, offset_t offset) :
   MasterElement(Id::MkAttachments, sizeLength, dataSize, offset)
 {
 }
 
-EBML::MkAttachments::MkAttachments(Id, int sizeLength, offset_t dataSize, offset_t offset):
+EBML::MkAttachments::MkAttachments(Id, int sizeLength, offset_t dataSize, offset_t offset) :
   MasterElement(Id::MkAttachments, sizeLength, dataSize, offset)
 {
 }
 
-EBML::MkAttachments::MkAttachments():
+EBML::MkAttachments::MkAttachments() :
   MasterElement(Id::MkAttachments, 0, 0, 0)
 {
+}
+
+bool EBML::MkAttachments::readMetadataOnly(File& file)
+{
+  const offset_t maxOffset = file.tell() + dataSize;
+  std::unique_ptr<Element> element;
+  while ((element = findNextElement(file, maxOffset))) {
+    if (element->getId() != Id::MkAttachedFile) {
+      element->skipData(file);
+      continue;
+    }
+    // Manually iterate MkAttachedFile children so we can skip the binary
+    // MkAttachedFileData payload, which can be very large (cover art, fonts).
+    auto attachedFile = std::make_unique<MasterElement>(Id::MkAttachedFile);
+    const offset_t childMaxOffset = file.tell() + element->getDataSize();
+    std::unique_ptr<Element> child;
+    while ((child = findNextElement(file, childMaxOffset))) {
+      if (child->getId() == Id::MkAttachedFileData) {
+        // Record size but skip the actual data read
+        child->skipData(file);
+      }
+      else if (!child->read(file)) {
+        return false;
+      }
+      attachedFile->appendElement(std::move(child));
+    }
+    elements.push_back(std::move(attachedFile));
+  }
+  return file.tell() == maxOffset;
 }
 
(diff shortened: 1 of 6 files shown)


ufleisch:
Thanks for your contributions. I have slightly adapted your "orphan chapters PR" and merged it.
Concerning this merge request:

- It seems to have changes from the other merge request, which now cause conflicts. Could you please rebase this branch to master and resolve any conflicts?
- The coding style does not adhere to the style established in TagLib, have a look at the placement of parentheses after `if`, `while`... and the pointer/reference `*` and `&` which should stick at the type. Please also amend this.

HomerJau:
Apologies for the code errors. I was copying and pasting between my test repo and the merge repo and text was being reformatted!

Tolriq:
> Increase Fast Mode read buffer to 10MB (from 512Kb in a previous PR) 

In the PR you change the scan limit and not buffer size. Why ? Fast mode should not read 10MB of data before failing.

ufleisch:
The PR tries to solve performance problems by reducing features in fast mode. If there are attached pictures and fonts, they are probably useful while streaming the content. What worries me more is the omission of the cues, which have to be updated if the offset of their track changes. If they are not read, they cannot be updated and modifying the file might make it invalid.

Rather than reducing the functionality to reduce performance problems, I would like something which tries to solve the problems you have discovered. `File` and `FileRef` take an `IOStream` as a parameter. We could try to create an alternative to `FileStream` implementing the `IOStream` abstract class, which behaves more performant with network filesystems, e.g. using buffering. Maybe using buffered I/O from the standard C-library instead of low level system calls could help.

Tolriq:
I think the main issue is that fastmode expectations, especially for those kind of complex format. (There's the same issue with MP4 and the new chapters that requires reading the full file in some cases, I have not yet checked the code to see if it now always read the full file even if we do not read chapters) is not very clear. I do not think someone should be able to edit a file after reading it via fastmode, but your message seems to imply that it can.

Clients apps may need only a part of the data and may need to get it quickly enough without reading the full files.

Maybe there should be a read only mode with more control over what we actually want to be read ?

HomerJau:
In my testing a larger scan limit improved reading over a network. I have offered to provide large test files.

I tried playing a 48GB tagged 4K concert in Kodi with TagLib 2.2.1 (which reads chapters on start). Over 2 minutes to start (NFS 2.5GBE, Intel NUC 2.5GBE NAS, RAID 5 7200rpm drives).

Performance is fine on local files. Using network storage, no!

Off Topic:
Writing is not usable over a network on files this size. A beta user of my Taggiing app reported over 20min to tag a 25GB file. I’ve reverted to mkvpropedit, write a temp xml, run mkvprop edit, saves same tags in 2 seconds (same file and 1GBE network).

Beta users were reporting corrupted files. That was caused by them quitting the app during saving, not realising the save was still in progress!

This needs detailed performance analysis. Reading is critical to player performance in Kodi. I’ve implemented a Kodi stream buffer which gives acceptable performance (very fast). Same stream buffer in my Win tagging app for reading. I could not find a solution for tag writing that’s acceptable. (Writing I have the workaround, until it’s solved in TagLib). 

I spent way over 100 hrs on TagLib performance. 5 Kodi beta testers involved too. I need to move on. 

I have both audio and video test files I can share, but I need to provide links privately, not on GitHub.

Tolriq:
For the record the 2.2.1 issue was already reported, analyzed and mostly fixed in master. The 512k was added recently after the perf fix, to limit the scan when there's no seek head. That value should not have any impact on the scan perf and if it have, increasing it is not the proper solution.

ufleisch:
@HomerJau @Tolriq There does not seem to be a consent about the "correct" amount of bytes which should be considered in `Properties::ReadStyle::Fast` mode. Currently, `FAST_SCAN_LIMIT` is at 512KB, but @HomerJau would like to have it at 10MB. As far as I can tell, the only thing which matters is that the seek head is not behind this limit, and I have not seen any file where it would not be in the first 512KB. The other case would be files without seek head, but I would not expect good performance for such files anyway. So could you please explain me, why the 512KB are not enough?

Tolriq:
512K is fine, as he explained he think it's buffer size related, probably due to mixing up things during his tests. 10MB scan for fast mode makes no sense.

ufleisch:
@Tolriq Thanks for the feedback. Maybe you can find some time to have a look at the new `WriteStyle` modes of PR #1351 with your use cases.

HomerJau:
I don’t have an issue with Fast Scan limit. It was one of a few changes that seemed to help performance, but I did not test that alone.

I need an official 2.2.2 TagLib release for Kodi 22. My Kodi PR can’t be merged with the 2.2.1 read performance. Kodi 22 is about to go beta, they won’t accept major changes once in beta. @ufleisch can you please look at a 2.2.2 release in next few days please?

ufleisch:
OK, I will make the next release in one week, May 10th. But it will be 2.3, because there are new methods added to the ABI. And I need your feedback for PR #1351, because I would like to include that too.

HomerJau:
Thank you! I’ll also try and find time to look at my writing PR. Maybe the recent Github bug contributed to my issue with code unexpectedly altered between versions. Unsure.

HomerJau:
I've redone this PR against current master (as of today). Closing this and will create a new Matroska Read Performance PR later today.

PR #1348, opened 2026-04-26
[FLAC] Add iXML and BEXT support via APPLICATION blocks

ryanfrancesconi:
Hi,
Continuing the iXML / BEXT support and implementation to FLAC mirroring what was added for WAVE. Minimal tests here and I've also tested at application level via my spfk-* packages:
https://github.com/ryanfrancesconi/spfk-metadata/blob/development/Tests/SPFKMetadataTests/Audio%20File/FLACTests.swift

****

Adds 6 public methods on FLAC::File mirroring RIFF::WAV::File's existing iXML/BEXT API: iXMLData/setiXMLData/hasiXMLData and the BEXT equivalents.

Reads APPLICATION blocks (RFC 9639 § 8.4) carrying either the IANA- registered "riff" foreign-metadata wrapper or the direct "iXML" / "bext" application IDs used by some third-party tools (e.g. Sequoia). Writes the spec-blessed "riff"-wrapped form. Unrecognized application IDs and "riff"-wrapped chunks other than iXML/bext (e.g. "fmt ", "JUNK") flow through unmodified, so existing files round-trip without churn.

Test coverage: read direct + riff-wrapped for both iXML and BEXT, write+reread round-trip, empty-clears-block, and an unknown-application- block preservation guard.

### DIFF
--- a/taglib/flac/flacfile.cpp
+++ b/taglib/flac/flacfile.cpp
@@ -70,6 +70,8 @@ public:
 
   std::unique_ptr<Properties> properties;
   ByteVector xiphCommentData;
+  String iXMLData;
+  ByteVector bextData;
   List<FLAC::MetadataBlock *> blocks;
 
   offset_t flacStart { 0 };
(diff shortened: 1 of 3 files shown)


ryanfrancesconi:
(sorry for the PR bombardment lately)

PR #1349, opened 2026-04-27
[Matroska] Fix: Avoid multi-minute network saves on large Matroska files

HomerJau:
[DRAFT] Fix: Avoid multi-minute network saves for large Matroska files
(I found an issue I'll fix)

When Matroska::File::save() modifies an existing Tags, Chapters, or Attachments element that grows beyond its on-disk size, it currently calls FileStream::insert() to shift every byte after the modified element. For files where these metadata elements appear before the cluster data (the typical layout), this means rewriting potentially gigabytes of cluster data to make room for a few extra bytes of metadata — saves on a 48 GB file over SMB can take several minutes.

This change adopts the strategy used by MKVToolNix mkvpropedit: when a modified element grows beyond its on-disk slot, it is relocated to the end of the file and a Void element of the original size is written at the old location. The SeekHead is updated to reference the new offset, so existing readers find the element at its new location. Cluster data is never moved.

For a 48 GB Matroska file with a 100-byte tag size increase, save time drops from minutes to ~1 second (a few KB written, no large reads or shifts).

API additions:
•	Matroska::SeekHead::updateEntry(ID, offset_t) — update an existing seek-head entry's offset (or add it if absent)
Behaviour changes:
•	Matroska::Element::write() now uses seek + writeBlock when the element offset is at or past the file end (append), avoiding the unnecessary FileStream::insert() round-trip.
•	Matroska::File::save() performs the relocation before writing.
Fallbacks (no behaviour change):
•	Files without a SeekHead — the relocation cannot be advertised, so the existing slow path is used.
•	Slots smaller than EBML::MIN_VOID_ELEMENT_SIZE (2 bytes) — cannot fit a Void marker, fall back to in-place.
•	Same-size or shrunk elements 
Tested on:
*	32 GB MKV over 1 GBE SMB: ~5 min  now ~1.9 s
*      4 GB  MKA over 1 GBE SMB:   ~1 min now ~1.1 s

This is about 20% faster than using mkvpropedit to do same tagging, same files, on same hardware! I had a very happy beta tester earlier today!!

Credit to Claude AI: 
Question: Why do TagLib writes to large Matroska files over an SMB network take > 100x longer than mkvpropedit writes?

### DIFF
--- a/taglib/matroska/matroskaelement.cpp
+++ b/taglib/matroska/matroskaelement.cpp
@@ -163,6 +163,15 @@ offset_t Matroska::Element::sizeRenderedOrWritten() const
 
 void Matroska::Element::write(File &file)
 {
-  file.insert(e->data, e->offset, e->size);
+  // If the element has been relocated to the end of the file (mkvpropedit-
+  // style fast save), use a plain seek + write instead of the slow insert()
+  // path which would shift cluster data.
+  if(e->offset >= file.length()) {
+    file.seek(e->offset);
+    file.writeBlock(e->data);
+  }
+  else {
+    file.insert(e->data, e->offset, e->size);
+  }
   e->size = e->data.size();
 }
--- a/taglib/matroska/matroskafile.cpp
+++ b/taglib/matroska/matroskafile.cpp
@@ -34,6 +34,7 @@
 #include "ebmlstringelement.h"
 #include "ebmluintelement.h"
 #include "ebmlmksegment.h"
+#include "ebmlvoidelement.h"
 #include "tlist.h"
 #include "tdebug.h"
 #include "tagutils.h"
@@ -144,7 +145,7 @@ PropertyMap Matroska::File::setProperties(const PropertyMap &properties)
 
 namespace {
 
-  constexpr offset_t FAST_SCAN_LIMIT = static_cast<offset_t>(512 * 1024);
+  constexpr offset_t FAST_SCAN_LIMIT = static_cast<offset_t>(10 * 1024 * 1024); // 10MB
 
   String keyForAttachedFile(const Matroska::AttachedFile &attachedFile)
   {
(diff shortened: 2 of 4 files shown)


ufleisch:
Thanks for your continued efforts. I had a look at the commit, unfortunately, quite a few things are not working correctly, the Matroska unit tests are not passing, even those which report invalid files, which is a sign that files are corrupted, so I would strongly advise against using this code on real files. The test assertions which check if the original file is restored after multiple file manipulations would be expected to fail when the write mode is changed as described, but all other tests must pass.

I tried to fix the commit, but I came to the conclusion, that it would be better to keep the current, tested behavior - after all it is sufficient for normally sized files on local storage, but provide additional write modes for use cases where files are very large or located on slow file systems. So I introduced a new `WriteStyle` enum, which can be passed to an overload of `Matroska::File::save()`. There are three possible values:

- `Compact`: Write tags, attachments and chapters as compact as possible. This is the default mode.
- `DoNotShrink`: Do not shrink elements; add void padding when content gets smaller. Allow inserts when content gets larger.
- `AvoidInsert`: Like `DoNotShrink` but also avoid inserts for non-last elements: replace a growing non-last element with a void of the old size and append the new element at the end of the segment. For very large files and/or slow (network) filesystems, using this mode will reduce write time significantly.

The first one is the current behavior and the default, the last one is what you specified in this PR, probably the procedure used by `mkvpropedit`. The second is a compromise, which can be used to reserve some space by writing tags (or attachements or chapters) of an expected maximum size, which can then be removed again to be replaced by void element padding. Subsequent writes will then use this padding as long as it is enough.

By the way: The code was mostly generated by AI (Anthropic Claude and Opus), but I gave it clear instructions and made sure that it is sufficiently tested. But this is no replacement for testing it with real files and real use cases. So could you and your beta testers please give it a try and test the new `AvoidInsert` mode thouroughly with your use cases? You can find the code in PR #1351.

HomerJau:
Cancelled. Replaced by updated PR.

PR #1351, opened 2026-05-01
Matroska: Provide different WriteStyle to trade-off size/speed

ufleisch:
A new Matroska::File::save(WriteStyle style) overload is provided to control how tags, attachments and chapters are written to the file.

- Compact: Write tags, attachments and chapters as compact as possible. This is the default mode.
- DoNotShrink: Do not shrink elements; add void padding when content gets smaller. Allow inserts when content gets larger.
- AvoidInsert: Like DoNotShrink but also avoid inserts for non-last elements: replace a growing non-last element with a void of the old size and append the new element at the end of the segment. For very large files and/or slow (network) filesystems, using this mode will reduce write time significantly.

### DIFF
--- a/taglib/CMakeLists.txt
+++ b/taglib/CMakeLists.txt
@@ -244,6 +244,7 @@ if(WITH_MATROSKA)
     matroska/matroskaproperties.h
     matroska/matroskasimpletag.h
     matroska/matroskatag.h
+    matroska/matroskawritestyle.h
   )
   set(tag_PRIVATE_HDRS ${tag_PRIVATE_HDRS}
     matroska/ebml/ebmlbinaryelement.h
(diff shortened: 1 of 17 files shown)


Tolriq:
I unfortunately do not use any edit function to avoid endless complains on file corruptions :)

Here's a run down from Codex 5.5 xhigh on this:

• Risks to consider:

  1. AvoidInsert may have an edge case where the old slot offset is not adjusted. voidAtOffset is captured when the element is moved, but only the element’s normal offset participates in later sizeChanged() adjustments. If an earlier element, such as the SeekHead, grows after that point, write() could overwrite the wrong file
     range with a Void element.
  2. The SeekHead re-render after updateEntry() is only a single pass, not another stabilization loop. If the moved element’s new offset requires a larger EBML encoding for SeekPosition, the SeekHead can grow after the main render loop. That may mark Cues or other listeners dirty, but they are not necessarily re-rendered again
     before writing.
  3. AvoidInsert relies on Cues to identify the last cluster/audio boundary. If a file has no Cues, audioBoundary remains 0, so an element before the media data can still be treated as “last” and grown in place. That seems to weaken the performance guarantee of AvoidInsert for cue-less files.

PR #1352, opened 2026-05-07
Build fix for Haiku's fcntl.h which can't be found in sys/fcntl.h

Begasus:


Begasus:
Got an email on the upcoming update here and if I wanted to check it out, this patch has been living in our repository at [haikuports](https://github.com/haikuports/haikuports), might be a good chance to get this included.

PR #1354, opened 2026-05-08
[Matrsoka] Read Performance over SMB/NFS

HomerJau:
Reading large Matroska files over network filesystems (SMB/NFS) is very slow because TagLib reads megabytes of data that aren't needed for a fast metadata read.

This PR adds two Fast-mode optimizations (AudioProperties::Fast):
1.	Skip the Cues element. Cues are only used by Accurate-mode validation. On large MKV files they can be tens of MB; reading them adds significant latency over network filesystems for no benefit in Fast mode.
2.	Skip the binary payload of attachments. Cover art and fonts are common multi-MB attachments. In Fast mode, MkAttachments::readMetadataOnly() reads the structure of each AttachedFile (filename, MIME, UID, description, size) but skips the MkAttachedFileData payload. The parsed AttachedFile objects expose all metadata; their data() is empty.

Both optimizations are applied in the seek-driven and linear-scan code paths of MkSegment::readLimited(). Accurate mode (scanLimit == dataSize) is unchanged: Cues and attachment payloads are read normally.

Supersedes #1347 (which had unrelated FAST_SCAN_LIMIT).

NOTE: Item 2 could be removed so a tag reader can still get Cover art. Item 1 is the main optimisation needed

### DIFF
--- a/taglib/matroska/ebml/ebmlmkattachments.cpp
+++ b/taglib/matroska/ebml/ebmlmkattachments.cpp
@@ -22,6 +22,8 @@
 #include "ebmlstringelement.h"
 #include "ebmluintelement.h"
 #include "ebmlbinaryelement.h"
+#include "ebmlutils.h"
+#include "tfile.h"
 #include "matroskaattachments.h"
 #include "matroskaattachedfile.h"
 
@@ -40,6 +42,32 @@ EBML::MkAttachments::MkAttachments(Id, int sizeLength, offset_t dataSize, offset
 EBML::MkAttachments::MkAttachments():
   MasterElement(Id::MkAttachments, 0, 0, 0)
 {
+}
+
+bool EBML::MkAttachments::readMetadataOnly(File &file)
+{
+  const offset_t maxOffset = file.tell() + dataSize;
+  std::unique_ptr<Element> element;
+  while((element = findNextElement(file, maxOffset))) {
+    if(element->getId() != Id::MkAttachedFile) {
+      element->skipData(file);
+      continue;
+    }
+    // Manually iterate MkAttachedFile children so we can skip the binary
+    // MkAttachedFileData payload, which can be very large (cover art, fonts).
+    auto attachedFile = std::make_unique<MasterElement>(Id::MkAttachedFile);
+    const offset_t childMaxOffset = file.tell() + element->getDataSize();
+    std::unique_ptr<Element> child;
+    while((child = findNextElement(file, childMaxOffset))) {
+      if(child->getId() == Id::MkAttachedFileData)
+        child->skipData(file);
+      else if(!child->read(file))
+        return false;
+      attachedFile->appendElement(std::move(child));
+    }
+    elements.push_back(std::move(attachedFile));
+  }
+  return file.tell() == maxOffset;
 }
 
 std::unique_ptr<Matroska::Attachments> EBML::MkAttachments::parse() const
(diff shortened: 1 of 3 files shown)


ufleisch:
Thanks for your contribution. I have split the commit into two, the first for skipping the Cues element, the second for skipping the attachment data. I have also made a few adjustments:

The Cues element is only skipped if the file has been opened with `AudioProperties::Fast` mode and in read-only mode, so you will have to use a `FileStream` with `openReadOnly=true`, for example with the _tagreader.cpp_ example:

```
-    TagLib::FileRef f(argv[i]);
+    auto stream = std::make_unique<TagLib::FileStream>(argv[i], true);
+    TagLib::FileRef f(stream.get(), true, TagLib::AudioProperties::Fast);
```

You have to explicitly free the stream when the file is no longer used, that's why I have used a `unique_ptr` in this example. The reason for this restriction is that the Cues have to be updated if the track positions are changed. Many Matroska files have the tags before the Cluster and Cues, and if the tags in such a file are enlarged, and it is read with skipping the Cues and written with the default `WriteStyle::Compact`, the contents of the Cues are not updated and become invalid. Therefore this optimization can only be allowed for files which are opened in read-only mode.

If this is OK for you, I will add the adapted "Skip Cues"-commit before releasing version 2.3. However, I would prefer not to include the other optimization which skips reading the attachment data. I would rather suggest that we try to read that data in a lazy way, i.e. only when the corresponding method is explicitly called. But for the implementation and testing of such a change, we need more time than the remaining day until the release of version 2.3, so I would suggest that we do this later.

HomerJau:
Thanks. I’m ok with your proposal. Good plan.

HomerJau:
@ufleisch Are you locked in for 2.3 release? I’m thinking I’d like an extra day to get these Matroska PRs tested by my 6 beta testers as a sanity/real world test release. 

Can you give me 24hrs once your cleanups are committed please? I’m ok if you need to release as planned, but for my Kodi PR another 24hrs is ok.

ufleisch:
@HomerJau I have now pushed the "Skip Cues"-commit, so the only thing else which I would add for the 2.3 release would be the CHANGELOG.md. So you and your beta testers can take the current master state to have the code which is expected to be in 2.3.

I am not urged to release tomorrow, without your notice, I probably would have released in 12 hours, I have now 19:40h Central European Summer Time, but there is no hurry. It is better to have the release a bit later than to have bugs in it. So take your time, and if you think that it is good to release, just write a message here, and I will release it.

HomerJau:
Thanks Urs!

HomerJau:
Beta testing all ok Urs, although reading was a little slower with current master.

I meant to try increasing the TagLib read stream buffer which is quite small from memory. I use 1MB in the Kodi buffer and that shouldn’t be any issue for devices playing/processing Matroska files, usually fairly large.

Maybe you could increase that please?

But please go ahead with the release.

Thanks for all your work getting Matroska into TagLib!

ufleisch:
Thanks for all your efforts and testing! I have now released the code as it is. I took a look at the `bufferSize()` in `FileStream`. As it is `static`, passing in a buffer size as an additional parameter in the constructor does not seem as obvious at it should be for a last minute fix just before the release. So I think that we should look at it without pressure and improve it in a later release. There is still the possibility to provide a custom `IOStream` implementation, so leaving it like this should be OK for now.

Maybe in the future an advanced `IOStream` could be provided with tweaks for multiple things. I have seen that you use your own `MatroskaTagLibStream` in your [Kodi branches](https://github.com/HomerJau/xbmc/), I have my own [FileIOStream for Kid3](https://invent.kde.org/multimedia/kid3/-/blob/master/src/plugins/taglibmetadata/taglibfileiostream.h) too to reduce the number of open file descriptors.

ufleisch:
The first commit is in v2.3, so I am closing this. For the attachment data, we could create a new PR which tries to fetch the data in a lazy fashion.

PR #1364, opened 2026-05-16
[FLAC] Track iXML/BEXT block presence with explicit flags

### LINKED ISSUE #1362: FLAC `iXML` and `BEXT` data
This issue was found by Copilot. It appears to be legitimate.

@ryanfrancesconi would it make sense to use boolean flags instead of `isEmpty`?

https://github.com/taglib/taglib/blob/1b94b93762636ebe5733180c3e825be4621e4c7f/taglib/flac/flacfile.cpp#L535-L540

These `has*Data()` methods report whether the in-memory payload is non-empty, not whether the file on disk had the metadata as documented. After `setiXMLData()`/`setBEXTData()` on a file that originally lacked those blocks they return true before save, and an existing empty block would return false.

ryanfrancesconi:
Fixes #1362.

`FLAC::File::hasiXMLData()` / `hasBEXTData()` were implemented as
`!data.isEmpty()` checks, which conflated in-memory payload with on-disk
block presence. Two consequences:

* `setiXMLData("foo")` on a file with no iXML block made `hasiXMLData()`
  return `true` immediately, before `save()`.
* A FLAC file carrying an iXML APPLICATION block with empty payload
  round-tripped fine, but `hasiXMLData()` reported `false`.

Switch to the same model `RIFF::WAV::File` already uses: explicit
`hasiXML` / `hasBEXT` bool flags on `FilePrivate`, set during `scan()`
when the APPLICATION block is recognised, updated during `save()` after
the block is (re)written or omitted, and returned verbatim by the
accessors. New regression test (`testHasiXMLAndBEXTReflectFileState`)
pins down the before/after-`save()` and empty-block cases.

## Test plan
- [x] Full local test suite: 556/556 green (`./test_runner`)
- [x] New regression test asserts presence flag stays false between `set*Data()` and `save()`
- [x] New regression test asserts presence flag stays true between `set*Data(String())` and `save()`

### DIFF
--- a/taglib/flac/flacfile.cpp
+++ b/taglib/flac/flacfile.cpp
@@ -72,6 +72,8 @@ public:
   ByteVector xiphCommentData;
   String iXMLData;
   ByteVector bextData;
+  bool hasiXML { false };
+  bool hasBEXT { false };
   List<FLAC::MetadataBlock *> blocks;
 
   offset_t flacStart { 0 };
@@ -279,6 +281,10 @@ bool FLAC::File::save()
     payload.append(ByteVector::fromUInt(xml.size(), false));
     payload.append(xml);
     d->blocks.append(new UnknownMetadataBlock(MetadataBlock::Application, payload));
+    d->hasiXML = true;
+  }
+  else {
+    d->hasiXML = false;
   }
   if(!d->bextData.isEmpty()) {
     ByteVector payload;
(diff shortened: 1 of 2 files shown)


taglib/flac/flacfile.cpp
```diff
@@ -72,6 +72,8 @@ class FLAC::File::FilePrivate
   ByteVector xiphCommentData;
   String iXMLData;
   ByteVector bextData;
+  bool hasiXML { false };
```
ufleisch: I would move these two `bool` fields after `scanned`. This should use one word less per FLAC file because it avoids that padding is introduced.

ufleisch (changes requested):

Issue #1311, opened 2026-02-22
[Matroska] How to read Chapter Tags?


HomerJau:
Thanks for the recent Matroska support in 2.2!!

I can't work out how to read tags attached to Chapters. Can you point me in the right direction?

This is what I see using ffprobe using my test file (only showing first chapter):
Chapters:
    Chapter #0:0: start 0.000000, end 814.814000
      Metadata:
        TRACK           : 1/5
        TITLE           : Shine On You Crazy Diamond, Pts. 1–5
        ARTIST          : Pink Floyd
        MUSICBRAINZ_ARTISTID: 83d91898-7763-47d7-b03b-b92132375c47
    ....

THX

ufleisch:
If you want to attach simple tags to a chapter via the TagChapterUID element (RFC 9559 5.1.8.1.1.5.), this is not yet possible because that element is not written (only `TagLib::Matroska::SimpleTag::trackUid()`). The Matroska files I have seen only use the ChapString element (RFC 9559 5.1.7.1.4.10.), which is supported via `TagLib::Matroska::Chapter::Display::string()`. Do you know of any player which can make use of simple tags attached to chapters?

Matroska supports a lot of features like attaching tags to editions, chapters, and attachments, and all the nesting of tags and chapters, which are not supported by most players and tag editors. The subset of features implemented by TagLib should be enough for most use cases. For new features, it would be helpful to include a list of applications which can use it and instructions where to get example files with those features or how to create such files.

HomerJau:
Thanks for you prompt reply.

The Matroska tagging spec gives this example to better explain my question. Chapters get tagged by adding the ChapterUID to a Tag Target (not to a chapter object):
Tag
Targets
ChapterUID = 123456
TargetTypeValue = 30
TargetType = “TRACK”
SimpleTag
TagName = “TITLE”
TagString = “Baby wants to bleep (pt.1)”
SimpleTag
TagName = “PART_NUMBER”
TagString = “1”
Simple tag
TagName =  “ARTIST”
Tag string = “Unknown”

MKVToolNix (mkvpropedit with an xml file) and MP3Tag can write these chapter tags by using the ChapterID target.

Currently I can’t find a way using TagLib 2.2 to read tags and get their ChapterUID. That’s the missing piece.

My use case is moving Kodi’s Matroska tag reading from using ffmpeg to TagLib. (Writing tags is not critical). 
[...]

PR #1365, opened 2026-05-19
[Matroska] Allow Chapters Without a ChapterUID

HomerJau:
Fix: Don't silently drop ChapterAtom elements that omit MkChapterUID

parseChapterAtom() returned chapterUid = 0 when the source file omitted the MkChapterUID element, and the existing call sites in MkChapters::parse() (ebmlmkchapters.cpp lines 106 and 127) use a C++17 init-if `chapter.uid()` predicate that silently dropped such chapters. The caller saw an empty ChapterEditionList for files that mkvinfo, MediaInfo and FFmpeg all handle gracefully — produced by some audiobook generators and older muxers that omit the per-chapter UID. The companion orphan-EditionEntry fix in PR #1311 / commit e07b956f doesn't cover this case because the ChapterAtoms there ARE wrapped in an EditionEntry, just without a ChapterUID inside each atom.

This change:
- Synthesises a process-unique ChapterUID inside parseChapterAtom() when the source file lacks MkChapterUID. The synthetic value sets the high bit (1ULL << 63) and increments via a static std::atomic counter; real ChapterUIDs are random 64-bit values from muxers, so collision with a generated one is practically impossible while keeping the distinction local to TagLib.
- The existing chapter.uid() filters at the call sites then always evaluate truthy and the chapter is exposed through the public ChapterEditionList API as if it were spec-compliant.

No existing behavior is changed — files that already conform to the spec (ChapterAtoms with a ChapterUID element) parse identically; only previously-dropped chapters are now surfaced.

Reported and verified against real-world chaptered Matroska audiobook files where mkvinfo / MediaInfo see all chapters but TagLib 2.3 returned an empty ChapterEditionList.

### DIFF
--- a/taglib/matroska/ebml/ebmlmkchapters.cpp
+++ b/taglib/matroska/ebml/ebmlmkchapters.cpp
@@ -25,6 +25,7 @@
 
 #include "ebmlmkchapters.h"
 #include "ebmlstringelement.h"
+#include <atomic>
 #include "ebmluintelement.h"
 #include "matroskachapters.h"
 #include "matroskachapteredition.h"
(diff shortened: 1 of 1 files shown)


HomerJau:
Thank Urs. Found the issue in Kodi testing. No rush, I’ve submitted a patch to Kodi that resolves the issue in 2.3. Just include this in your next official release on your schedule.

PR #1369, opened 2026-06-14
Fix XM save misalignment for sample data

EvilGivEn:
This PR fixes an XM save issue in taglib/xm/xmfile.cpp. The save logic advanced past sample headers but did not skip the sample data, which caused subsequent writes to become misaligned. The fix now accumulates and skips the corresponding sampleLength after traversing sample headers. I could not find a CONTRIBUTING file, so I am submitting this PR directly to contribute back to the project. Please advise if this workflow does not match TagLib’s preferred contribution process.

### DIFF
--- a/taglib/xm/xmfile.cpp
+++ b/taglib/xm/xmfile.cpp
@@ -457,6 +457,7 @@ bool XM::File::save()
     }
 
     unsigned long sampleHeaderSize = 0;
+    unsigned long sampleDataSize = 0;
     if(sampleCount > 0) {
       seek(pos + 29);
       if(instrumentHeaderSize < 33U || !readU32L(sampleHeaderSize))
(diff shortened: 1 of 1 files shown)


ufleisch:
Thanks for the PR. Would it be possible that you add a unit test, preferably one which would fail with the old code (i.e. show the misalignment bug) but pass with your fix?

ufleisch:
Using the xm files in the _tests/data/_ folder to reproduce the bug is probably not possible because they do not contain real samples and the changed code path has no effect. Therefore I downloaded some xm files, e.g. liah.xm from https://modarchive.org/ and looked at the with MilkyTracker. Your patch fixed saving of these files.

EvilGivEn:
Thank you for testing the patch with real-world XM files and merging it. Sorry that I didn't manage to add a regression test in time. I appreciate your help validating the fix.

PR #1373, opened 2026-06-24
Fix overflow for large Apple Music cnID values

abbott3344:
Apple Music cnID values now exceed the range of a 32-bit integer. A 64-bit type is required to prevent overflow.


sbooth:
This change makes sense and looks good in general. Do you have a test case for this PR, ideally one that currently fails but passes with the change? As it stands now it doesn't compile because the function name is misspelled (`toLonglong` vs. `toLongLong`).

abbott3344:
> This change makes sense and looks good in general. Do you have a test case for this PR, ideally one that currently fails but passes with the change? As it stands now it doesn't compile because the function name is misspelled (`toLonglong` vs. `toLongLong`).

i'm a newbie coder and this is my first pull request.. i don'.t know what test case means.. do you mean test with real file?

ufleisch:
Thanks for the PR, I have added a unit test.

PR #1374, opened 2026-06-24
Matroska: Ignore some invalid Elements.

anarcobra:
Out of this PR I think the first commit can be used either way since it fixes what I think is a general bug.

The last two commits are in order to still parse tags on mkv files that might have invalid tag information.
Specifically, I have a file for which `mkvinfo` prints the tags, but `tagreader` does not.
There is a broken `Chapter display` tag. See mkvinfo output:

```
|+ Segment information at 4410
| + Segment UID: 0x52 0x54 0x89 0x1f 0xb9 0xc7 0x12 0x72 0xba 0x17 0xc5 0xe8 0x3d 0x54 0x0b 0xf9 at 4415
| + (Known element, but invalid at this position: Chapter display; ID: 0x80 size: 14) at 4434
| + Writing application: HandBrake 0.9.4 at 4448
| + Timestamp scale: 1000000 at 4466
| + Duration: 00:24:57.035250000 at 4473
```

The last two patches are an attempt to make taglib skip over this broken tag.
With these changes the output from `tagreader` now looks like:
```
TagLib: EBML: datasize too great: 8813 > (4448 - 4440)
TagLib: EBML: Invalid MasterElement
-- TAG (basic) --
title   - ""
artist  - ""
album   - ""
year    - "0"
comment - ""
track   - "0"
genre   - ""
CHAPTERS:
  chapters    - [{"displays": [{"string": "Chapter 1"}], "timeStart": 0, "uid": 1759487646}, {"displays": [{"string": "Chapter 2"}], "timeStart": 94027266666, "uid": 1910690585}, {"displays": [{"string": "Chapter 3"}], "timeStart": 734400333333, "uid": 1277432406}, {"displays": [{"string": "Chapter 4"}], "timeStart": 1346178166666, "uid": 67924569}, {"displays": [{"string": "Chapter 5"}], "timeStart": 1466256466666, "uid": 136275841}]
  isDefault   - true
  uid         - 1995116152
-- AUDIO --
bitrate     - 1766
sample rate - 48000
channels    - 2
length      - 24:57
```

### DIFF
--- a/taglib/matroska/ebml/ebmlelement.cpp
+++ b/taglib/matroska/ebml/ebmlelement.cpp
@@ -41,7 +41,7 @@ using namespace TagLib;
 #define RETURN_ELEMENT_FOR_CASE(eid) \
   case (eid): return make_unique_element<eid>(id, sizeLength, dataSize, offset)
 
-std::unique_ptr<EBML::Element> EBML::Element::factory(File &file)
+std::unique_ptr<EBML::Element> EBML::Element::factory(File &file, offset_t maxOffset)
 {
   // Get the element ID
   const offset_t offset = file.tell();
@@ -55,6 +55,13 @@ std::unique_ptr<EBML::Element> EBML::Element::factory(File &file)
   const auto &[sizeLength, dataSize] = readVINT(file);
   if(!sizeLength)
     return nullptr;
+
+  if(const offset_t currentOffset = file.tell();
+     static_cast<offset_t>(dataSize) > maxOffset - currentOffset) {
+    debug(Utils::formatString("EBML: datasize too great: %lu > (%lld - %lld)",
+      dataSize, maxOffset, currentOffset));
+    return nullptr;
+  }
 
   // Return the subclass
   // The enum switch without default will give us a warning if an ID is missing
--- a/taglib/matroska/ebml/ebmlelement.h
+++ b/taglib/matroska/ebml/ebmlelement.h
@@ -122,7 +122,7 @@ namespace TagLib
       int64_t getDataSize() const;
       ByteVector renderId() const;
       virtual ByteVector render();
-      static std::unique_ptr<Element> factory(File &file);
+      static std::unique_ptr<Element> factory(File &file, offset_t maxOffset);
       static unsigned int readId(File &file);
 
     protected:
--- a/taglib/matroska/ebml/ebmlmasterelement.cpp
+++ b/taglib/matroska/ebml/ebmlmasterelement.cpp
@@ -109,16 +109,24 @@ bool EBML::MasterElement::read(File &file, int depth)
   std::unique_ptr<Element> element;
   while((element = findNextElement(file, maxOffset))) {
     if(auto master = dynamic_cast<MasterElement *>(element.get())) {
-      if(!master->read(file, depth + 1))
-        return false;
+      if(!master->read(file, depth + 1)) {
+        debug("EBML: Invalid MasterElement");
+        continue;
+      }
     }
     else {
-      if(!element->read(file))
-        return false;
+      if(!element->read(file)) {
+        debug("EBML: Invalid Element");
+        continue;
+      }
     }
     elements.push_back(std::move(element));
   }
-  return file.tell() == maxOffset;
+  if(file.tell() == maxOffset) {
+    return true;
+  }
+  file.seek(maxOffset);
+  return false;
 }
 
 bool EBML::MasterElement::read(File &file)
--- a/taglib/matroska/ebml/ebmlmksegment.cpp
+++ b/taglib/matroska/ebml/ebmlmksegment.cpp
@@ -45,7 +45,7 @@ std::unique_ptr<ElementType> readElementAt(File &file,
   }
 
   file.seek(offset);
-  auto element = EBML::Element::factory(file);
+  auto element = EBML::Element::factory(file, maxOffset);
   if(!element || element->getId() != Id) {
     return nullptr;
   }
--- a/taglib/matroska/ebml/ebmlutils.cpp
+++ b/taglib/matroska/ebml/ebmlutils.cpp
@@ -34,7 +34,7 @@ std::unique_ptr<EBML::Element> EBML::findElement(File &file, Element::Id id, off
 {
   std::unique_ptr<Element> element;
   while(file.tell() < maxOffset) {
-    element = Element::factory(file);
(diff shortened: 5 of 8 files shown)


ufleisch:
Thanks for the PR, I will have a look at it.

anarcobra:
Great, let me know if you have any questions about what I did.
Also, a bit more information about the broken `Chapter display` tag in my file and why I chose to try to solve it this way.
`MkChapterDisplay`  element is supposed to be a `MasterElement` that then has some more Elements.
However looking at a hexdump of my file it just contains a string. So that string then gets parsed as an Element, which of course results in garbage.

```
00001150  0b f9 80 8c 6c 69 62 6d  6b 76 20 30 2e 36 2e 34  |....libmkv 0.6.4|
00001160  57 41 8f 48 61 6e 64 42  72 61 6b 65 20 30 2e 39  |WA.HandBrake 0.9|
```

In fact it looks to me like it was supposed to have been 0x4D80 `Muxing App`, but some byte got lost somehow.

ufleisch:
Sorry for the delay, I was on vacation for a week. I have made a few changes, I wanted to add them as GitHub suggestions, but unfortunately, this did not suggest the correct changes in all cases, so I comment my pushed changes here.

```diff
diff --git a/taglib/matroska/ebml/ebmlelement.cpp b/taglib/matroska/ebml/ebmlelement.cpp
index 7308cf3b..bde73ff6 100644
--- a/taglib/matroska/ebml/ebmlelement.cpp
+++ b/taglib/matroska/ebml/ebmlelement.cpp
@@ -56,11 +56,11 @@ std::unique_ptr<EBML::Element> EBML::Element::factory(File &file, offset_t maxOf
   if(!sizeLength)
     return nullptr;
 
-  if(static_cast<offset_t>(dataSize) > (maxOffset - file.tell())) {
-    debug("EBML: datasize too great: " + std::to_string(dataSize)
-          + " > (" + std::to_string(maxOffset) + " - "
-          + std::to_string(file.tell()) + ")");
-    return nullptr; 
+  if(const offset_t currentOffset = file.tell();
+     static_cast<offset_t>(dataSize) > maxOffset - currentOffset) {
+    debug(Utils::formatString("EBML: datasize too great: %lu > (%lld - %lld)",
+      dataSize, maxOffset, currentOffset));
+    return nullptr;
   }
 
   // Return the subclass
diff --git a/taglib/matroska/ebml/ebmlmasterelement.cpp b/taglib/matroska/ebml/ebmlmasterelement.cpp
index da6d5cb0..0b38416b 100644
--- a/taglib/matroska/ebml/ebmlmasterelement.cpp
+++ b/taglib/matroska/ebml/ebmlmasterelement.cpp
@@ -122,9 +122,11 @@ bool EBML::MasterElement::read(File &file, int depth)
     }
     elements.push_back(std::move(element));
   }
-  auto tell = file.tell();
-  file.seek(maxOffset, File::Position::Beginning);
-  return tell == maxOffset;
+  if(file.tell() == maxOffset) {
+    return true;
+  }
+  file.seek(maxOffset);
+  return false;
 }
 
 bool EBML::MasterElement::read(File &file)
diff --git a/taglib/matroska/matroskafile.cpp b/taglib/matroska/matroskafile.cpp
index 4d070594..7ff001bd 100644
--- a/taglib/matroska/matroskafile.cpp
+++ b/taglib/matroska/matroskafile.cpp
@@ -381,7 +381,7 @@ void Matroska::File::read(bool readProperties, Properties::ReadStyle readStyle)
     head->skipData(*this);
   }
 
-  offset_t maxOffset = fileLength - tell();
+  offset_t maxOffset = fileLength;
   if (readStyle == Properties::ReadStyle::Fast && maxOffset > FAST_SCAN_LIMIT) {
     maxOffset = FAST_SCAN_LIMIT;
   }
@@ -389,7 +389,7 @@ void Matroska::File::read(bool readProperties, Properties::ReadStyle readStyle)
   // Find the Matroska segment in the file
   const std::unique_ptr<EBML::MkSegment> segment(
     EBML::element_cast<EBML::Element::Id::MkSegment>(
-      EBML::findElement(*this, EBML::Element::Id::MkSegment, maxOffset + tell())
+      EBML::findElement(*this, EBML::Element::Id::MkSegment, maxOffset)
     )
   );
   if(!segment) {
```

- The `debug()` output is more consistent with the existing code and one less `tell()` call is executed.
- The `seek()` is only executed in the less likely error case.
- The `maxOffset` in the "non fast scan" case was probably wrong, it should be absolute. I have not removed the `- tell()` so it is no longer necessary to `+ tell()` below, which would be wrong in the "fast scan" case.

anarcobra:
> Sorry for the delay, I was on vacation for a week. 

No worries, and thanks for the rework.

~~The only change that concerns me a bit is the last one in matroskafile.
Obviously you know the code better than me to know if it should be absolute or relative, but if I follow the flow correctly it's passed to `readLimited` which seems to use it as a relative.~~

```cpp
bool EBML::MkSegment::readLimited(File &file, offset_t scanLimit)
{
  const offset_t filePos = file.tell();
  const offset_t maxOffset = filePos + dataSize;
  const offset_t maxScanOffset = filePos + std::min(scanLimit, dataSize);
```

~~So I think something needs to change there as well, or not?~~

Nevermind, ignore all that. I didn't notice it gets reassigned.
So I think there should be no problem.

```cpp
  maxOffset = segment->getDataSize();
  if (readStyle == Properties::ReadStyle::Fast && maxOffset > FAST_SCAN_LIMIT) {
    maxOffset = FAST_SCAN_LIMIT;
  }
  if(!segment->readLimited(*this, maxOffset)) {
    debug("Failed to read segment");
    setValid(false);
    return;
  }
```

One more edit:

 I didn't look into how the FAST_SCAN_LIMIT is supposed to work, but since you say it should be absolute, that  could mean a similar problem could exist in the readLimited code?

ufleisch:
Thanks for reviewing my changes. The `maxOffset` passed to `readLimited()` is its `scanLimit` parameter, which is relative. So in this code fragment, the variable name is a bit misleading, but consistent with the same check a few lines above. As the `scanLimit` in the "fast scan" case is arbitrary anyway (512KB) and the code for that case is not changed by this PR, I would say that it is OK.

anarcobra:
I'm sorry to say I found a regression with my patch.
I'm still looking into the best way to fix it.
But there is an unfortunate interaction between elements of unknown size and the new size checking code.

PR #1377, opened 2026-07-05
[Matroska] Fix unknown size elements partially

anarcobra:
In matroska an element data size of a VINT with all bits 1 means the data size is unknown. Unknown data size can only apply to Master Elements.

Unknown sized elements are described in
https://datatracker.ietf.org/doc/rfc8794/ section 6.2 It gives the following 5 conditions for detecting the end of an unknown sized element:

 *  Any EBML Element that is a valid Parent Element of the Unknown- Sized Element according to the EBML Schema, Global Elements excluded.

 *  Any valid EBML Element according to the EBML Schema, Global Elements excluded, that is not a Descendant Element of the Unknown-Sized Element but shares a common direct parent, such as a Top-Level Element.

 *  Any EBML Element that is a valid Root Element according to the EBML Schema, Global Elements excluded.

 *  The end of the Parent Element with a known size has been reached.

 *  The end of the EBML Document, either when reaching the end of the file or because a new EBML Header started.

In this patch we use the higher level maxOffset to determine the maximum data size for the element, which matches the fourth condition, but is incomplete without the other four methods.

### DIFF
--- a/taglib/matroska/ebml/ebmlelement.cpp
+++ b/taglib/matroska/ebml/ebmlelement.cpp
@@ -39,7 +39,12 @@
 using namespace TagLib;
 
 #define RETURN_ELEMENT_FOR_CASE(eid) \
-  case (eid): return make_unique_element<eid>(id, sizeLength, dataSize, offset)
+  case (eid): \
+    if (sizeUnknown && !std::is_base_of_v<MasterElement, GetElementTypeById<eid>::type>) { \
+      debug(Utils::formatString("EBML: non MasterElement with unknown size %x" , eid)); \
+      return nullptr; \
+    } \
+    return make_unique_element<eid>(id, sizeLength, dataSize, offset)
 
 std::unique_ptr<EBML::Element> EBML::Element::factory(File &file, offset_t maxOffset)
 {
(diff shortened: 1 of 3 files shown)


anarcobra:
I'll see if I can make some small mkv with an unknown size and add a test.
Do you know any good tools to do so?

anarcobra:
So I was able to make a valid file (as far as I can tell) with unknown size by editing `no-tags.mka` with a hex editor.
However, even with my changes taglib won't parse it (although with my changes it does successfully read the original file I had with the unknown size). The main problem is that detecting the end of the unknown sized element is not straight forward without knowing which elements are allowed in each type of master element according to the mkv schema. Is there anything like that already, and is this even something that should be supported in taglib?
If not then maybe simply giving an error on unknown sized elements might give more consistent results instead of try to parse and hope it succeeds by chance.

ufleisch:
I have added a commit with the following changes:

- The unknown size can have any possible VINT size, but your check was only for the 1-byte case,
- I do not really want to generate lots of debug messages with the macros. If a check is really desired to test if an element has the permission for an unknown size, it could probably be a bit simpler because as far as I can see, only Segment and Cluster elements can have unknown size. Cluster elements are not really read by TagLib (their offsets can only be used for the "accurate validity check"). What remains is the Segment, which is probably handled good enough by your code, i.e. it can be assumed that it extends to the end of the file. A simple and tolerant solution might be good enough.

I have also taken `no-tags.mka` and edited it with a hex editor. The file seems to be read correctly. Could you please test the code with some real world files you have, I unfortunately do not have such files, the use case seems to be live streaming. If your tests are positive, I can add some unit tests if you like.

anarcobra:
> The unknown size can have any possible VINT size, but your check was only for the 1-byte case,

You're right! I completely missed that.

Fair enough about the debug messages. I will try your changes against my files, but on first glance it looks good. It's possible I made a mistake editing `no-tags.mka`.

### Edit:
I've run it on my files and compared to  the output with my changes and didn't find any differences.
So it looks good to me!
I did also try with my modified no-tags-mka, and also got the same output error I had before, but since all my actual files seem to work I think it's probably not worth looking in to more.

ufleisch:
Thanks for testing. I have added a unit test which makes the same things we did: It modifies a copy of _no-tags.mka_ and checks if it is still readable and considered valid, and if it can be modified. I have used unknown size length for both the Segment and the Cluster elements of that file, and I would say that it looks good. After modification, the Segment has an updated size, but the Cluster is not touched because it is not used by TagLib. The file is still considered valid by TagLib and other tools.

If you want to find out how our modified _no-tags.mka_ files differ, you could just set a breakpoint in the test and have a look at the modified copy, which can be found during the test in _/tmp/taglib-test.mka_ on Linux.

anarcobra:
Thanks! I'll take a look at it.

PR #1379, opened 2026-07-15
MP4: Fix QT chapters excessive sample allocation with invalid stsc

ufleisch:
See https://mail.kde.org/pipermail/taglib-devel/2026-July/003122.html

### DIFF
--- a/taglib/mp4/mp4qtchapterlist.cpp
+++ b/taglib/mp4/mp4qtchapterlist.cpp
@@ -825,7 +825,8 @@ namespace
     const std::vector<unsigned int> chunkOffsets = readStco(file, chapterTrak);
     if(chunkOffsets.empty())
       return {};
-    if(sizeInfo.sampleCount == 0)
+    // Sample count cannot be > file length / 4 as every sample consumes 4 bytes in stsz
+    if(sizeInfo.sampleCount == 0 || sizeInfo.sampleCount > file->length() / 4)
       return {};
 
     // Read stsc entries
--- a/taglib/mp4/mp4qtchapterlist.cpp
+++ b/taglib/mp4/mp4qtchapterlist.cpp
@@ -825,6 +825,8 @@ namespace
     const std::vector<unsigned int> chunkOffsets = readStco(file, chapterTrak);
     if(chunkOffsets.empty())
       return {};
+    if(sizeInfo.sampleCount == 0)
+      return {};
 
     // Read stsc entries
     struct StscEntry {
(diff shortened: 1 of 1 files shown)


ufleisch:
@ryanfrancesconi Could you as the original author have a look at this? For me it looks good.

hibrian827:
Thank you for opening a PR! The commit seems good to me. If you need anything more on my side for help, please let me know.

ryanfrancesconi:
Hi @ufleisch, thank you for fixing this quickly...

One thing to consider adding:

sizeInfo.sampleCount (read at `mp4qtchapterlist.cpp:806`, `info.sampleCount = data.toUInt(4U);`) is a raw 32-bit value from the 8-byte `stsz` header and is only checked against real file bytes when defaultSize == 0 

(the perSampleSizes loop at line 810, bounded by `pos + 4 <= data.size()`). When defaultSize != 0 -- the common case, and the one this patch's clamp relies on -- sampleCount is trusted.

That means the new clamp:

```c++
if(samplesInChunk > sizeInfo.sampleCount - sampleIndex)
    samplesInChunk = sizeInfo.sampleCount - sampleIndex;
```

stops a mismatched stsz/stsc pair, but not a consistent one. 

A file with stsz = {defaultSize: 1, sampleCount: 0x7FFFFFFF} and stsc.samplesPerChunk = 0x7FFFFFFF passes this check unchanged and still drives the same multi-GB push_back loop in resolveSampleOffsets -- same crash, just no longer requiring the two fields to disagree.

stco/stsc entry counts are bounded because readBlock truncates at EOF, but `stsz.sampleCount` has no equivalent tether back to real data.

Suggest also sanity-bounding sampleCount against something derived from the file size -- a text sample needs at least 2 bytes for its length prefix, so sampleCount can't legitimately exceed roughly file->length() / 2. 

Happy to send that as a follow-up PR if unclear.

Ryan

taglib/mp4/mp4qtchapterlist.cpp
```diff
@@ -825,6 +825,8 @@ namespace
     const std::vector<unsigned int> chunkOffsets = readStco(file, chapterTrak);
     if(chunkOffsets.empty())
       return {};
+    if(sizeInfo.sampleCount == 0)
```
ufleisch:
```suggestion
    // Sample count cannot be > file length / 4 as every sample consumes 4 bytes in stsz
    if(sizeInfo.sampleCount == 0 || sizeInfo.sampleCount > file->length() / 4)
```

ufleisch: @ryanfrancesconi Do you agree with this change? I wanted to make the condition dependent on `sizeInfo.defaultSize > 0`, but looking at the code I think that `samplesInChunk` and `sizeInfo.sampleCount` could also be large in that case.

ryanfrancesconi: Yes, that looks right to me!

PR #1381, opened 2026-07-21
Fix Shorten AIFF chunk offset wrap

acts-1631:
A crafted Shorten file can use an oversized embedded AIFF chunk length
to wrap the parser cursor and loop indefinitely.

Validate AIFF chunk headers and padded sizes against the remaining
verbatim-header data, then advance each chunk to its checked end.


ufleisch:
Thanks for your contribution.

PR #1383, opened 2026-07-26
[MP4] allow extracting covr with wrong flags

MSOB7YY:
hello, thanks for this great tool!

i have some files that contains artwork `covr` but with wrong flags, usually from a buggy encoder. if im correct it also means that any tagger that doesn't provide correct flag while editing artwork, will cause taglib to no longer be able to read image again

sorry im not an expert, im confused why taglib has to enforce these flags? shouldn't the `covr` contain image bytes anyways and it wouldn't matter to know the format/type of that image?

- this pr allows returning cover art with/without correct flags
  - sniff the magical bytes to detect type
  - return image anyways even if type is unknown

- 2 sample files: [no_artwork_mp4.zip](https://github.com/user-attachments/files/30389409/no_artwork_mp4.zip)

---

⚠️ please note that the code changes are suggested by claude, while the code looks simple and correct, and tested to be working great, i still have to disclose it since im not experienced with c++

please let me know if im mistaken, thanks again ^^

### DIFF
--- a/taglib/mp4/mp4itemfactory.cpp
+++ b/taglib/mp4/mp4itemfactory.cpp
@@ -607,6 +607,28 @@ std::pair<String, Item> ItemFactory::parseFreeForm(
   return {atom->name(), Item()};
 }
 
+namespace {
+
+// Sniff the image format from magic bytes.
+MP4::CoverArt::Format sniffCoverFormat(const ByteVector &payload)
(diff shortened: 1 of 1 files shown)


ufleisch:
Thanks for your contribution. I've only made a few minor adjustments.

Issue #1380, opened 2026-07-20
Doubt regarding .pc file handling of zlib


EZ4Stephen:
I want to know whether there's any consideration for ZLIB_LIBRARIES_FLAGS to be changed to be set as `-lzlib` instead of `-lz` for MSVC-like compilers (if HAVE_ZLIB), or maybe make it add zlib to the `Requires:` line of the .pc template.

This inquiry comes due to me recently getting taglib from xmake-repo, and when a building a project I got `LINK : fatal error LNK1181: cannot open input file 'z.lib'`.

Previously when I got taglib from vcpkg, I didn't have such problems, but it appears vcpkg handles it by replacing out the "-lz".
https://github.com/microsoft/vcpkg/blob/3caf327b2369f05b26997f6e9d523e8edbd60269/ports/taglib/portfile.cmake

They also apparently try to add zlib to the Requies: section, but from the .pc file I have installed from vcpkg, it doesn't actually get inserted.

vcpkg's .pc looks as follows:

```
prefix=${pcfiledir}/../..
exec_prefix=${prefix}
libdir=${prefix}/lib
includedir=${prefix}/include

Name: TagLib
Description: Audio meta-data library
Requires:
Version: 2.3
Libs: "-L${libdir}" -ltag
Cflags: "-I${includedir}" "-I${includedir}/taglib"


```

I hope that gives sufficient context to understand why I have doubts on whether -lz is needed in Libs, or whether zlib should go to Requires.

ufleisch:
Thanks for the report. I have created a pull request #1384 to address this issue. I have tested it on Linux. Please test it on your development environment with MSVC on Windows.

PR #1384, opened 2026-07-28
Use Requires.private for zlib with pkg-config (#1380)

ufleisch:
Express dependency on zlib using Requires.private in pkg-config .pc file instead of adding -lz to the libs.

This will cause the following changes when using TagLib with pkg-config:
- When using a dynamic library, -lz will no longer be present in in linker command, but being linked transitively.
- When using a static library, pkg-config must be used with the --static command line argument, then libs will still contain -lz, which is needed when linking statically.
- The flags needed for zlib will be provided by pkg-config, so if -lz is not appropriate (e.g. with MSVC), correct linker flags will be provided.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -174,6 +174,8 @@ locations defined when building TagLib. Since we did not give a
 
 ```
 PKG_CONFIG_PATH=$HOME/pkg/lib/pkgconfig pkg-config --libs --cflags taglib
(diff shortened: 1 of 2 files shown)


EZ4Stephen:
@ufleisch I can confirm the PR worked. It has Requires.private: zlib and Libs: -L${libdir} -ltag.

Issue #1300, opened 2026-01-24
More than 2GB of memory allocated when constructing TagLib::MPEG::File


tsdgeos:
I have a fuzzed 3K file that when given to TagLib::MPEG::File allocates more than 2GB of memory

We are using taglib to create thumbnails for audio files in KDE, which means that it runs without user interaction. This makes it "easy" to DOS the user system.

I can attach the file if you want.

PR #1385, opened 2026-07-31
Fix ID3v2.4 frames with per-frame Unsynchronisation being silently emptied

bergos:
**This PR is based on analysis and code by Claude, manually reviewed by me.**

## Summary

Since `397b6c1d` ("Add check for ID3v2 frame data length (#1300)", first released in v2.2beta), any ID3v2.4 frame that carries its own **per-frame** Unsynchronisation flag — as opposed to the tag-wide flag — has its content silently discarded: `Frame::fieldData()` returns an empty `ByteVector` for it, even though the frame is perfectly well-formed. This affects any frame type that happens to use per-frame rather than tag-level unsynchronisation; I hit it specifically via `TIT2`/`TPE1`/`TALB` (title/artist/album).

## Root cause

`FrameFactory::prepareFrameHeader()` (`taglib/mpeg/id3v2/id3v2framefactory.cpp`) de-unsynchronises a frame's raw bytes via `SynchData::decode()` when the frame has its own Unsynchronisation flag set, which strips the `0x00` stuffing bytes and shrinks the buffer — but it never updates the frame header's declared `frameSize()` to match the new, shorter length. That mismatch between declared and actual size predates this bug and was harmless on its own, since `frameSize()` is also relied on elsewhere (e.g. `ID3v2::Tag`'s read loop) to skip to the next frame in the tag's original, on-disk layout, so it can't be corrected in place at decode time.

`397b6c1d` added a bounds check to `Frame::fieldData()` (`taglib/mpeg/id3v2/id3v2frame.cpp`) comparing that declared size against the actual buffer size, to catch malformed/truncated frames. It didn't account for the legitimate post-decode shrink above, so for any per-frame-unsynchronized frame the check sees a declared length larger than the real buffer and treats an entirely valid frame as corrupt, discarding its content.

Frames that only rely on the tag-level Unsynchronisation flag (or aren't unsynchronized at all) never hit the `SynchData::decode()` shrink path in `prepareFrameHeader()`, so they're unaffected — which is why, in my reproduction, `TXXX`/`TRCK`/`TDRC` frames read fine while `TIT2`/`TPE1`/`TALB` (which happened to carry the per-frame flag) came back empty.

## Fix

In `Frame::fieldData()` (`taglib/mpeg/id3v2/id3v2frame.cpp`), instead of rejecting the frame outright when `frameDataOffset + frameDataLength > frameData.size()`, clamp `frameDataLength` to what's actually available (`frameData.size() - frameDataOffset`). The only case that still bails out with `ByteVector()` is when even `frameDataOffset` doesn't fit inside `frameData` — i.e. the frame is genuinely too short to contain any data.

This is safe because `FrameFactory::prepareFrameHeader()` already rejects any frame whose declared size doesn't fit the raw tag data *before* the unsynchronisation decode step runs. So by the time `fieldData()` sees a too-large `frameDataLength`, it can only be explained by the legitimate post-decode shrink described above — never by a frame that was already invalid on disk.

## Test

Adds `testUnsynchDecodeID3v24Frame()` to `tests/test_id3v2.cpp`, alongside the existing `testUnsynchDecode()` (which only covers the tag-wide Unsynchronisation flag). The new test loads a minimal, hand-built 28-byte ID3v2.4 fixture, `tests/data/unsynch24.id3`, containing a single `TIT2` frame — UTF-16 encoded "Hi" with a BOM, unsynchronised per TagLib's own `SynchData::decode()`/`encode()` convention — with only the frame's own Unsynchronisation flag set and no tag-wide unsync flag. Without the fix, this frame's title comes back empty; with it, `f.tag()->title()` correctly returns `"Hi"`.

## Reproduction / real-world impact

Found via two independent, real-world-ripped MP3 albums with ID3v2.4 tags where `TIT2`/`TPE1`/`TALB` are UTF-16-encoded with the per-frame Unsynchronisation flag set (confirmed with `exiftool -v3`). Built native, unmodified TagLib at `v2.3.1` and ran the bundled `examples/framelist` and `examples/tagreader` directly against these files — no bindings or wrappers involved — and both tools show empty `TALB`/`TPE1`/`TIT2`.

Notably, `Tag::title()`/`artist()`/`album()` can mask this for files that also carry a legacy ID3v1 tag with the same data, since those getters fall back to ID3v1 when the ID3v2 value comes back empty — so this bug can look "fixed" for some files and not others purely based on incidental tag redundancy, not on whether the ID3v2 data was actually parsed correctly. Files without a redundant ID3v1 tag show missing title/artist/album with no fallback at all. Bisected to `397b6c1d` with `git bisect` between `v2.1.1` (good) and `v2.2beta` (bad).


ufleisch:
Thanks for the fix.

PR #1390, opened 2026-08-01
MP4: reject undersized table atoms

acts-1631:
A structurally valid MP4 can contain a QT chapter stts atom shorter than
its 12-byte full-box header. readStts() subtracts 12 from its length and
passes the negative result to readBlock(), where it becomes a very large
size. On a 256 MB file, reading chapters reached about 530 MB peak RSS.

This validates the minimum table-header sizes before reading chapter
payloads. It also applies the corresponding checks to the stco/co64/tfhd
offset-rewrite paths shared by MP4 tags and chapter lists, so malformed
tables cannot trigger an underflow during save.

The parser can still accept the surrounding file, but it now treats the
truncated table as empty. The full bundled test-data corpus gives the same
108 successful and 7 null results after the change.

I considered restricting FileStream::readBlock() globally, but the atom
parser is the layer that knows the table-specific minimum sizes. Rejecting
undersized tables avoids changing the semantics for valid large payloads.

### DIFF
--- a/taglib/mp4/mp4nerochapterlist.cpp
+++ b/taglib/mp4/mp4nerochapterlist.cpp
@@ -77,6 +77,8 @@ namespace
       for(const auto &atom : stco) {
         if(atom->offset() > offset)
           atom->addToOffset(delta);
+        if(atom->length() < 16)
+          continue;
         file->seek(atom->offset() + 12);
         ByteVector data = file->readBlock(atom->length() - 12);
         unsigned int count = data.toUInt();
@@ -96,6 +98,8 @@ namespace
       for(const auto &atom : co64) {
         if(atom->offset() > offset)
           atom->addToOffset(delta);
+        if(atom->length() < 20)
+          continue;
         file->seek(atom->offset() + 12);
         ByteVector data = file->readBlock(atom->length() - 12);
         unsigned int count = data.toUInt();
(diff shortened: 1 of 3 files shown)


ufleisch:
First, I want to thank you for all your contributions.

For this PR, I have a little remark. There are a few modifications like

```
+        if(atom->length() < 16)
+          continue;
         if(atom->offset() > offset)
           atom->addToOffset(delta);
```

This code is executed when the offsets of the atoms may have to be updated because a block was inserted into the file which made the offsets change. Here an invalid atom is detected, in some cases it is a valid but useless atom with an empty table. It is good that the code following the code fragment shown above is no longer executed because this would lead to negative results and large allocations. But wouldn't it be better to update the offset also for these invalid or valid but useless atoms? After all, they are still part of the MP4 file and some code could use the `offset()` which would be wrong because it has not been updated. What I am suggesting is changing these modifications to

```
         if(atom->offset() > offset)
           atom->addToOffset(delta);
+        if(atom->length() < 16)
+          continue;
```

acts-1631:
Thanks, that makes sense. I updated the branch so the in-memory atom
offset is adjusted before the size check, while malformed tables still
skip the payload read and rewrite.

I reran the focused reproducer and the bundled test-data corpus.

PR #1392, opened 2026-08-02
MP4: limit atoms in nested containers

acts-1631:
MP4 applies a 50,000-atom limit at the root level, but nested containers
such as moov did not enforce it. A valid 3.9 MB file containing 500,000
free atoms inside moov parsed successfully and reached about 145 MB RSS.

Apply the same per-level limit while parsing container children. Files
that exceed it are rejected before the tree can grow without bound.

The patched parser rejects the reproducer after 50,000 children. The
bundled test-data corpus produces the same 108 successful and 7 null
results. I also ran the focused input under ASAN/UBSAN without a report.

### DIFF
--- a/taglib/mp4/mp4atom.cpp
+++ b/taglib/mp4/mp4atom.cpp
@@ -40,6 +40,7 @@ namespace {
     "stbl", "minf", "moof", "traf", "trak",
     "stsd", "stem"
   };
+  constexpr int MAX_MP4_ATOM_COUNT_PER_LEVEL = 50000;
 } // namespace
 
 class MP4::Atom::AtomPrivate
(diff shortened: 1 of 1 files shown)


taglib/mp4/mp4atom.cpp
```diff
@@ -116,6 +117,12 @@ MP4::Atom::Atom(File *file, int depth)
         return;
       }
       while(file->tell() < d->offset + d->length) {
+        if(d->children.size() >= MAX_MP4_ATOM_COUNT_PER_LEVEL) {
+          debug("MP4: Maximum atom count exceeded");
+          d->length = 0;
+          file->seek(0, File::End);
+          return;
+        }
```
ufleisch, line 126:
```suggestion
        if(d->children.size() >= MAX_MP4_ATOM_COUNT_PER_LEVEL) {
          debug("MP4: Maximum atom count exceeded");
          d->children.clear();
          d->length = 0;
          file->seek(0, File::End);
          return;
        }
```

We could clear the large list of atoms, so the memory gets freed. This is also consistent with the top-level case.

acts-1631: Applied in 0053173e. Thanks.

PR #1398, opened 2026-08-04
fix(matroska): find Segments past the fast scan limit

clementperon:
Element::factory() rejects any element whose declared size runs past the bound it is given, and read() passes the Fast scan limit as that bound. A Segment spans practically the whole file, so under ReadStyle::Fast every Matroska over 512 KiB is rejected and no tags are read:

  EBML: datasize too great: 1003369 > (524288 - 52)
  Failed to find Matroska segment

The limit is readLimited()'s, which already applies it to skip Cues and to bound the walk over a segment with no usable SeekHead. The lookup only needs the file length.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/taglib/matroska/ebml/ebmlmksegment.cpp
+++ b/taglib/matroska/ebml/ebmlmksegment.cpp
@@ -95,13 +95,13 @@ bool EBML::MkSegment::readLimited(File &file, offset_t scanLimit)
   MasterElement *pendingPaddingTarget = nullptr;
   offset_t accumulatedPadding = 0;
   std::unique_ptr<Element> element;
-  while((element = findNextElement(file, maxScanOffset))) {
+  while((element = findNextElement(file, maxOffset, maxScanOffset))) {
     if(const Id id = element->getId(); id == Id::MkSeekHead) {
       seekHead = element_cast<Id::MkSeekHead>(std::move(element));
       if(!seekHead->read(file))
         return false;
       // We have a seek head, let's use it for faster access to the other elements
-      if(const auto elementAfterSeekHead = findNextElement(file, maxScanOffset);
+      if(const auto elementAfterSeekHead = findNextElement(file, maxOffset, maxScanOffset);
          elementAfterSeekHead && elementAfterSeekHead->getId() == Id::VoidElement)
         seekHead->setPadding(elementAfterSeekHead->getSize());
       const offset_t segDataOffset = segmentDataOffset();
--- a/taglib/matroska/ebml/ebmlutils.cpp
+++ b/taglib/matroska/ebml/ebmlutils.cpp
@@ -30,10 +30,13 @@
 
 using namespace TagLib;
 
-std::unique_ptr<EBML::Element> EBML::findElement(File &file, Element::Id id, offset_t maxOffset)
+std::unique_ptr<EBML::Element> EBML::findElement(
+  File &file, Element::Id id, offset_t maxOffset, offset_t maxScanOffset)
 {
+  if(maxScanOffset < 0)
+    maxScanOffset = maxOffset;
   std::unique_ptr<Element> element;
-  while(file.tell() < maxOffset) {
+  while(file.tell() < maxScanOffset) {
     element = Element::factory(file, maxOffset);
     if(!element || element->getId() == id)
       return element;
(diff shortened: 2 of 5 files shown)


clementperon:
@acts-1631 AI investigated + fixed. 

This issue has been triggered by bumping taglib into Kodi see https://github.com/xbmc/xbmc/pull/28770#issuecomment-5167084694

ufleisch:
Thanks for the fix, and sorry for the bug.

We mixed up a few things, and the problems are the consequence of this:

- `maxOffset`: Maximum offset from the beginning of the file; the end of
  the element must be before this offset.
- `scanLimit`: Offset from the current file position until which scanning
  for elements is allowed. Normally, elements are scanned up to the end
  of the enclosing master element or the end of the file, but in Fast
  reading mode, it is limited to `FAST_SCAN_LIMIT`, which is 512 kB.
- `maxScanOffset`: scanLimit from the current file position

Your patch fixes the current problem. However, I think that it would not make the people happy which introduced the `FAST_SCAN_LIMIT` (see the discussion in PRs #1347, #1354, #1374). I have introduced additional parameters where necessary and tried to name them in the correct way, so their purpose is clear. Please have a look at the current changes of the PR and feel free to tell me if I made a mistake.

clementperon:
Thanks @ufleisch I quickly read and it seems fine.

Ask AI what he propose to keep a fast mode.

```
One behaviour change worth deciding deliberately: with the size bound gone from elements inside
the segment, a large element that *starts* within the scan window is now read whole. Measured with
a 4 MiB attachment, `AudioProperties::Fast`, via `attachedFileList()[0].data().size()`:

|                 | before          | after           |
| --------------- | --------------- | --------------- |
| no SeekHead     | not read        | 4194304 bytes   |
| behind SeekHead | 4194304 bytes   | 4194304 bytes   |

That is the cost `skipCues` exists to avoid — your comment there applies word for word — but
`Id::MkAttachments` (ebmlmksegment.cpp:173 and :232) has no equivalent guard, and
`BinaryElement::read()` loads the payload unconditionally.

Two options, and I'd rather ask than guess:

1. Apply the existing `file.readOnly() && scanLimit < dataSize` condition to attachments too.
   Small, mirrors `skipCues`, but `attachments()` then returns nothing in fast mode.
2. Read the structure without the payload, so a fast scan still reports that cover art is there
   with its name, media type and size. More useful, but the data has to stop being eager all the
   way up to `Matroska::AttachedFile` — probably an `AttachedFile::dataSize()`, hence a public API
   question.
```

Happy to prepare either.

ufleisch:
The intention was that `Properties::ReadStyle::Fast` for Matroska still will have no disadvantages for its users as long as the Matroska file is well behaving according to the specification, that means it has a seek head within the first 512 kB which references all important elements. It just avoids long scans with files not having such a seek head.

The thing with the Cues is that reading them is not needed when the file is not written, reading them is only necessary to be able to update its offsets in the case that the position of the Cues is changed because a write operation inserts or removes a block earlier in the file. That's why the Cue reading is only omitted when `readOnly()`. I would not omit other things like Attachments or binary tags in this situation, there is a high chance that some users want to use fast mode without missing the attachments.

I like the idea of making things lazy instead of eager. I could imagine that it is possible to just register offset and size of attachments during read and only read them when they are really requested, e.g. by the user calling `Matroska::File::attachments()`. I am, however, not sure whether it is really worth it or if the time to get the offsets and sizes is nearly the same as reading them -  maybe with network file systems.

If you find the time to work on that, feel free. Regarding this PR I will merge it if no one objects.

clementperon:
@ufleisch thanks for the explanation.

I don't have all the history but I think the main idea of Properties::ReadStyle::Fast was mainly to browse libraries that are over network filesystem. @HomerJau

I like your idea and I think the lazy worth it, we are talking about <1kB of tags that could increase to ~200KB-1MB with the attachement. If you have library with >100 of files and a cable connection this could be an issue.

I will try to open a different PR for this feature to not block this one and this if this feature stay localized and keep the code clean.

HomerJau:
Thanks!

HomerJau:
The issue with Kodi is it still tries to support old CPUs and devices with crappy networking chain. That should not hold back modern performant devices and infrastructure. It’s cheap to upgrade, new Kodi version = time to update. Users have the option to remain on older releases.

clementperon:
@HomerJau is it only a question of infrastructure? What if you have hundreds of mastroka files over FTP? Is it not better have the cover art loaded asynchronously?

Open a PR AI driven: https://github.com/taglib/taglib/pull/1402

HomerJau:
I do have hundreds of Matroska tagged albums and Concerts. It’s testable and perf tested with TagLib.

NFS is x5 faster than SMB for small reads (tag reads). NFS 4 is fester than earlier versions. Linux 7 has kernel perfs too. I am recommending Linux NUCs to everyone. More expensive than AMLogic but. I want perf for music. I use NUCs.

clementperon:
But you store all of these files locally right? so with a 1gbps local connection?

What if this library is remotely stored on a FTP where it's a shared server and you have like only few Mbps ? it will take from one to several seconds per title to get this attachement no ?

HomerJau:
The reality is Matroka file are huge compared to FLAC. An album with TrueHD Atmos, DTS-HD 5.1 and 99/24 stereo will be in excess of 2GB with no video stream. 4K Concert with Atmos over 50GB.

clementperon:
@HomerJau correct I will let @ufleisch decide if he want to pick it or not

PR #1400, opened 2026-08-05
Fix data races in lazily initialized shared caches

triplef:
#1331 fixed a data race in `MP4::ItemFactory`'s lazy map initialization. The same pattern — a mutable function-local `static` populated behind an unsynchronized `isEmpty()` check — exists in four other places:

* `ID3v2::FrameFactory::rebuildAggregateFrames()`: `tiplKeys` (read)
* `TextIdentificationFrame::involvedPeopleMap()`: `m` (read)
* `RIFF::Info::Tag::setProperties()`: `idForPropertyKey` (write)
* `ASF::Tag::setProperties()`: `reverseKeyMap` (write)

The first two are reached from `ID3v2::Tag::parse()` → `MPEG::File::read()`, i.e. plain `FileRef` construction, so parsing different files concurrently on several threads is enough to trigger them. `involvedPeopleMap()` also hands out a `const &` to a map another thread may be `insert()`ing into. We hit this in production when scanning a media library on multiple threads.

Since TagLib is built with C++17 I used function-local statics to fix this.

Also added `tests/test_threadsafety.cpp`, which releases 8 threads from a spin barrier so they contend on the very first use of each cache. Reverting the fixes and rerunning under ThreadSanitizer shows races/crashes. Tested on macOS/arm64, Apple clang 21.

### DIFF
--- a/taglib/asf/asftag.cpp
+++ b/taglib/asf/asftag.cpp
@@ -341,12 +341,13 @@ void ASF::Tag::removeUnsupportedProperties(const StringList &props)
 
 PropertyMap ASF::Tag::setProperties(const PropertyMap &props)
 {
-  static Map<String, String> reverseKeyMap;
-  if(reverseKeyMap.isEmpty()) {
+  static const Map<String, String> reverseKeyMap = [] {
+    Map<String, String> map;
     for(const auto &[k, t] : keyTranslation) {
-      reverseKeyMap[t] = k;
+      map[t] = k;
     }
-  }
+    return map;
+  }();
 
   const PropertyMap origProps = properties();
   for(const auto &[prop, _] : origProps) {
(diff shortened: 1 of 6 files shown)


triplef:
I fixed the CI failure in the new threadsafety test on MinGW, which was due to concurrent r/w access to "has-tags.m4a", which isn’t supported on Windows. Now we’re only opening the file once.

ufleisch:
GitHub actions were down for some time, so I pushed an empty commit to trigger another build. The MinGW build succeeded, but now the MSVC build failed.

triplef:
Thanks @ufleisch. The test should be fixed now — could you run GH Actions again please?

ufleisch:
Finally all tests pass. Thanks for the patch. I must confess that this is the first time I have seen these "Thread-Safe Dynamic Initialization". My code using TagLib uses Qt and tries to avoid multi-threading as much as possible. I just hope that I will never have to debug it. If you find other things, feel free to contribute.

triplef:
Thanks, I appreciate it! In our app we sometimes have to get data from large numbers of files (10k+), so multi-threading is essential for us for performance. We’ll keep an eye on similar issues in the future.

PR #1408, opened 2026-08-10
MP4: merge the QT chapter reference into an existing tref

ryanfrancesconi:
### Problem

A `trak` may hold at most one `tref`. `setQtChapters()` appended a second one to carry the `chap` reference instead of joining the atom already present, so any file whose audio track already references another track ends up with two.

That shape is common rather than exotic: camera recordings reference a timecode track from their audio track, so writing chapters to one produces an invalid `trak`.

Parsers differ on the result, which is what makes it costly. Lenient ones ignore the surplus atom and keep reading the track — TagLib itself among them, so a round-trip through TagLib looks perfectly healthy. Stricter ones discard the whole track rather than just the reference, and report the file as having no audio at all.

The damage lands when something remuxes the file from a strict parser's track list: the audio track is not in that list, so it is not copied, and the result is a valid file with the audio permanently gone. The chapter write itself is recoverable — the samples are untouched and only the `moov` is malformed — but nothing warns in between, because every tool that can still see the track reports success.

### Fix

`QtChapterList::write` inserts only the `chap` box when the audio track already has a `tref`, growing that atom, and inserts a whole `tref` only when there is none. The two insertions are ordered highest-offset-first, and the atom tree is corrected once per insertion — a single combined delta would move the `stco` atoms lying between them further than the file did, and the following seek would write chunk offsets into the next `trak`.

`removeAudioTref` follows the same rule inverted: only the `chap` box is removed when the `tref` is shared, and the `tref` itself goes only when `chap` was its sole child. It now also looks for the `tref` that actually contains `chap` rather than the first one on the track, so a file already carrying two is repaired rather than stripped of the wrong reference.

### Test

`testQTChapterListSharedTref` injects a `tref > tmcd` into `no-tags.m4a`'s audio track, since no file in `tests/data` carries one, then asserts:

- one `tref` holding `tmcd` before the write
- one `tref` holding `tmcd` and `chap` after it
- one `tref` holding `tmcd` again after removal, the timecode reference intact

It fails on `master` in both directions: two `tref` atoms after the write, and the timecode reference deleted by the removal.

Full suite passes locally (`OK (571)`).

### DIFF
--- a/taglib/mp4/mp4qtchapterlist.cpp
+++ b/taglib/mp4/mp4qtchapterlist.cpp
@@ -690,13 +690,68 @@ namespace
 
   // -- tref / chap builder --------------------------------------------------
 
-  //! Builds a tref atom containing a chap reference to the given track ID.
-  ByteVector buildTref(unsigned int chapterTrackId)
+  //! Builds a chap reference box pointing at the given track ID.
+  ByteVector buildChap(unsigned int chapterTrackId)
   {
     ByteVector chapData;
     chapData.append(ByteVector::fromUInt(chapterTrackId));
-    const ByteVector chap = renderAtom("chap", chapData);
-    return renderAtom("tref", chap);
+    return renderAtom("chap", chapData);
+  }
+
+  //! Builds a tref atom containing a chap reference to the given track ID.
+  ByteVector buildTref(unsigned int chapterTrackId)
+  {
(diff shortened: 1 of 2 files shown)


ryanfrancesconi:
Above description of the issue is by Claude. I found the bug when implementing video editing on a pre-tagged file (chapter marker'd) and had the audio track of the MP4 disappear on me.

PR #1413, opened 2026-08-13
RIFF: support RF64 and BW64

### LINKED ISSUE #1412: RIFF: a tag write on an RF64 file destroys the audio
An RF64 file is a `.wav` past 4 GB: `RF64` at offset 0 instead of `RIFF`, a `0xFFFFFFFF` sentinel in the 32-bit size field at offset 4, the same sentinel as the `data` chunk's 32-bit size, and the true 64-bit sizes in a `ds64` chunk that must come first. This is what field recorders (Sound Devices, Zaxcom), Reaper, ffmpeg and Core Audio write for long takes. BW64 is the same layout with different magic.

TagLib does not reject these files. It opens them, reports nothing, and a `save()` that returns `true` leaves the audio unreadable.

### The detection check exists but is never reached

`RIFF::WAV::File::isSupported` correctly returns `false` for RF64 — it requires `RIFF` at 0 and `WAVE` at 8 (`riff/wav/wavfile.cpp`). But `FileRef` tries `detectByExtension` first (`fileref.cpp`), which constructs a `RIFF::WAV::File` for any `.wav` and keeps it when `isValid()`. And it is valid, because `RIFF::File::read()` never inspects the magic: it reads a size at offset 4 and walks chunks from offset 12, which the RF64 layout satisfies structurally.

On a 4.8 GB RF64 written by Core Audio (25000 s, 48 kHz/16-bit stereo, 4,800,004,096 bytes):

```
WAV::isSupported = 0        FileRef isNull = 0        isValid = 1
properties = 0 entries      length = 0 s, bitrate = 0, sampleRate = 0, channels = 0
setProperties(TITLE, ARTIST) -> 0 unsupported
save() = 1                  <-- success
```

The `0 s` has a specific cause worth recording: `read()` stores the `data` chunk's declared size unchanged, so `RIFF::WAV::Properties::read()` computes `streamLength = 0xFFFFFFFF + 1` padding byte, which overflows `unsigned int` to exactly 0 and is then reported as "'data' chunk not found".

### What the save does

| | before `save()` | after |
| --- | --- | --- |
| size field at offset 4 | `ff ff ff ff` | `6e 14 00 00` (5230) |
| `ds64` riffSize / dataSize | 4,800,004,088 / 4,800,000,000 | unchanged |
| `data` chunk header size | `ffffffff` | unchanged |
| `afinfo` duration | 25000.000000 sec | **0.005958 sec** |
| `afinfo` audio bytes | 4,800,000,000 | **1144** |

`RIFF::File::updateGlobalSize()` recomputes a 32-bit total and inserts it at offset 4 unconditionally. In RF64 that slot must stay `0xFFFFFFFF`; once it holds a real number, readers stop consulting `ds64` and believe it. Four and a half hours of audio now reads as six milliseconds.

### The audio is also modified in place

This is the part that makes it more than a header problem. `setChunkData(name, data)` computes its append offset as `last.offset + last.size + last.padding`, and `last.size` is the `data` chunk's declared size — the `0xFFFFFFFF` sentinel. So the new tag chunk is not appended after the audio; it is **spliced into the middle of it**, and everything after is displaced by the chunk's length.

Measured on the same file: the `ID3 ` chunk landed at offset **4,294,971,392** — exactly 4 GiB + 4096, i.e. `data.offset + 0xFFFFFFFF + 1` — 505,032,704 bytes short of the end of the audio.

Restoring the four bytes at offset 4 brings the duration back (`ds64` is untouched), but the spliced block remains embedded, so recovery also requires excising it. Verified by removing both and confirming an identical SHA-256 over all 4,800,000,000 bytes of the payload.

### Below 4 GB the same shape loses the tags instead

On a small RF64, `save()` returns true, the sentinel is overwritten with an accurate size, and a re-read finds **0 properties**: the appended `LIST`/`INFO` falls inside the region `read()` clamped the sentinel `data` chunk to (the clamp added in #1329), so it is never seen again. The audio survives, because at that size a real 32-bit total describes the file correctly.

There is a nasty symmetry here: a reader built on this code and a writer built on this code agree with each other. The chunk walk advances by `0xFFFFFFFF` and lands exactly on the chunk the writer placed there, so tags round-trip and an application using TagLib on both sides sees nothing wrong, while every other reader sees a destroyed file.

### Wave64 is unaffected

`.w64` is in no `detectByExtension` branch and no content sniffer matches it, so `FileRef` is null and every write path fails loudly. Confirmed against a Core Audio-written `.w64`.

### Environment

taglib 2.3.1, macOS 15.5 / arm64. Measurements taken 2026-08-13 with a standalone C++ probe linked against the library, and `afinfo` (macOS Core Audio) as the independent reader. The same file reads correctly in Core Audio, AVFoundation and the Adobe XMP SDK before the save.

I have a fix and will open a PR against this issue.

ryanfrancesconi:
Fixes #1412.

RF64 and BW64 are the long forms of WAVE used past 4 GB: each 32-bit size field holds a `0xFFFFFFFF` sentinel and the real sizes live in a leading `ds64` chunk. `RIFF::WAV::File::isSupported` rejected them, but `FileRef` reaches the class by extension for any `.wav` and `RIFF::File::read()` never inspected the magic, so these files opened as valid and a tag save destroyed them. The issue has the measurements.

### What this changes

- **`RIFF::WAV::File::isSupported`** accepts `RF64` and `BW64` at offset 0 alongside `RIFF`, still requiring `WAVE` at 8. Content detection then works, and the check stops being a lie about what the class accepts.
- **`RIFF::File::read()`** reads the magic, records whether the file is long-form, parses `ds64`, and uses its 64-bit `dataSize` for a `data` chunk whose 32-bit size is the sentinel.
- **`RIFF::File::updateGlobalSize()`** writes the new total into `ds64`'s `riffSize` and writes the sentinel back at `sizeOffset`.
- **`RIFF::WAV::Properties::read()`** takes the stream length through a new 64-bit accessor, so a long file reports its real duration instead of `0 s`.

### Design decisions worth reviewing

**The `ds64` table is deliberately not parsed.** It carries 64-bit sizes for *additional* oversized chunks; the `data` chunk has its own dedicated field and is the only one that is ever large. Any chunk listed in that table stays on the existing clamp path from #1329, which this leaves untouched. Core Audio writes an empty table. If someone produces a real file with a non-empty one, the layout should be verified against EBU Tech 3306 before code is written against it — the four fixed fields here are confirmed empirically, the table row is not.

**`Chunk::size` becomes `offset_t`, but no public or protected signature changes.** This is needed for correctness rather than tidiness: `setChunkData`'s append path computes `last.offset + last.size`, which in 32 bits lands ~505 MB short of the end of a 4.8 GB `data` chunk — that is the splice described in the issue. `struct Chunk` is file-scope in `rifffile.cpp` and `FilePrivate` is only forward-declared in the header, so the change is invisible outside the translation unit: **no ABI break for anyone subclassing `RIFF::File`.**

`chunkDataSize()` still returns `unsigned int`, now saturating at `0xffffffff` rather than truncating. A new protected `chunkDataSize64()` carries the real value where it matters. Widening the existing accessor instead would have been more truthful but changes a protected signature, and nothing in the tree reads audio samples through that API — every `chunkData()` call is for `LIST`, `iXML`, `bext`, `fmt ` or `fact` in WAV and `COMM` in AIFF, never for `data`. `chunkData()` now saturates its read length for the same reason, since a `ByteVector` cannot hold more than 32 bits anyway.

**Writing the sentinel rather than skipping the write** is what repairs a file damaged by an earlier version. A long-form file always carries `0xFFFFFFFF` there, so any other value is malformed; writing it unconditionally is simply correct, and it means one ordinary tag save restores a file whose sentinel was clobbered. Measured: the 4.8 GB file from the issue went from 0.006063 sec back to 25000.000000 sec. It does not undo the splice — that needs explicit repair.

**AIFF cannot reach the new branch.** It is big-endian and has no long form, and the long-form detection is gated on little-endian magic. `RIFF::AIFF::File` and `RIFF::WAV::File` are the only subclasses of `RIFF::File` (`DSDIFF::File` derives from `TagLib::File` directly).

### Tests

Four new cases in `tests/test_wav.cpp`, all failing before this change:

- `testRF64IsSupported` — RF64 and BW64 accepted, garbage magic rejected.
- `testRF64Properties` — duration, sample rate and channels come from `ds64`. A trailing chunk is appended first, because with `data` last the clamp coincidentally lands on the right answer and the test would pass against the bug.
- `testRF64Save` — tags round-trip; afterwards the sentinel is intact, `ds64`'s `riffSize` tracks the file, and `dataSize` is unchanged.
- `testRF64SaveRepairsClobberedSize` — a file with a real total where the sentinel belongs is repaired by a save.

`tests/data/rf64.wav` is 9,680 bytes and built by construction rather than by an encoder, so it can be committed and is deterministic. Core Audio reads it as RF64 at 0.050000 sec. The sentinels behave identically at any size, so a small fixture covers the detection failure, the overwrite and the repair; the >4 GB truncation was proved by hand and is in the issue.

The full suite passes (`OK (576)`). The existing WAV tests are the guardrail that matters most here — this changes `RIFF::File`, not an RF64-only path.

### Not in scope

- BW64's `axml`/`chna` chunks (ADM metadata). Reading them is a separate feature; this is about not corrupting the container.
- Wave64 (`.w64`), which is a different container with GUID chunk headers and is already safe — `FileRef` is null for it and every write path fails loudly.


ryanfrancesconi:
Fix and previous bug have also been live tested against ShadowTag 1.8.2 (current taglib) - and a build with the fix in place.

ufleisch:
Thanks! I just added a commit to fix two `-Wconversion` warnings in the Ubuntu build of the CI.

PR #1416, opened 2026-08-15
DSF: do not convert an out of range length to unsigned int

manuaudio:
### What

`DSF::Properties::read` computes the length as

```cpp
d->length = d->samplingFrequency > 0
    ? static_cast<unsigned int>(static_cast<double>(d->sampleCount) * 1000.0 / d->samplingFrequency + 0.5)
    : 0;
```

`sampleCount` is a `long long` read straight from the file and `samplingFrequency` is an unsigned int from the file, so the millisecond count can land well outside `unsigned int`. Converting a floating point value the destination type cannot represent is undefined:

```
taglib/dsf/dsfproperties.cpp:132:35: runtime error: 8.41595e+09 is outside
the range of representable values of type 'unsigned int'
    #0 TagLib::DSF::Properties::read(TagLib::ByteVector const&) dsfproperties.cpp:132
    #1 TagLib::DSF::Properties::Properties(...) dsfproperties.cpp:59
    #10 TagLib::FileRef::FileRef(char const*, bool, ...) fileref.cpp:361
```

It is on the plain parse path — constructing a `FileRef` over the file is enough. A negative `sampleCount` converts just as badly, so the guard covers that too.

### The change

Report an unknown length rather than converting, which is what a zero sampling frequency already does.

Nothing changes for a valid file — `tests/data/empty10ms.dsf` reads `lengthMs=10 bitrate=5645 rate=2822400 ch=2` both before and after. Reverting just this hunk and rebuilding brings the report straight back.

I could not run the CppUnit suite (CppUnit isn't installed here, so `BUILD_TESTING` skips the tests directory) — the checks above are what I have.

### `bitrate` on the line above is fine

Worth saying since it looks similar: `samplingFrequency * bitsPerSample * channelNum` is unsigned arithmetic, so it wraps rather than overflowing, and dividing by 1000.0 leaves at most ~4.3e6, comfortably inside `unsigned int`.

### One I could not demonstrate

`DSDIFF::Properties` (`dsdiffproperties.cpp:60`) has the same shape, casting to `int` instead. There the sample count comes from a chunk size that `chunkFits` has already validated against the file length, so triggering it needs a genuinely large file — roughly 262 KB with a sample rate of 1 — rather than a few mutated bytes. I did not build one, so I have left it alone rather than claim it. Happy to follow up if you'd like it guarded on the same grounds.

### How it was found

Mutating the files in `tests/data` and running each through a parse → read every property → modify → `save()` round trip, under ASan and UBSan.

I calibrated the harness first: with a deliberate one byte over-read injected into `FileStream::readBlock`, ASan reports it through that path, and the report goes away when the injection is removed. Without that check a clean run would not have meant anything.

### Disclosure

Written with Claude Code assisting, under my direction; the commit carries an `Assisted-By` trailer. I verified the before/after myself on macOS/arm64.

X: manuaudiomix

### DIFF
--- a/taglib/asf/asffile.cpp
+++ b/taglib/asf/asffile.cpp
@@ -25,6 +25,8 @@
 
 #include "asffile.h"
 
+#include <limits>
+
 #include <utility>
 
 #include "tdebug.h"
@@ -236,8 +238,14 @@ void ASF::File::FilePrivate::FilePropertiesObject::parse(ASF::File *file, long l
 
   const long long duration = data.toLongLong(40, false);
   const long long preroll  = data.toLongLong(56, false);
-  file->d->properties->setLengthInMilliseconds(
-    static_cast<int>(static_cast<double>(duration) / 10000.0 - static_cast<double>(preroll) + 0.5));
+  // duration and preroll are both read from the file, so the result can land outside
+  // int, and converting a double the destination type cannot represent is undefined.
+  const double milliseconds =
+    static_cast<double>(duration) / 10000.0 - static_cast<double>(preroll) + 0.5;
+
+  if(milliseconds > static_cast<double>(std::numeric_limits<int>::min()) &&
+     milliseconds < static_cast<double>(std::numeric_limits<int>::max()))
+    file->d->properties->setLengthInMilliseconds(static_cast<int>(milliseconds));
 }
 
 ByteVector ASF::File::FilePrivate::StreamPropertiesObject::guid() const
--- a/taglib/dsdiff/dsdiffproperties.cpp
+++ b/taglib/dsdiff/dsdiffproperties.cpp
@@ -25,6 +25,8 @@
 
 #include "dsdiffproperties.h"
 
+#include <limits>
+
 #include "tstring.h"
 
 using namespace TagLib;
@@ -57,9 +59,17 @@ DSDIFF::Properties::Properties(unsigned int sampleRate,
   d->sampleWidth = 1;
   d->sampleRate = sampleRate;
   d->bitrate = bitrate;
-  d->length = d->sampleRate > 0
-    ? static_cast<int>(static_cast<double>(d->sampleCount) * 1000.0 / d->sampleRate + 0.5)
-    : 0;
+  // sampleCount is derived from a chunk size in the file and sampleRate is read from
+  // it, so the millisecond count can land outside int, and converting a double the
+  // destination type cannot represent is undefined. Report an unknown length instead,
+  // which is what a zero sample rate already does.
+  d->length = 0;
+  if(d->sampleRate > 0 && d->sampleCount > 0) {
+    const double milliseconds =
+      static_cast<double>(d->sampleCount) * 1000.0 / d->sampleRate + 0.5;
+    if(milliseconds < static_cast<double>(std::numeric_limits<int>::max()))
+      d->length = static_cast<int>(milliseconds);
+  }
 }
 
 DSDIFF::Properties::~Properties() = default;
--- a/taglib/dsf/dsfproperties.cpp
+++ b/taglib/dsf/dsfproperties.cpp
@@ -26,6 +26,8 @@
 
 #include "dsfproperties.h"
 
+#include <limits>
+
 using namespace TagLib;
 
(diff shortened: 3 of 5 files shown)


manuaudio:
Continued fuzzing turned up a third instance of the same shape, so I have added it here rather than opening another PR — `ASF::File` computes its length as

```cpp
static_cast<int>(static_cast<double>(duration) / 10000.0 - static_cast<double>(preroll) + 0.5)
```

with `duration` and `preroll` both read from the file properties object:

```
taglib/asf/asffile.cpp:240:22: runtime error: -2.80375e+14 is outside
the range of representable values of type 'int'
```

The guard only skips the conversion when the value will not fit, so values that do fit — negative ones included — are set exactly as before. The three `.wma` files in `tests/data` still report `lengthMs` 3549, 96502 and 3712, unchanged.

That makes three commits, each a self-contained instance of "a length computed from file-supplied numbers is converted to a type that cannot represent it": DSF, DSDIFF, ASF. Happy to split them if you would rather take them separately.

---

Separately, and **not** in this PR: the same fuzzing found that the ID3v2 frame parsers cast the text encoding byte straight out of the file,

```cpp
d->textEncoding = static_cast<String::Type>(data[0]);
```

with no check that the byte is one of the five values `String::Type` defines. UBSan reports it at parse time, and the invalid value then flows into `Frame::checkTextEncoding`, `renderFields` and `String::data` — the last of which switches on the type with no `default:` arm:

```
textidentificationframe.cpp:226: runtime error: load of value 4294967269,
which is not a valid value for type 'Type'
```

There are eleven of these across nine files, plus a twelfth of the same kind for `AttachedPictureFrame::Type`. I have a change ready that validates the byte in one place and falls back to `String::Latin1` — which is what the frames already declare as their default — and leaves the tags of all 30 tagged files in `tests/data` byte-identical.

I have deliberately not opened it, because where the helper belongs (`tutils.h` seemed natural) and whether an unknown byte should fall back or reject the frame feel like your calls rather than mine. Say the word and I will send it, or open an issue with the reproducer if you would rather keep it as a report.

manuaudio:
Two more of the same shape turned up, so this is now five commits. Both verified the same way — reverting only the hunk brings the report back, and the affected files in `tests/data` report identical properties either way.

**AIFF** (`aiffproperties.cpp`) — the sample rate is an 80 bit float read from the file, so it reaches values with no realistic bound:

```
aiffproperties.cpp:145:38: runtime error: 6.29416e+49 is outside the range of ... 'int'
aiffproperties.cpp:150:35: runtime error: 1.00707e+48 is outside the range of ... 'int'
```

All seven AIFF files in `tests/data` unchanged.

**MPC** (`mpcproperties.cpp`, `readSV8`) — `frameCount` is `sampleFrames - begSilence`, both from the file:

```
mpcproperties.cpp:244:39: runtime error: 4.18294e+17 is outside the range of ... 'int'
```

This one showed up independently from two different seed files, `sv8_header.mpc` and `zerodiv.mpc`.

So the five are DSF, DSDIFF, ASF, AIFF and MPC — one commit each, all "a length or rate computed from file-supplied numbers is converted to a type that cannot represent it". Happy to split, squash or drop any of them.

---

On the enum question I raised above, I have finished the survey and it is wider than the eleven ID3v2 sites I mentioned. Three different enums are built by casting a file-supplied value without checking it:

| enum | sites | valid range |
|---|---|---|
| `String::Type` | 11, across 9 files | `Latin1`..`UTF16LE`, 0–4 |
| `AttachedPictureFrame::Type` | 2 — `attachedpictureframe.cpp:147, 208` | 0–20 |
| `MP4::AtomDataType` | 2 — `mp4itemfactory.cpp:455, 464` | up to `TypeUndefined` = 255 |

Observed values were 4294967269 for the first, and 3321888768 for `AtomDataType`. Loading an enumeration object whose value is outside the enumeration is undefined, and in the `String::Type` case the value then reaches `String::data()`, which switches on it with no `default:` arm.

Still not opening anything for this — it is fifteen sites and the fallback policy is a design decision. Just wanted the survey to be complete rather than leave you with a partial number.

ufleisch:
Thanks for your contribution!

PR #1420, opened 2026-08-15
Validate enumerations decoded from file bytes before casting to them

manuaudio:
This is the class I mentioned on #1416. That PR merged without an answer either way, so rather than ask again I have built it out with the evidence attached.

Seven enumerations are decoded from file bytes by a plain `static_cast`. Loading an enumeration object whose value lies outside the range of the enumeration is undefined, and these ranges are small — `TimestampFormat` spans only 0..3, the picture type 0..31.

| commit | enumeration | sites |
|---|---|---|
| 1 | `String::Type` | 11 |
| 2 | the shared picture type | 4, across ID3v2 / FLAC / ASF |
| 3 | `RelativeVolumeFrame::ChannelType` | 1 |
| 4 | `SynchronizedLyricsFrame::TimestampFormat` + `::Type`, `EventTimingCodesFrame::TimestampFormat` | 3 |
| 5 | `MP4::AtomDataType` | 2 |

Each commit quotes the UBSan report it removes and how the file reaches it.

## Not just a sanitizer complaint

Two of these are visible through the public API. On a file whose picture type byte is `0xFF`:

- `AttachedPictureFrame::type()` returns **-1**
- `FLAC::Picture::type()` returns **-1**, or **65536** for a 32-bit value

`FLAC::Picture` is the widest of the three, since it casts a whole 32-bit field rather than a byte. And `String::data()` switches on the encoding with no `default:` arm, so an unrecognised encoding silently returns an empty `ByteVector` on the render path.

## One helper for the picture type, not three

The picture type enumeration comes from `DECLARE_PICTURE_TYPE_ENUM`, so the same defect is copied into three classes. Rather than three separate checks, the macro gains a `typeFromByte()` alongside the existing `typeFromString()`, backed by `Utils::pictureTypeFromByte()`. If you would prefer three local checks and no change to the shared header, that is an easy rework — say the word.

## What I checked and deliberately left alone

`eventtimingcodesframe.cpp:109` reads one line below a site I did change and looks identical:

```cpp
auto type = static_cast<EventType>(static_cast<unsigned char>(data[pos++]));
```

`EventType` enumerates up to `0xFE`, so its range is 0..255 and an `unsigned char` cannot leave it. That inner cast is exactly what the line above was missing. Not a bug, not touched.

Also unchanged: `typeFromString()` (`pictureTypeFromString` only ever returns 0..20) and `tvariant.cpp:289` (casts a `std::variant` index).

One precision worth stating, since it affects what the guards must do: a value that is *unnamed but in range* is not undefined. Text encoding byte 5 produces no sanitizer report at all, because `String::Type` enumerates 0..4 and its range is therefore 0..7. It is still not a valid encoding, and the helper rejects it — but on correctness grounds, not on undefined behaviour.

## Verification

- **19 purpose-built files** covering every enumeration: 16 UBSan reports before, **0 after**.
- Controls keep their valid values — RVA2 `channel=1`, SYLT `tsFormat=2 type=1`, picture `type=3`.
- Regression checked **four ways over the 123 files in `tests/data`** — audio properties, picture types, RVA2 channels, SYLT/ETCO fields and MP4 items — all byte-identical to `master`. The comparison is not vacuous: it includes real `APIC type=3` and real `covr` items.
- Suite runs 576 tests either way.
- Every before/after binary was hash-checked as actually different before its output was compared.

Happy to split this into separate PRs per enumeration if that is easier to review.

### DIFF
--- a/taglib/asf/asfpicture.cpp
+++ b/taglib/asf/asfpicture.cpp
@@ -133,7 +133,7 @@ void ASF::Picture::parse(const ByteVector& bytes)
   if(bytes.size() < 9)
     return;
   int pos = 0;
-  d->type = static_cast<Type>(bytes[0]); ++pos;
+  d->type = typeFromByte(static_cast<unsigned char>(bytes[0])); ++pos;
   const unsigned int dataLen = bytes.toUInt(pos, false); pos+=4;
 
   const ByteVector nullStringTerminator(2, 0);
--- a/taglib/asf/asfpicture.cpp
+++ b/taglib/asf/asfpicture.cpp
@@ -133,7 +133,7 @@ void ASF::Picture::parse(const ByteVector& bytes)
   if(bytes.size() < 9)
     return;
   int pos = 0;
-  d->type = typeFromByte(static_cast<unsigned char>(bytes[0])); ++pos;
+  d->type = typeFromByte(bytes[0]); ++pos;
   const unsigned int dataLen = bytes.toUInt(pos, false); pos+=4;
 
   const ByteVector nullStringTerminator(2, 0);
--- a/taglib/flac/flacpicture.cpp
+++ b/taglib/flac/flacpicture.cpp
@@ -68,7 +68,7 @@ bool FLAC::Picture::parse(const ByteVector &data)
   }
 
   unsigned int pos = 0;
-  d->type = typeFromByte(data.toUInt(pos));
+  d->type = typeFromUInt(data.toUInt(pos));
   pos += 4;
   unsigned int mimeTypeLength = data.toUInt(pos);
   pos += 4;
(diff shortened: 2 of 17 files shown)


ufleisch:
Sorry, you are right, I missed the comment about enums in #1416. As AI generated comments tend to be very informative, I usually stop reading when I have understood the code changes.

I have cleaned up the code a bit so that `FromByte` methods do not surprise with taking an `unsigned int`. Just tell me if this is not OK.

manuaudio:
That is better, thanks — taking the raw `char` and doing the `static_cast<unsigned char>` inside `typeFromByte` means no call site can sign-extend, and the FLAC 4-byte field is correctly on `typeFromUInt`. Fine by me. Noted on comment length, I will keep them shorter.

PR #1422, opened 2026-08-17
XM: reject zero-sized sample headers

acts-1631:
FileRef routes XM files to XM::File. XM::File::read() uses the file-controlled sample count and sample header size to process each instrument. If the sample header size is zero, each iteration consumes no input but still appends a sample entry. A crafted file can combine maximum instrument and sample counts to drive excessive CPU and memory use while an application inspects an untrusted module file.

Reject zero-sized sample headers before entering the sample loop.

### DIFF
--- a/taglib/xm/xmfile.cpp
+++ b/taglib/xm/xmfile.cpp
@@ -594,7 +594,10 @@ void XM::File::read(bool)
       sumSampleCount += sampleCount;
       // wouldn't know which header size to assume otherwise:
       READ_ASSERT(instrumentHeaderSize >= inCnt + 4 && readU32L(sampleHeaderSize));
-      READ_ASSERT(sampleHeaderSize > 0);
+      // Some trackers wrote zero here even though 40-byte sample headers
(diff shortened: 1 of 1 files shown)


ufleisch:
Looking at _xmfile.cpp_, I agree that the current behaviour with `sampleCount > 0` and `sampleHeaderSize == 0` just leads to a large comment with empty lines. The fix avoids this and marks the file as invalid. I am not an expert for tracker modules so I asked AI if a `sampleHeaderSize == 0` makes an XM file invalid. The answer is not clear, but it seems that there exist valid files with zero `sampleHeaderSize` and one way to cope with that is to assume a default of 40 bytes, which is the sum of the sizes of the readers set in line 613, so maybe using such a default would make sense.

Can someone with a collection of XM files give some insights, maybe @EvilGivEn ?

acts-1631:
Thanks for checking this. Zero is used by real XM files, including early Sk@leTracker files. I revised the fix to treat zero as a 40-byte sample header instead of rejecting the file. This preserves compatibility while ensuring the sample loop advances through the input and avoids the resource-exhaustion case.

PR #1425, opened 2026-08-17
Bound total elements in recursive Matroska and MP4 parsing

acts-1631:
Matroska::File::read() and MP4::File::read() recursively build element trees while parsing untrusted files. Both parsers limited children per container but reset that limit for nested containers. A crafted file could repeat tens of thousands of small children across nesting levels, causing excessive memory and CPU use when opened.\n\nCarry a shared element count through recursive Matroska master parsing and the MP4 atom tree while preserving the existing per-container limits.

### DIFF
--- a/taglib/matroska/ebml/ebmlmasterelement.cpp
+++ b/taglib/matroska/ebml/ebmlmasterelement.cpp
@@ -100,7 +100,14 @@ void EBML::MasterElement::setMinRenderSize(offset_t minimumSize)
 
 bool EBML::MasterElement::read(File &file, int depth)
 {
+  unsigned int elementCount = 0;
+  return read(file, depth, elementCount);
+}
+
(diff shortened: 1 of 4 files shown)


ufleisch:
I am a bit unsure about this PR. I can imagine that I am not the only developer staring at the `shared_pointer` scratching his head and wondering about the intentions behind it. Maybe I have to be convinced that the added complexity is really worth it or it needs at least a bit of explanation.

acts-1631:
The shared_ptr is only there to carry one parsing-local count across root atoms and recursively constructed children. It does not represent shared ownership of parser state. I’ll replace it with a stack-owned counter passed by reference through a private parsing helper.

The purpose of the count is to make the existing 50,000-per-container limit a total budget for one parse, since nested containers could otherwise each consume that limit and retain millions of element objects.

ufleisch:
Thanks! The MP4 case seems now consistent with the Matroska handling and is easier to understand.

PR #1428, opened 2026-08-20
Resolve QuickTime chunk tables linearly

acts-1631:
A QuickTime chapter track can provide large stco and stsc tables. QtChapterList::resolveSampleOffsets() currently walks every stsc entry for every chunk, so a file with N entries can force O(N²) work during MP4::File::qtChapters(). A 3.2 MB file with 200,000 entries took about 20 seconds in the release build.

Keep a cursor into stscEntries and advance it as chunk numbers increase. This resolves the same ordered table in linear time while avoiding attacker-controlled repeated scans.


ufleisch:
Thanks!

PR #1437, opened 2026-09-06
Ignore GUID when hex string length not 32

### LINKED ISSUE #1436: ASF: empty GUID attribute written for malformed string
https://github.com/taglib/taglib/blob/deadc2990767dfbda0701e0ab35fdeea653db08f/taglib/asf/asftag.cpp#L481

If the provided GUID string does not contain exactly 32 hex digits, this code still writes a GuidType attribute with an empty ByteVector. That risks emitting malformed ASF GUID attributes; it should reject/ignore invalid GUID values instead.

A suggested fix is the check:

```c++
if(data.size() != 16) {
  ignoredProps.insert(prop, str);
  break;
}
```

(disclaimer: discovered by AI)

sbooth:
Fixes #1436

### DIFF
--- a/taglib/asf/asftag.cpp
+++ b/taglib/asf/asftag.cpp
@@ -457,7 +457,6 @@ PropertyMap ASF::Tag::setProperties(const PropertyMap &props)
           addAttribute(name, !str.isEmpty() && str != "0" && str.upper() != "TRUE");
           break;
         case Attribute::GuidType: {
-          ByteVector data;
           String hexStr;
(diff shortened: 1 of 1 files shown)


ufleisch:
The property map returned by `setProperties()` should contain the *unsupported* properties, and not properties with an invalid value. The current behavior sets an empty value if an invalid value is given, which is consistent with the behavior of other such cases, e.g. when a value cannot be converted to a number. Therefore I would rather keep the existing behavior.

PR #1441, opened 2026-09-11
ASF: Verify if header size is valid

ufleisch:
When saving a .wma file, ASF::File::save() does not validate the header's headerSize field before using it in arithmetic. If its value is below 30, an unsigned underflow and a subsequent integer overflow cause FileStream::removeBlock() to compute a corrupted target offset, resulting in an unbounded copy loop.

Credits for the discovery of this bug go to Ximena Molina Portilla, Software Engineering student at Instituto Tecnológico de Costa Rica (GitHub: @ximemolina).

### DIFF
--- a/taglib/asf/asffile.cpp
+++ b/taglib/asf/asffile.cpp
@@ -713,7 +713,11 @@ void ASF::File::read()
 
   bool ok;
   d->headerSize = readQWORD(this, &ok);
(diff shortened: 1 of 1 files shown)


ximemolina:
This succesfully fixes the ASF vulnerability found!

PR #1443, opened 2026-09-15
bugfix #1442: RIFF/WAV: do not abort tags parsing when encountering invalid chunk names, skip them instead

### LINKED ISSUE #1442: RIFF/WAV: one invalid chunk ID silently discards ALL valid tags
## Summary

`RIFF::File::read()` `break`s out of the whole RIFF chunk scan on the first
chunk whose fourCC fails `isValidChunkName()`. A single 8-byte malformed chunk
*before* the tag chunks therefore makes TagLib return **empty** ID3v2 and INFO
tags for a file whose tags are valid and readable by ffmpeg / music-metadata.

## Repro

`demo.wav` (attached, 1858 B): `fmt ` + `data` + stray zero-size chunk
`"\x00n\x00i"` + valid `LIST INFO` + valid `id3 ` ID3v2.3. ffprobe plays it
(`duration=0.100000`). 

```cpp
#include <taglib/fileref.h>
#include <taglib/tag.h>
#include <iostream>

int main(int argc, char **argv) {
    TagLib::FileRef f(argv[1]);
    if (f.isNull()) return 1;
    TagLib::Tag *t = f.tag();
    std::cout << "title=[" << t->title().to8Bit(true) << "]\n";
    std::cout << "album=[" << t->album().to8Bit(true) << "]\n";
}
```

Expected: `Demo Title / Demo Album`.
Actual (TagLib 2.3.2): nothing

## Root Cause

`taglib/riff/rifffile.cpp`, `RIFF::File::read()` (lines 342-344):

```cpp
if(!isValidChunkName(chnkName)) {
  debug("RIFF::File::read() -- Chunk '" + chnkName + "' has invalid ID");
  break;   // this aborts the whole scan
}
```

The valid `LIST`/INFO and `id3`/ID3v2 chunks sit *after* the stray chunk, so
they are never scanned.

## Suggested Fix

An invalid chunk *name* should be skipped, not terminal (RIFF ignores unknown
chunks; `JUNK`/padding are legal). Keep the `break` only for oversized chunks
(the real safety stop):

```cpp
if(!isValidChunkName(chnkName)) {
  debug("RIFF::File::read() -- Chunk '" + chnkName + "' has invalid ID; skipping");
  offset += 8;   // offset still points at the chunk header
  continue;
}
```

MinnieTheMoocher:
fixes: #1442


MinnieTheMoocher:
see discussion at linked ticket: closed

